PR #469, opened 2013-05-15
Issue463

bvancea:
I hope I've understood the field schema correctly. Please let me know if I need to change anything :)

I've also added small .vcf files for testing purposes.


sdumitriu:
This functionality is now part of [the variant store](https://github.com/phenotips/variant-store/)

PR #1239, opened 2014-11-25
Issue #830: Added a Debian package creator

avpoliakov:
Assuming that the current directory contains only one version of the Phenotips Debian package, here is how you install it:
sudo dpkg -i phenotips-debian-common\* phenotips-debian-tomcat7*


marta-:
@allasm please take a look

marta-:
@itaiGershtansky please take a look

PR #1324, opened 2014-12-16
Issue-#1311:Permissions summary does not properly display the name of gr...

mjshepherd:
...oups

### DIFF
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
@@ -1067,6 +1067,9 @@ $xwiki.ssx.use($serviceDocName)
   #else
     #if($owner.isGroup())
       #set ($ownerName = $xwiki.getDocument($ownerReference.toString()).getTitle())
+      #if ($ownerName=="" || !$ownerName)
+        #set ($ownerName = $xwiki.getUserName("$ownerReference", false))
+      #end
     #else
       #set ($ownerName = $xwiki.getUserName("$ownerReference", false))
     #end
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
@@ -1065,7 +1065,11 @@ $xwiki.ssx.use($serviceDocName)
   #elseif ($ownerReference.equals($xcontext.userReference))
     #set ($ownerName = 'you')
   #else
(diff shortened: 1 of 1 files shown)


components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -1065,7 +1065,11 @@ $xwiki.ssx.use($serviceDocName)
   #elseif ($ownerReference.equals($xcontext.userReference))
     #set ($ownerName = 'you')
   #else
-    #set ($ownerName = $xwiki.getUserName("$ownerReference", false))
+    #if($owner.isGroup())
+      #set ($ownerName = $xwiki.getDocument($ownerReference.toString()).getTitle())
```
sdumitriu: There's no need to call `ownerReference.toString()`, for several reasons:

- Velocity automatically converts parameters to strings, if a method with the requested name and provided parameter type doesn't exist but a method with `String` parameters exists
- using strings as document names is deprecated, you should try to use `DocumentReference`s as often as possible
- while `Owner.getUser()` officially returns an `EntityReference`, the actual type is `DocumentReference`, and there is an `XWiki.getDocument(DocumentReference)` method, which is the best one to use

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -1065,7 +1065,14 @@ $xwiki.ssx.use($serviceDocName)
...
+      #if ("$!ownerName"=='')
```
sdumitriu: Actually you should use `#set ($ownerName = $xwiki.getDocument($ownerReference).plainTitle)`, this one already falls back to the document name if a title is not set, plus it is safer since it discards any wiki or HTML markup, which is safer.

sdumitriu: Codestyle: use spaces around `==`.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -1065,7 +1065,14 @@ $xwiki.ssx.use($serviceDocName)
...
+        #set ($ownerName = $ownerReference.name)
+      #end
+    #else
+      #set ($ownerName = $xwiki.getUserName("$ownerReference", false))
+    #end
```
sdumitriu: See 1babd10afda8348e4c6bc037e813b3ea5acb798b

Issue #1330, opened 2014-12-18
Patient age does not reset after birth date is cleared


mjshepherd:
Create a patient with a birth date and some measurements
Clear the birth date
Note that age does not reset to blank value

PR #1332, opened 2014-12-18
Issue-#1330:Patient age does not reset after birth date is cleared

mjshepherd:


components/patient-measurements/api/src/main/java/org/phenotips/measurements/internal/MeasurementAgeUpdater.java
```diff
@@ -95,6 +93,8 @@ public void onEvent(Event event, Object source, Object data)
             Date measurementDate = measurement.getDateValue(DATE_PROPERTY_NAME);
             if (measurementDate == null) {
                 measurement.removeField(AGE_PROPERTY_NAME);
+            } else if (birthDate == null) {
```
sdumitriu: Since the same action is performed as in the above branch, update the condition:

    if (measurementDate == null || birthDate == null) {

Issue #1309, opened 2014-12-15
Users can transfer ownership to non-existing users and groups they are not members of


sdumitriu:
The value entered in the _transfer ownership to another user_ input is not validated on the server side, so a user can transfer ownership to a non-existing user, or to a group that they shouldn't normally be allowed to transfer ownership to.

PR #1347, opened 2015-01-06
Issue-#1309: Users can transfer ownership to non-existing users and grou...

mjshepherd:
...ps they are not members of

### DIFF
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
@@ -96,20 +96,19 @@
   #set ($owner = "$!{request.get('owner')}")
   #if ("$!{owner}" != "")
     ##check the validity of the new owner
-    #set($ownerDoc = $services.model.resolveDocument($owner))
-    #set($ownerDocRef = $services.model.resolveDocument("$owner"))
+    #set ($ownerDocRef = $services.model.resolveDocument("$owner"))
     #if ($xwiki.getDocument($ownerDocRef).getObject("PhenoTips.PhenoTipsGroupClass"))
-      #set($validGroup = true)
-      ##check to see if user has permission to transfer ownership
+      #set ($validGroup = true)
+      ## Check to see if user has permission to transfer ownership
       #set ($adminGroup = $owner + " Administrators")
       #if (!$xwiki.getUser().isUserInGroup($owner) &amp;&amp; !$xwiki.getUser().isUserInGroup($adminGroup) &amp;&amp; !$hasAdmin)
         #set ($cannotTransfer = true)
       #end
-    #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
-      #set($validUser = true)
+    #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute().count() &gt; 0)
+      #set ($validUser = true)
     #end
-    #if ( !($validUser || $validGroup))
-      #set($invalidOwner = true)
+    #if (!($validUser || $validGroup))
+      #set ($invalidOwner = true)
     #end
   #end
   #if (!$patient)
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
@@ -102,18 +102,15 @@
       #set($validGroup = true)
       ##check to see if user has permission to transfer ownership
       #set ($adminGroup = $owner + " Administrators")
-      #if (!$xwiki.getUser().isUserInGroup($owner) &amp;&amp; !$xwiki.getUser().isUserInGroup($adminGroup))
+      #if (!$xwiki.getUser().isUserInGroup($owner) &amp;&amp; !$xwiki.getUser().isUserInGroup($adminGroup) &amp;&amp; !$hasAdmin)
         #set ($cannotTransfer = true)
-     #end
-     #if ($hasAdmin)
-       #set($cannotTransfer = false)
-     #end
-   #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
+      #end
+    #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
       #set($validUser = true)
-   #end
-   #if ( !($validUser || $validGroup))
+    #end
+    #if ( !($validUser || $validGroup))
       #set($invalidOwner = true)
-   #end
+    #end
   #end
   #if (!$patient)
     #set ($discard = $response.setStatus(400))## no patient : bad request
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
@@ -94,12 +94,39 @@
   #set ($collaborators = $request.getParameterValues('collaborator'))
   #set ($accessLevels = $request.getParameterValues('accessLevel'))
   #set ($owner = "$!{request.get('owner')}")
+  #if ("$!{owner}" != "")
+    ##check the validity of the new owner
+    #set($ownerDoc = $services.model.resolveDocument($owner))
+    #set($ownerDocRef = $services.model.resolveDocument("$owner"))
+    #if ($xwiki.getDocument($ownerDocRef).getObject("PhenoTips.PhenoTipsGroupClass"))
+      #set($validGroup = true)
+      ##check to see if user has permission to transfer ownership
+      #set ($adminGroup = $owner + " Administrators")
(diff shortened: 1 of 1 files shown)


components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,39 @@
...
+     #end
+     #if ($hasAdmin)
+       #set($cannotTransfer = false)
+     #end
+   #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
+      #set($validUser = true)
+   #end
+   #if ( !($validUser || $validGroup))
+      #set($invalidOwner = true)
+   #end
```
antonkatz: Missing an indent

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,39 @@
   #set ($collaborators = $request.getParameterValues('collaborator'))
   #set ($accessLevels = $request.getParameterValues('accessLevel'))
   #set ($owner = "$!{request.get('owner')}")
+  #if ("$!{owner}" != "")
```
antonkatz: It doesn't seem that the case where the ```owner``` is empty is checked.
The ```#elseif ($invalidOwner)``` and ```#elseif ($cannotTransfer)``` below will evaluate to ```false```.
As far as I can read in the code, that will allow changing the collaborators with an empty ```owner```.

The safest thing to do would be to initially assume that the owner is invalid, and ```cannotTransfer``` is true, by setting those variables before the ```if``` clause.

mjshepherd: You can change the collaborators with an empty owner. Note that the owner in this case is the *new owner* that this patient is being transferred to. An empty $owner variable refers to no change in ownership.

antonkatz: Ah. I see.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
...
+    ##check the validity of the new owner
+    #set($ownerDoc = $services.model.resolveDocument($owner))
```
sdumitriu: This one isn't used.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
...
+      #set($validGroup = true)
+      ##check to see if user has permission to transfer ownership
+      #set ($adminGroup = $owner + " Administrators")
+      #if (!$xwiki.getUser().isUserInGroup($owner) &amp;&amp; !$xwiki.getUser().isUserInGroup($adminGroup) &amp;&amp; !$hasAdmin)
+        #set ($cannotTransfer = true)
+      #end
+    #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
+      #set($validUser = true)
+    #end
+    #if ( !($validUser || $validGroup))
```
sdumitriu: CS: no space before `!`

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
   #set ($collaborators = $request.getParameterValues('collaborator'))
   #set ($accessLevels = $request.getParameterValues('accessLevel'))
   #set ($owner = "$!{request.get('owner')}")
+  #if ("$!{owner}" != "")
+    ##check the validity of the new owner
+    #set($ownerDoc = $services.model.resolveDocument($owner))
+    #set($ownerDocRef = $services.model.resolveDocument("$owner"))
+    #if ($xwiki.getDocument($ownerDocRef).getObject("PhenoTips.PhenoTipsGroupClass"))
+      #set($validGroup = true)
+      ##check to see if user has permission to transfer ownership
```
sdumitriu: CS: space after `##`, uppercase first letter

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
   #set ($collaborators = $request.getParameterValues('collaborator'))
   #set ($accessLevels = $request.getParameterValues('accessLevel'))
   #set ($owner = "$!{request.get('owner')}")
+  #if ("$!{owner}" != "")
+    ##check the validity of the new owner
+    #set($ownerDoc = $services.model.resolveDocument($owner))
+    #set($ownerDocRef = $services.model.resolveDocument("$owner"))
+    #if ($xwiki.getDocument($ownerDocRef).getObject("PhenoTips.PhenoTipsGroupClass"))
+      #set($validGroup = true)
```
sdumitriu: CS: space at `#set (` (two lines above, too)

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
sdumitriu: This check isn't valid. Any group member can transfer ownership **to** the group, but only group admin can transfer ownership **away from** it.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
...
+      ##check to see if user has permission to transfer ownership
+      #set ($adminGroup = $owner + " Administrators")
+      #if (!$xwiki.getUser().isUserInGroup($owner) &amp;&amp; !$xwiki.getUser().isUserInGroup($adminGroup) &amp;&amp; !$hasAdmin)
+        #set ($cannotTransfer = true)
+      #end
+    #elseif ($services.query.xwql("from doc.object(XWiki.XWikiUsers) as user where user.name = :ownerName").bindValue("ownerName", "$owner").execute() != "[]")
```
sdumitriu: Weird check... Instead of running the query, getting the documents, and comparing with an empty list, use `.count() > 0`.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,36 @@
...
   #if (!$patient)
     #set ($discard = $response.setStatus(400))## no patient : bad request
     Could not find patient with identifier "$!{request.patient}", cannot complete the update action.
-  #elseif($collaborators.size() != $accessLevels.size())
+  #elseif ($collaborators.size() != $accessLevels.size())
     #set ($discard = $response.setStatus(400))## mismatched user/level parameters, bad request
     The request is malformed: could not match each collaborator with exactly one level of access.
+  #elseif ($invalidOwner)
+    #set ($discard = $response.setStatus(400))## No user found to transfer ownership to
+    Could not find the user or group: "$!owner".
```
sdumitriu: Since you're working on translations, you could have already used some here.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,35 @@
...
+    #end
+  #end
   #if (!$patient)
     #set ($discard = $response.setStatus(400))## no patient : bad request
     Could not find patient with identifier "$!{request.patient}", cannot complete the update action.
-  #elseif($collaborators.size() != $accessLevels.size())
+  #elseif ($collaborators.size() != $accessLevels.size())
     #set ($discard = $response.setStatus(400))## mismatched user/level parameters, bad request
     The request is malformed: could not match each collaborator with exactly one level of access.
+  #elseif ($invalidOwner)
```
sdumitriu: `$invalidOwner` might not have been initialized. Same for `$cannotTransfer`. Make sure you always initialize variables.

This code here can't have any bad security consequences, but you could actually be hit by a different bug, possibly affecting security, if you check the value of an uninitialized variable that is actually initialized outside your code and is left in the context. You're only setting `invalidOwner` to `true` once here, but it could be set to `true` by a different template, with a different meaning. Since all Velocity variables are global by default, you should always be careful about their initialization.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientAccessRightsManagement.xml
```diff
@@ -94,12 +94,39 @@
...
   #set ($accessLevels = $request.getParameterValues('accessLevel'))
   #set ($owner = "$!{request.get('owner')}")
+  #if ("$!{owner}" != "")
+    ##check the validity of the new owner
+    #set ($invalidOwner = false)
+    #set ($validGroup = false)
+    #set ($validUser = false)
+    #set ($cannotTransfer = false)
+    #set ($ownerDocRef = $services.model.resolveDocument("$owner"))
+    #if ($xwiki.getDocument($ownerDocRef).getObject("PhenoTips.PhenoTipsGroupClass"))
```
sdumitriu: For performance reasons, if you don't have any variables to interpolate in a string, in Velocity code it is better to use single quotes.

Issue #1049, opened 2014-08-07
Phenotypes appearing more than once in the predefined list don't show up in the selection summary on re-edit


sdumitriu:
To reproduce:
- go to the administration, _Phenotypes displayed by default_
- select the _Detailed physical phenotype configuration_, save
- create a new patient
- select _Broad finger_ and _Arachnodactyly_ from _Upper extremities_
- save
- edit
- in the selection summary, only _Broad finger_ shows up, although _Arachnodactyly_ is selected (`Y`), and it shows up in view mode as well

allasm:
I think Anton has a point, since his example phenotype ("arachnodactyly") describes long fingers, which in theory can be applied separately to lower and upper limbs. And if the doctor meant to show long fingers on the lower limbs, he/she may be surprised if after a reload it will be checked in the upper limbs section.

PR #1348, opened 2015-01-06
Issue #1049: Phenotypes appering more than once in the predefined list d...

antonkatz:
...on't show up in the selection summary on re-edit. Made the behaviour of the phenotype display function less toggle like.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -735,10 +735,20 @@ document.observe('xwiki:dom:loaded', function() {
              if (forceUpdate) {
                 this._silentRemove(inputElt.value, type);
              }
+             var group = this._getGroup(sectionId);
              var exists = this._find(inputElt.value, type);
-             if (!exists) {
-               var group = this._getGroup(sectionId);
+             var existsDifferentGroup = false;
+             if (exists) {
+               var existsGroup = exists.up('.summary-group');
+               existsDifferentGroup = group != existsGroup;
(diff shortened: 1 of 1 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -735,10 +735,20 @@ document.observe('xwiki:dom:loaded', function() {
              if (forceUpdate) {
```
antonkatz: Sergiu, is the following what you're looking for?

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -735,10 +735,20 @@ document.observe('xwiki:dom:loaded', function() {
...
+               var existsGroup = exists.up('.summary-group');
+               existsDifferentGroup = group != existsGroup;
+             }
+
+             if (!exists || existsDifferentGroup) {
                if (!group) {return;}
+               if (existsDifferentGroup) {
+                 // The section id of the `exists` variable should ideally be passed into `_silentRemove`.
+                 this._silentRemove(inputElt.value, type)
+               }
```
antonkatz: If there is to be no ```_silentRemove``` here, then when someone adds the same phenotype in two different sections it will appear twice in the summary (under different sections).
I wounder if that is the better approach.

sdumitriu:
I remember that this didn't fix the issue, but I can't see my comment where I mentioned this...

veronikaslc:
Unreproducible in current master.

Issue #1323, opened 2014-12-16
With no birthdate set, unparsed code is displayed in view mode for the measurements evaluation


sdumitriu:
- create a new patient (no birthdate)
- add a new set of measurements
- specify a few values
- save
- `${pValue}th pctl (${sdValue}SD)` is displayed next to the measurement value

PR #1356, opened 2015-01-09
Issue-#1323:With no birthdate set, unparsed code is displayed in view mo...

mjshepherd:
...de for the measurements evaluation

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -1531,7 +1531,6 @@ document.observe('xwiki:dom:loaded', function() {
     _updateDataRow : function(row) {
       // Generate the query string to pass to the measurement evaluation service
       var qs = this._createRowQueryString(row);
-      if (qs == '') {return;}
 
       var sex = $(document.documentElement).down('input[name$=gender]:checked');
       var requestID = ++this.lastRequestID;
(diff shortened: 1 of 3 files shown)


components/patient-measurements/ui/src/main/resources/PhenoTips/MeasurementsClass.xml
```diff
@@ -172,6 +172,7 @@
   #set ($dValue = "${pValue}#if($pValue%10==1)^^st^^#elseif($pValue%10==2)^^nd^^#elseif($pValue%10==3)^^rd^^#{else}^^th^^#end pctl (#if ($sdValue &gt;= 0)+#end${sdValue}SD~)")
 #end
 (% class="${name} displayed-value $!{services.measurements.getFuzzyValue($!sdValue)}" %)${dValue}(%%)
+
```
sdumitriu: Whe the extra newline?

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -1531,7 +1531,6 @@ document.observe('xwiki:dom:loaded', function() {
     _updateDataRow : function(row) {
       // Generate the query string to pass to the measurement evaluation service
       var qs = this._createRowQueryString(row);
-      if (qs == '') {return;}
```
sdumitriu: I don't think that removing this line is optimal. Instead of removing it, which would cause a lot of extra work to be performed, there should be some extra cleanup before returning.

Issue #796, opened 2014-03-11
Add the ability to choose whether to present all studies or only the ones selected by workgroups


sdumitriu:
Currently, in order to be able to choose a study for a patient, the current user has to be a member of a group that explicitly "subscribed" to that study. It should be possible in the administration to allow users to select from all the defined studies.

PR #1377, opened 2015-01-20
Issue #796: Add the ability to choose whether to present all studies or ...

mjshepherd:
...only the ones selected by workgroups


components/studies/ui/src/main/resources/PhenoTips/StudyClass.xml
```diff
@@ -125,6 +125,46 @@
   <object>
     <name>PhenoTips.StudyClass</name>
     <number>0</number>
+    <className>PhenoTips.StudyVisibilityClass</className>
```
sdumitriu: I don't think you need to add this object to the `Study`__`Class`__, since this isn't an actual Study. Perhaps you want to add it to the `StudyTemplate`?

antonkatz: I think ```StudyClass``` is the configuration document, not a template class.

sdumitriu: Right, I didn't read the rest of the code. But in this case, the configuration shouldn't be here anyway. Don't mix classes with real configuration, it makes upgrading harder.

components/skin/ui/src/main/resources/PhenoTips/CreatePatientRecordForm.xml
```diff
@@ -50,28 +50,40 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
       ## Is there more than one sheet option?
       ## Check out all the groups this user is part of
-      #set ($toProcess = $collectionstool.queue)
-      #set ($discard = $toProcess.add($xcontext.userReference.toString()))
-      #set ($groups = [])
-      #set ($studies = [])
-      #foreach ($iteration in [0..1000])
-        #if ($toProcess.empty)
-          #break
+      #set($studiesVisibilityPreference=$xwiki.getDocument('PhenoTips.StudyClass').getObject('PhenoTips.StudyVisibilityClass').getProperty('visibility').value)
```
sdumitriu: CS: whitespace!

components/skin/ui/src/main/resources/PhenoTips/CreatePatientRecordForm.xml
```diff
@@ -50,28 +50,40 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
+      #if ($studiesVisibilityPreference == 'open')
+        #set ($studyClassName = 'PhenoTips.StudyClass')
+        #set ($studyTemplateName = 'PhenoTips.StudyTemplate')
+        #set ($query = $services.query.xwql("from doc.object(${studyClassName}) as std where doc.fullName &lt;&gt; :template").bindValue('template', $studyTemplateName))
+        #set ($studyDocs = $xwiki.wrapDocs($query.execute()))
+        #set ($studies = $collectionstool.sortedSet)
+        #foreach ($study in $studyDocs)
+          #set ($discard = $studies.add("$study"))
```
sdumitriu: It's safer to use:

    #set ($discard = $studies.add($study.documentReference.toString()))

It's safest to use:

    #set ($discard = $studies.add($services.model.serialize($study.documentReference, 'default')))

components/skin/ui/src/main/resources/PhenoTips/CreatePatientRecordForm.xml
```diff
@@ -50,28 +50,40 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
-        #set ($item = $toProcess.poll())
-        #foreach ($group in $xwiki.wrapDocs($xwiki.rightsmanager.getAllGroupsNamesForMember($item)))
-          #set ($groupObj = $group.getObject('PhenoTips.PhenoTipsGroupClass'))
-          #if (!$groupObj &amp;&amp; $group.documentReference.name.endsWith(' Administrators'))
-            #set ($groupObj = $xwiki.getDocument($services.model.resolveDocument($stringtool.removeEnd($group.documentReference.name, ' Administrators'), $group.documentReference.parent)).getObject('PhenoTips.PhenoTipsGroupClass'))
+      #else
+        #set ($toProcess = $collectionstool.queue)
+        #set ($discard = $toProcess.add($xcontext.userReference.toString()))
+        #set ($groups = [])
+        #set ($studies = [])
```
sdumitriu: This code here only checks if there are any studies that apply to the current user, so it doesn't care about the order. For performance, this could be greatly simplified: the `open` case could just `count()` the number of studies and check that it's `> 0`, while the `restricted` case could `#break` at the first study found.

components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
```diff
@@ -622,28 +634,40 @@ $xwiki.ssx.use($className)
...
-  #set ($item = $toProcess.poll())
-  #foreach ($group in $xwiki.wrapDocs($xwiki.rightsmanager.getAllGroupsNamesForMember($item)))
-    #set ($groupObj = $group.getObject('PhenoTips.PhenoTipsGroupClass'))
-    #if (!$groupObj &amp;&amp; $group.documentReference.name.endsWith(' Administrators'))
-      #set ($groupObj = $xwiki.getDocument($services.model.resolveDocument($stringtool.removeEnd($group.documentReference.name, ' Administrators'), $group.documentReference.parent)).getObject('PhenoTips.PhenoTipsGroupClass'))
+#else
+  #set ($toProcess = $collectionstool.queue)
+  #set ($discard = $toProcess.add($xcontext.userReference.toString()))
+  #set ($groups = [])
+  #set ($studies = [])
```
sdumitriu: Here the order in which the studies are listed is important, so leave it as:

    #set ($studies = $collectionstool.sortedSet)

components/studies/ui/src/main/resources/PhenoTips/StudyClass.xml
```diff
@@ -544,7 +804,7 @@
       <name/>
     </property>
     <property>
-      <parse>0</parse>
+      <parse>1</parse>
```
sdumitriu: Don't change this.

Issue #1224, opened 2014-11-17
Phenotype mapping documents should use plain JSON syntax


sdumitriu:
Currently it's almost a JSON, wrapped with some Velocity syntax.

PR #1387, opened 2015-01-26
Issue #1224: Phenotype mapping documents should use plain JSON syntax

mjshepherd:


components/patient-tools/src/main/java/org/phenotips/tools/PhenotypeMappingService.java
```diff
@@ -226,13 +228,9 @@ private Object getMapping(String mappingName)
         DocumentReference mappingDoc = getMappingDocument();
         Object result = getMapping(mappingDoc, mappingName);
         if (result == null) {
-            Map<String, Object> mappings;
             try {
-                if (this.bridge.getDocumentContentForDefaultLanguage(mappingDoc).contains("{{velocity")) {
-                    mappings = parseVelocityMapping(mappingDoc);
-                } else {
-                    mappings = parseGroovyMapping(mappingDoc);
-                }
```
sdumitriu: Don't remove this code, backwards compatibility is important. Existing users might have custom mappings still using the old syntax, which will get broken after an upgrade. Add a new branch on top, checking if the content starts with `{`.

mjshepherd: Ok. I'm guessing that the phenotype preview also has to be backwards compatible.

Issue #1372, opened 2015-01-15
Selecting "HP:0012533 Allodynia" as a symptom breaks the diagnosis suggestions


sdumitriu:
- create a new patient
- select a few phenotypes, including HP:0012533 Allodynia
- go to the _Diagnosis_ section
- with _Allodynia_ included in the terms, no diagnosis is displayed
- disabling _Allodynia_ will bring suggestions back

buske:
Is this caused by the term not being present in the version of HPO that BOQA uses?

sdumitriu:
I don't know, I'm "just reporting it".

PR #1390, opened 2015-01-28
Issue #1372: Selecting "HP:0012533 Allodynia" as a symptom breaks the di...

mjshepherd:
...agnosis suggestions

### DIFF
--- a/components/diagnosis-suggestion/api/pom.xml
+++ b/components/diagnosis-suggestion/api/pom.xml
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.79</coverage.instructionRatio>
+    <coverage.instructionRatio>0.78</coverage.instructionRatio>
   </properties>
 
   <dependencies>
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>de.charite.compbio</groupId>
       <artifactId>boqa</artifactId>
-      <version>1.0.2</version>
+      <version>1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/components/diagnosis-suggestion/api/pom.xml
+++ b/components/diagnosis-suggestion/api/pom.xml
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.81</coverage.instructionRatio>
+    <coverage.instructionRatio>0.79</coverage.instructionRatio>
   </properties>
 
   <dependencies>
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>de.charite.compbio</groupId>
       <artifactId>boqa</artifactId>
-      <version>1.0.1</version>
+      <version>1.0.2</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/components/diagnosis-suggestion/api/pom.xml
+++ b/components/diagnosis-suggestion/api/pom.xml
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.78</coverage.instructionRatio>
+    <coverage.instructionRatio>0.82</coverage.instructionRatio>
   </properties>
 
   <dependencies>
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>de.charite.compbio</groupId>
       <artifactId>boqa</artifactId>
-      <version>1.0-SNAPSHOT</version>
+      <version>1.0.1</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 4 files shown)


components/diagnosis-suggestion/api/pom.xml
```diff
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.81</coverage.instructionRatio>
+    <coverage.instructionRatio>0.79</coverage.instructionRatio>
```
sdumitriu: That's not good...

components/diagnosis-suggestion/api/src/main/java/org/phenotips/diagnosis/internal/DefaultDiagnosisService.java
```diff
@@ -138,6 +138,10 @@ public void initialize() throws InitializationException
         }
 
         // Get marginals
+        if (o.isEmpty()) {
+            return null;
```
sdumitriu: Return an empty list (`Collections.emptyList()`).

components/diagnosis-suggestion/ui/src/main/resources/PhenoTips/DiseasePredictService.xml
```diff
@@ -56,7 +56,7 @@
     #end
 #if ("$!{request.format}" == 'html')
 #set ($results = $services.diagnosis.get($data, $limit))
-#if ($results.size() &gt; 0)
+#if ($results &amp;&amp; $results.size() &gt; 0)
```
sdumitriu: If you returned an empty list, this change wouldn't be needed.

components/diagnosis-suggestion/api/src/main/java/org/phenotips/diagnosis/internal/DefaultDiagnosisService.java
```diff
@@ -204,9 +208,14 @@ public int compare(Integer o1, Integer o2)
...
-        int id = this.boqa.getTermIndex(t);
-        o.observations[id] = true;
-        this.boqa.activateAncestors(id, o.observations);
+        try {
+            int id = this.boqa.getTermIndex(t);
+            o.observations[id] = true;
+            this.boqa.activateAncestors(id, o.observations);
+        } catch (NullPointerException e) {
+            this.logger.warn(String.format(
+                "Unable to find the boqa index of '%s'.", t));
```
sdumitriu: `org.slf4j.Logger` has a really nice feature, [_parameterized logging_](http://slf4j.org/faq.html#logging_performance). You should write this as:

    this.logger.warn("Unable to find the boqa index of [{}].", t);

`{}` is an anchor, and as a convention we put them inside square brackets to make it easier to delimit variable values from the fixed logging message.

components/diagnosis-suggestion/api/pom.xml
```diff
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>de.charite.compbio</groupId>
       <artifactId>boqa</artifactId>
-      <version>1.0.1</version>
+      <version>1.0.2</version>
```
sdumitriu: Until 1.0.2 is actually released, this change is invalid. You may use `1.0-SNAPSHOT` to use the latest version.

components/diagnosis-suggestion/api/src/main/java/org/phenotips/diagnosis/internal/DefaultDiagnosisService.java
```diff
@@ -202,11 +208,17 @@ public int compare(Integer o1, Integer o2)
         return results;
     }
 
-    private void addTermAndAncestors(Term t, Observations o)
+    private Boolean addTermAndAncestors(Term t, Observations o)
```
sdumitriu: There's a difference between the primitive type `boolean` and the class `Boolean`, and you seem to be using a primitive in the rest of the code. Use `boolean` as the return type.

components/diagnosis-suggestion/api/pom.xml
```diff
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.81</coverage.instructionRatio>
+    <coverage.instructionRatio>0.78</coverage.instructionRatio>
```
sdumitriu: It's rarely OK to decrease the coverage ratio, you should add a new test instead.

sdumitriu: Since this commit fixes a bug, you should add a test that checks the fix stays in place, so that a future regression will be caught by the test. That's why some tests are called "regression tests".

components/diagnosis-suggestion/api/pom.xml
```diff
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>de.charite.compbio</groupId>
       <artifactId>boqa</artifactId>
-      <version>1.0.1</version>
+      <version>1.0-SNAPSHOT</version>
```
sdumitriu: Is it still needed to change the version?

Issue #1361, opened 2015-01-13
Spaces in gene name breaks GeneNameService


buske:
For example, gene "hello world":

Querying with a space in the field works fine:
https://phenomecentral.org/get/PhenoTips/GeneNameService?outputSyntax=plain&q=hello%20world

But if the gene is already part of a record, it breaks the GeneNameService for all genes in the document, causing none of them to have details:
https://phenomecentral.org/get/PhenoTips/GeneNameService?outputSyntax=plain&g=hello%20world
```
Failed to execute the [velocity] macroorg.xwiki.rendering.macro.MacroExecutionException:
[...]
Caused by: java.net.URISyntaxException: Illegal character in path at index 44: http://rest.genenames.org/fetch/symbol/hello world
	at java.net.URI$Parser.fail(URI.java:2829)
	at java.net.URI$Parser.checkChars(URI.java:3002)
	at java.net.URI$Parser.parseHierarchical(URI.java:3086)
	at java.net.URI$Parser.parse(URI.java:3034)
	at java.net.URI.<init>(URI.java:595)
	at java.net.URI.create(URI.java:857)
```

PR #1397, opened 2015-01-30
Issue #1361: Spaces in gene name breaks GeneNameService

mjshepherd:


components/ontology-access/hgnc/api/src/main/java/org/phenotips/ontology/internal/GeneNomenclature.java
```diff
@@ -151,7 +151,8 @@ public OntologyTerm getTerm(String id)
     {
         OntologyTerm result = this.cache.get(id);
         if (result == null) {
-            HttpGet method = new HttpGet(fetchServiceURL + "symbol/" + id);
+            id = id.replaceAll("\\s", "");
```
sdumitriu: Are you sure it's OK to just remove whitespace? You should encode the ID instead.

components/ontology-access/hgnc/api/src/main/java/org/phenotips/ontology/internal/GeneNomenclature.java
```diff
@@ -151,7 +151,8 @@ public OntologyTerm getTerm(String id)
...
+            HttpGet method = new HttpGet(this.fetchServiceURL + "symbol/" + id);
```
sdumitriu: ... + URLEncoder.encode(id, Consts.UTF_8.name())

Issue #1434, opened 2015-02-13
Ugly stacktrace logged if the GeneNames service is unavailable


sdumitriu:
```
net.sf.json.JSONException: Invalid JSON String
        at net.sf.json.JSONSerializer.toJSON(JSONSerializer.java:143) ~[json-lib-2.4-jdk15.jar:na]
        at net.sf.json.JSONSerializer.toJSON(JSONSerializer.java:103) ~[json-lib-2.4-jdk15.jar:na]
        at net.sf.json.JSONSerializer.toJSON(JSONSerializer.java:84) ~[json-lib-2.4-jdk15.jar:na]
        at org.phenotips.ontology.internal.GeneNomenclature.getInfo(GeneNomenclature.java:306) ~[ontology-hgnc-access-api-1.0.jar:na]
        at org.phenotips.ontology.internal.GeneNomenclature.getVersion(GeneNomenclature.java:292) ~[ontology-hgnc-access-api-1.0.jar:na]
        at org.phenotips.ontology.listeners.PatientOntologyVersionRecorder.getOntologiesVersions(PatientOntologyVersionRecorder.java:128) ~[ontology-access-api-1.0.jar:na]
....

buske:
Could this explain many of the issues with saving that occur when the internet connection is slow or intermittent?

PR #1438, opened 2015-02-18
Issue #1434: If the GeneNames service is unavailable, saving patient rec...

mjshepherd:
...ords fails


components/ontology-access/hgnc/api/pom.xml
```diff
@@ -32,7 +32,7 @@
 
   <properties>
     <!-- More tests should be added when the Solr core becomes a component -->
-    <coverage.instructionRatio>0.83</coverage.instructionRatio>
```
sdumitriu: I like when that happens :smile:

Issue #1463, opened 2015-03-09
Bad display of the buttons in the "close without saving" warning in the pedigree editor


sdumitriu:
![1463](https://cloud.githubusercontent.com/assets/88663/6548699/febc84b4-c5d7-11e4-8edf-7ca2e41fa1d0.png)

PR #1467, opened 2015-03-10
Issue #1463: Bad display of the buttons in the "close without saving" wa...

mjshepherd:
...rning in the pedigree editor


sdumitriu:
I don't think the `min-width` is required at all, just remove the `80px` class.

allasm:
Without the width (or min-width) buttons look tiny-ugly (to my liking at least) with default font/zoom settings in most browsers

sdumitriu:
We can add some padding.

Issue #1162, opened 2014-10-12
Add an ontology access method for searching terms based on user input


sdumitriu:
The current quick search uses complex queries constructed in JavaScript, which requires copy/pasting code in several places. There should be a Java method (mirrored in the script service) that suggests terms best matching the user's input.

allasm:
Is it supposed to be used everywhere, e.g. here (https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js#L111) as well?

sdumitriu:
Yes. There's still some cleanup to do.

PR #1469, opened 2015-03-10
Issue #1162: Add an ontology access method for searching terms based on ...

antonkatz:
...user input

### DIFF
--- a/components/ontology-access/api/src/main/java/org/phenotips/ontology/OntologyTerm.java
+++ b/components/ontology-access/api/src/main/java/org/phenotips/ontology/OntologyTerm.java
@@ -23,6 +23,8 @@ import org.xwiki.stability.Unstable;
 
 import java.util.Set;
 
+import net.sf.json.JSON;
+
 /**
  * A term from an {@link OntologyService ontology}. A few common properties are available as explicit individual
  * methods, and any property defined for the term can be accessed using the generic {@link #get(String)} method. As a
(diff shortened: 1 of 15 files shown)


components/ontology-access/api/src/main/java/org/phenotips/ontology/OntologyTerm.java
```diff
@@ -101,4 +103,30 @@
      * @return the owner ontology
      */
     OntologyService getOntology();
+
+    /**
+     * @since 1.1-rc1
```
sdumitriu: `@since` should be after `@param @return @throws`.

components/ontology-access/api/src/main/java/org/phenotips/ontology/OntologyTerm.java
```diff
@@ -101,4 +103,30 @@
...
+     *      "is_a":[
+     *      "HP:0004325 ! Decreased body weight"
+     *    ],
+     *    "id":"HP:0001824",
+     *    "name":"Weight loss",
+     *    "def":"Reduction inexisting body weight."
+     * }
+     * @throws java.lang.Exception could happen if casting fails
+     */
+    JSON toJson() throws Exception;
```
sdumitriu: It's a bad practice to throw `Exception`, that's way too generic.

antonkatz: It's bad practice to have exceptions. When are we getting ```Option```s?

components/ontology-access/api/src/main/java/org/phenotips/ontology/internal/solr/SolrOntologyTerm.java
```diff
@@ -220,6 +241,31 @@ private long processAncestorsAtDistance(int localDistance, Set<OntologyTerm> sou
...
 
     @Override
+    public JSON toJson() throws Exception
+    {
+        JSONObject json = new JSONObject();
+
+        Iterator<Map.Entry<String, Object>> docIterator = this.doc.iterator();
+
+        while (docIterator.hasNext()) {
+            Map.Entry<String, Object> param = docIterator.next();
```
sdumitriu: The convention is that `thingIterator` describes what are the values returned by the iterator, not the type that contains the things to be iterated. So `docIterator` sounds like it iterates over a list of documents. It should be named `fieldIterator`.

Issue #1476, opened 2015-03-12
The pedigree legend is broken by special HTML characters


sdumitriu:
Due Fri, Mar 20th.

PR #1499, opened 2015-03-17
Issue #1476: The pedigree legend is broken by special HTML characters

mjshepherd:


allasm:
We also need to write a converter. Since desanitizeID() was removed IDs which are already stored in sanitized form will be incorrectly loaded by the new code

Issue #1528, opened 2015-03-31
Improve the positioning and styling of the hint text at the top of the "You may want to investigate" box


marta-:
![issue](https://cloud.githubusercontent.com/assets/651980/6923465/e4ac0a66-d79f-11e4-856f-213ac69f0035.png)

PR #1532, opened 2015-03-31
Issue #1528: Improve the positioning and styling of the hint text at the...

danielpgross:
... top of the "You may want to investigate" box


danielpgross:
I'm going to separate these issues into their own branches and then open new requests for each.

PR #1533, opened 2015-03-31
Issue 1451

antonkatz:


components/navigation/ui/src/main/resources/XWiki/SearchSuggestConfig.xml
```diff
@@ -289,6 +289,12 @@
       </sectionPriority>
     </class>
     <property>
+      <advancedOnly/>
+    </property>
+    <property>
+      <categoryPriority/>
+    </property>
+    <property>
```
antonkatz: Is this trash?

sdumitriu: Not really, PhenoTips uses an extended version of the `ConfigurableClass`, with more fields than what XWiki has, while this document comes from XWiki. To keep the changes from the official document, we should not include these extra fields.

resources/ui/src/main/resources/PhenoTips/SuggestPatientsService.xml
```diff
@@ -61,23 +61,84 @@
   #set ($query = $services.query.xwql($query))
   #set ($discard = $query.setLimit($nb))
   #set ($discard = $query.bindValue('input', "%${input}%"))
-  #set ($discard = $response.setContentType("text/xml"))
-  &lt;?xml version="1.0" encoding="UTF-8"?&gt;
   #set ($results = $query.execute())
-  &lt;results&gt;
+  #if ($request.permission)
+    #set ($requiredPermissions = "$request.permission")
```
sdumitriu: There's no need to use a string here:

    #set ($requiredPermissions = $request.permission)

resources/ui/src/main/resources/PhenoTips/SuggestPatientsService.xml
```diff
@@ -61,23 +61,84 @@
...
+  #else
+    #set ($requiredPermissions = "view")
+  #end
+  #if ($request.json)
+    $response.setContentType('application/json')
+    { "matchedPatients": [
+    #set ($first = true)
     #foreach ($item in $results)
-      #if ($xwiki.hasAccessLevel('view', $xcontext.user, $item))
+      #if ($xwiki.hasAccessLevel("$requiredPermissions", $xcontext.user, $item))
```
sdumitriu: Don't string strings.

antonkatz: But a stringed string is a string indeed

PR #1541, opened 2015-04-01
Added "solved" data controller

buske:
Includes:
- Issue #1534: pubmed id
- Issue #1535: solved status
- Issue #1536: resolution notes
- Issue #1540: solved gene symbol

Would have added tests but I couldn't see any example of testing a PatientDataController class, and the tests that were there seemed misconfigured with my Eclipse environment. :/

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/SolvedController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/SolvedController.java
@@ -44,10 +44,10 @@ import org.apache.commons.lang3.StringUtils;
 import net.sf.json.JSONObject;
 
 /**
- * Handled fields for solved patient records, including solved status, PubMed ID, gene symbol, and notes.
+ * Handles fields for solved patient records, including solved status, PubMed ID, gene symbol, and notes.
  *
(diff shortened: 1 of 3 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/SolvedController.java
```diff
@@ -0,0 +1,163 @@
...
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.apache.commons.lang3.StringUtils;
+
+import net.sf.json.JSONObject;
+
+/**
+ * Handled fields for solved patient records, including solved status, PubMed ID, gene symbol, and notes.
```
sdumitriu: `Handles`

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/SolvedController.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @version $Id$
+ * @since 1.2m2
```
sdumitriu: Uppercase `M`

Issue #1482, opened 2015-03-16
New child nodes should not have rounded corners


buske:
When creating a new node, the square and diamond icons have rounded corners for some reason:

![screenshot](https://cloud.githubusercontent.com/assets/4251264/6659826/f21a84cc-cb63-11e4-818b-f26f432cb0e9.png)

sdumitriu:
@allasm Some Unicode alternatives:
* [ ] [⬜ U+2B1C](http://www.fileformat.info/info/unicode/char/2b1c/index.htm) looks nice, but I think it's not supported by the default fonts on Windows
* [ ] [☐ U+2610](http://www.fileformat.info/info/unicode/char/2610/index.htm) seems to be available in many fonts
* [ ] [◻ U+25FB](http://www.fileformat.info/info/unicode/char/25fb/index.htm) looks misaligned on my system
* [ ] [□ U+25A1](http://www.fileformat.info/info/unicode/char/25a1/index.htm) seems to be more common, but it's too small

Can you check which of these look good on Windows/Mac/Linux?

allasm:
We also need to find a circle symbol o the same size to make it look consistent. I used "fa" icons in this case, and part of the reason was that I couldn't find a square and a circle character which would scale similarly in all the browsers/zoom levels.

danielpgross:
The sizes of the unicode characters were inconsistent across OSes, so I instead used a ```span``` with a 1px black border for the square and diamond icons.

Chrome41/OSX10.10.2:
![pedigree_editor___phenotype_database](https://cloud.githubusercontent.com/assets/4006062/6944079/fe109bdc-d85f-11e4-8a34-ab8e00405ee3.png)
FF16/Ubuntu12.10:
![ubuntu-12_10-desktop-i386__running_](https://cloud.githubusercontent.com/assets/4006062/6944170/732dd0f6-d860-11e4-901b-d02b222cb087.png)
IE11/Win8.1:
![ie11_-_win8_1__running_](https://cloud.githubusercontent.com/assets/4006062/6944107/2a4ce17e-d860-11e4-8782-7fda6456944f.png)

PR #1542, opened 2015-04-01
Issue #1482: New child nodes should not have rounded corners

danielpgross:


sdumitriu:
I would say that for consistency you should also define the circle as a rounded span with precise dimensions, otherwise a FontAwesome update might change its dimensions in the future.

Issue #1492, opened 2015-03-16
Expand/collapse all should be made more visual


buske:
These buttons should be replaced or augmented with icons so people notice them without having to read them.

Reported by Veronika.

PR #1546, opened 2015-04-01
Issue #1492: Expand/collapse all should be made more visual

danielpgross:
p0000001___phenotype_database
p0000001___phenotype_database

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -736,7 +736,10 @@ document.observe('xwiki:dom:loaded', function() {
                }
                newItem.__source = inputElt;
                var tools = new Element('span', {'class' : 'tools hint'});
-               var deleteTool = new Element('span', {'class' : 'tool'}).update("$services.localization.render('phenotips.patientSheetCode.termSuggest.delete')");
+               var deleteTool = new Element('span', {'class' : 'buttonwrapper'}).insert(
+                 new Element('a', {'class' : 'tool button secondary', 'href' : 'javascript:void(0)'}).update("$services.localization.render('phenotips.patientSheetCode.termSuggest.delete')")
+               );
+
                tools.insert(deleteTool);
                deleteTool.observe('click', function(event) {
                  var source = event.findElement('.summary-item').__source;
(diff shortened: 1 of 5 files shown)


danielpgross:
I see now that the plus icon is the same as the one used for the "New Patient Record" link. This dual meaning is confusing.

Some other options are:
- ```plus-square-o``` and ```minus-square-o```
- The unicode arrows used for expanding the hierarchy levels on the patient sheet
- ```arrow-circle-up``` and ```arrow-circle-down```

What do you think?

sdumitriu:
I'd say [`plus-square-o`](http://fortawesome.github.io/Font-Awesome/icon/plus-square-o/)/[`minus-square-o`](http://fortawesome.github.io/Font-Awesome/icon/minus-square-o/] looks the best. @marta- ?

marta-:
OK with [`plus-square-o`](http://fortawesome.github.io/Font-Awesome/icon/plus-square-o/)/[`minus-square-o`](http://fortawesome.github.io/Font-Awesome/icon/minus-square-o/). The neutral color and tiny font are part of the problem, so these should be improved as well to make "Expand all" & "Collapse all" look more like actions.

danielpgross:
Great.

In terms of making them look more like actions, there are many ways to go about this. Here are a few design options:
- Apply the "action button" styling from the skin. I don't think this is ideal since this style signifies an action which will change data: 
p0000001___phenotype_database
- Add a rounded border, background, and hover effect to give a pseudo-button impression without implying that this button will perform an action which impacts data:
p0000001___phenotype_database

I think the second option is best. What do you think?

marta-:
We have two "standard" styles for buttons. The brightly colored one (blue in this color theme) is for primary actions, commonly used for things with important effects, e.g. adding a new entry in a data table, saving the document, etc. The secondary actions have a gray-ish gradient instead, and look somewhat like what you proposed:

secondary_actions

The css class is ```button secondary```

sdumitriu:
This looks good, but I think this issue was about this part of the UI:

danielpgross:
Ah, ok, thanks for clarifying that point, @sdumitriu . 

So what exactly is the scope of this issue? Should all instances of "Expand all"/"Collapse all" and "Show"/"Hide" in the UI be replaced? Or should only "Expand all"/"Collapse all" for "Browse Categories" be replaced for now? Thanks.

marta-:
The issue was initially just about expand/collapse, but the look of these is consistent with the show/hide for sections and the 'add details'/'edit details'/'clear details'/'delete' for selected phenotypes, so we might as well change all of them.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -736,7 +736,10 @@ document.observe('xwiki:dom:loaded', function() {
                }
                newItem.__source = inputElt;
                var tools = new Element('span', {'class' : 'tools hint'});
-               var deleteTool = new Element('span', {'class' : 'tool'}).update("$services.localization.render('phenotips.patientSheetCode.termSuggest.delete')");
+               var deleteTool = new Element('span', {'class' : 'buttonwrapper'}).insert(
+                 new Element('a', {'class' : 'tool button secondary', 'href' : 'javascript:void(0)'}).update("$services.localization.render('phenotips.patientSheetCode.termSuggest.delete')")
```
sdumitriu: Don't use `javascript:` hrefs, that's a bad practice. If you don't want that to be a link that points somewhere, then don't make it a link.

danielpgross: Agreed, I'll use ```button```s instead.

Issue #1527, opened 2015-03-30
Non-standard phenotypes break the OMIM suggestions


sdumitriu:

danielpgross:
Error details as returned by the DiseasePredictService with a non-standard term set:
```
org.xwiki.rendering.macro.MacroExecutionException: Failed to evaluate Velocity Macro for content [#if ($xcontext.action == 'get')
  $response.setHeader("X-ReqNo", "$!request.reqNo")##
  #if ("$!request.q" != '')##
    #set ($limit = $mathtool.toInteger("$!&#123;request.limit}"))
    #if (!$limit || $limit &lt; 0)
      #set ($limit = 20)
    #end
    #set ($data = [])
    #foreach ($piece in $!request.getParameterValues('symptom'))
      #set($discard = $data.add($piece))
    #end
    #foreach ($piece in $!request.getParameterValues('free_symptom'))
      #set($discard = $data.add($piece))
    #end
#if ("$!&#123;request.format}" == 'html')
#set ($results = $services.diagnosis.get($data, $limit))
#if ($results.size() &gt; 0)
&#123;&#123;html clean="false"  wiki="false"}}##
&lt;ul&gt;
#foreach($term in $results)
  #set($name = $term.getName())
  #set ($indexOfSeparator = $name.indexOf(';'))
  #if ($indexOfSeparator &gt; 0)
    #set($shortName = $name.substring(0, $indexOfSeparator))
    #set($fullName = $name.replaceAll(';;', ';'))
  #else
    #set($shortName = $name)
    #set($fullName = $name)
  #end
  #set($id = $term.getId())
  &lt;li class="disorder"&gt;
    &lt;span class="id" title="$&#123;id}"&gt;&lt;a href="http://www.omim.org/entry/$&#123;id}" target="_blank" title="Read about this disorder on OMIM"&gt;[MIM:$&#123;id}]&lt;/a&gt;&lt;/span&gt;
    &lt;span class="title" title="$&#123;fullName}"&gt;$&#123;shortName}&lt;/span&gt;
  &lt;/li&gt;
#end ## foreach item
&lt;/ul&gt;
&#123;&#123;/html}}
   #set ($hasOutput = true)
#else
   #set ($hasOutput = false)
#end## results.size() &gt; 0
    #end## html format
  #else
    #set ($hasOutput = false)
  #end## non-empty query
#end## get action
##
#if (!$hasOutput)
(% class="hint" %)No matches found
#end]
[...]

PR #1547, opened 2015-04-02
Issue #1527: Non-standard phenotypes break the OMIM suggestions

danielpgross:
Fixed by not searching ontology for free-input terms

### DIFF
--- a/components/diagnosis-suggestion/api/pom.xml
+++ b/components/diagnosis-suggestion/api/pom.xml
@@ -29,7 +29,7 @@
   <name>PhenoTips - Diagnosis Suggestion Java Code</name>
 
   <properties>
-    <coverage.instructionRatio>0.82</coverage.instructionRatio>
+    <coverage.instructionRatio>0.81</coverage.instructionRatio>
   </properties>
 
   <dependencies>
--- a/components/diagnosis-suggestion/api/pom.xml
+++ b/components/diagnosis-suggestion/api/pom.xml
@@ -84,4 +84,28 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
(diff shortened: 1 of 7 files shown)


danielpgross:
Build is failing because the API has changed and clirr is failing. I'm adding an ignore for this change.

components/diagnosis-suggestion/api/pom.xml
```diff
@@ -84,4 +84,28 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
```
sdumitriu: Don't add this here, there's a special section in the root pom.xml. With one global clirr configuration, it's easy to have an overview of what changed. Otherwise, it will be hard to search for all the clirr rules in every pom file.

components/diagnosis-suggestion/api/pom.xml
```diff
@@ -84,4 +84,28 @@
...
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
+        <configuration>
+          <ignored> 
+            <!-- Remove the following ignores after we release 1.2m2 -->
+            <difference>
```
sdumitriu: You should add a comment explaining why this changed, or even better, a `<justification>` tag:

```
<justification>The API should accept both ontology terms and non-standard features.</justification>

components/diagnosis-suggestion/api/pom.xml
danielpgross: Thanks, that makes sense. I've moved it to the root pom.xml.

components/diagnosis-suggestion/api/pom.xml
danielpgross: Thanks, I've added a ```justification``` tag for each ```difference```

PR #1578, opened 2015-04-16
Issue #1573: Newly added candidate genes don't get the "reject" button

danielpgross:
Fixed


sdumitriu:
While the button does show up, it doesn't work properly. The candidate data is lost, the new rejected row is empty.

danielpgross:
The behaviour in my testing was that if the reject table was empty, the candidate data would be properly filled into the (new) first row of the reject table, but otherwise the behaviour was as you described. 

I opened this pull request to get feedback on my approach and because the issue with the data not being transferred might be beyond the scope of this issue.

What should my next action be on this issue? Should I continue with this approach and address the problem that the candidate data is lost?

sdumitriu:
Yes, the approach so far is good, but the data loss needs to be investigated.

PR #1601, opened 2015-04-18
Issue PT-1569: Implement `save` function for some of the existing contro...

antonkatz:
...llers.

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractSimpleController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractSimpleController.java
@@ -25,6 +25,7 @@ import org.phenotips.data.PatientData;
 import org.phenotips.data.PatientDataController;
 
 import org.xwiki.bridge.DocumentAccessBridge;
+import org.xwiki.context.Execution;
 
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractSimpleController.java
```diff
@@ -59,6 +61,10 @@
     @Inject
     private Logger logger;
 
+    /** Provides access to the current execution context. */
+    @Inject
+    private Execution execution;
```
sdumitriu: Use this instead:

```
@Inject
private Provider<XWikiContext> xcontext;

File: components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractSimpleController.java [hunk 2]
```diff
@@ -85,7 +91,27 @@
     @Override
     public void save(Patient patient)
     {
-        throw new UnsupportedOperationException();
+        try {
+            XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
+            BaseObject xwikiDataObject = doc.getXObject(Patient.CLASS_REFERENCE);
+            if (xwikiDataObject == null) {
+                throw new NullPointerException(ERROR_MESSAGE_NO_PATIENT_CLASS);
```
sdumitriu: Better than this would be `IlegalArgumentException`.

But I wonder if a custom exception would be better (`InvalidPatientException`), since I don't like that `ERROR_MESSAGE_NO_PATIENT_CLASS`.

antonkatz: I agree that it would be, but where will it live?

PR #1602, opened 2015-04-18
Issue-1575: Information about Solved cases is not included in the Excel ...

antonkatz:
...export.

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -1215,6 +1215,85 @@ public class DataToCellConverter
         return bodySection;
     }
 
+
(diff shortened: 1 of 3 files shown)


components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1215,6 +1215,85 @@ public DataSection isNormalBody(Patient patient)
...
+        String sectionName = "isSolved";
+        Set<String> present = this.enabledHeaderIdsBySection.get(sectionName);
+        if (present == null || present.isEmpty()) {
+            return null;
+        }
+        DataSection bodySection = new DataSection();
+        PatientData<String> patientData = patient.getData("solved");
+
+        Integer x = 0;
+        if (present.contains("solved")) {
```
antonkatz: These are left uncolored on purpose.
I'm assuming that majority of cases will be unsolved, and it could be psychologically draining to see too much red.

Issue #1572, opened 2015-04-15
Navigating by tab from the month of birth to the day takes two tabs


sdumitriu:

marta-:
@tmarathe @DeanWay can you confirm this is fixed?

tmarathe:
Yes, it is working properly now.

sdumitriu:
Did you try in Firefox and Chrome?

veronikaslc:
I did Chrome Version 43.0.2357.130 m, FireFox Version 38.0.5,

PR #1604, opened 2015-04-20
Issue #1572: Navigating by tab from the month of birth to the day takes two tabs

veronikaslc:
Fixed

### DIFF
--- a/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
@@ -4306,7 +4306,7 @@ document.observe("xwiki:dom:loaded", function() {
     },
 
     getElement : function() {
-      return this.span;
+      return this.span.down();
     },
 
     onSelect : function(callback) {
--- a/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
@@ -4254,7 +4254,7 @@ document.observe("xwiki:dom:loaded", function() {
 
   widgets.FuzzyDatePickerDropdown = Class.create({
     initialize : function(options) {
-      this.span     = new Element('span');
+      this.span     = new Element('span', {"tabIndex":"-1"});
       this.options  = options;
       this.callback = null;
     },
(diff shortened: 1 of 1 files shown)


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4306,7 +4306,7 @@ document.observe("xwiki:dom:loaded", function() {
     },
 
     getElement : function() {
-      return this.span;
+      return this.span.down();
     },
```
allasm: This will break other things, since now instead of inserting <span> the content of <span> will be inserted into the document. And <span> is there for a reason, I think IE9 (which som ehoospitals still use) is broken without it (see comment a few lines above).

Instead, if <span? is the problem, you can exclude <span> from the tab order, e.g. set tabindex="-1" or something like that
(http://stackoverflow.com/questions/3895710/explicitly-exclude-an-html-element-from-the-tab-order, https://www.google.ca/search?q=html+exclude+form+tab+order&oq=html+exclude+form+tab+order)

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4254,7 +4254,7 @@ document.observe("xwiki:dom:loaded", function() {
 
   widgets.FuzzyDatePickerDropdown = Class.create({
     initialize : function(options) {
-      this.span     = new Element('span');
+      this.span     = new Element('span', {"tabIndex":"-1"});
```
sdumitriu: I doubt this is valid, the `tabindex` attribute doesn't apply to `span` elements.

sdumitriu: It doesn't fix the issue on Firefox 37/Linux.

allasm: I remember testing this in FF on Linux without any <span> elements, with just a <select> - it still takes two tabs iff a value has just been selected (one tab closes drop-down, another navigates to the next box); without "tabindex" it still took two tabs even from a closed drop-down, but only one tab with "tabindex" fix... Maybe a more complicated JavaScript solution is necessary to manually handle tab key presses and manually navigate to the next box?

Issue #1549, opened 2015-04-07
Suggest dialogs should hide when they loose focus


meatcar:
Currently, dismissing the Phenotype Search & Suggest dialog is only possible by only clicking the "hide suggestion" links at the top and bottom corners of the dialog.

It should instead behave like a proper modal, and be dismissed when any other part of the page is clicked (i.e. the dialog looses focus). When clicking back on the search box, the dialog should pop back up (it doesn't now). This will move Phenotips closer to being well-behaved on mobile, as well as improve the UI/UX.

![capture](https://cloud.githubusercontent.com/assets/191622/7028141/f327691e-dd21-11e4-9a80-147bbb67364c.PNG)

mjshepherd:
This behaviour should also be present in other suggest dialogs. Such as the patient search or gene symbol suggestions.

allasm:
This should be very easy to do as this is the way phenotype/gene suggest dialogs work in pedigree editor

buske:
Duplicate of #1490, but will close the other as it is not as thorough.

PR #1605, opened 2015-04-21
Issue #1549 Suggest dialogs should hide when they loose focus

veronikaslc:
Fixed

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -929,6 +929,18 @@ document.observe('xwiki:dom:loaded', function() {
     // Create and attach the suggests
     // -----------------------------------------------------------------
     if (typeof(PhenoTips.widgets.Suggest) != "undefined") {
+
+      // close all suggest pickers when clicking outside the picker and dialogues invoked by the picker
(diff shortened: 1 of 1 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -929,6 +929,19 @@ document.observe('xwiki:dom:loaded', function() {
     // Create and attach the suggests
     // -----------------------------------------------------------------
     if (typeof(PhenoTips.widgets.Suggest) != "undefined") {
+
+
+      // close all suggest pickers when clicking outside the picker and dialogues invoked by the picker
+      document.observe('mousedown', function(event) {
+        if (!event.findElement('.suggestItems') && !event.findElement('.dialog-ontology-browser')) {
```
antonkatz: I think you forgot to run ```mvn xar:format```

danielpgross: I think Anton is right but you can just fix it manually by replacing ```&&``` with ```&amp;&amp;```

allasm: Yeah, for small fixes xar:format screws up a diff by introducing way too many unwanted changes

Issue #1486, opened 2015-03-16
Clarify hierarchy structure of inheritance terms


buske:
It is not clear to some new users why the inheritance term options are indented and organized in this way. Having some lines or rethinking this display might help.

Reported by Veronika.

marta-:
Same issue for age of onset.

![issue2](https://cloud.githubusercontent.com/assets/651980/6923582/8e6b18f8-d7a0-11e4-8565-1c470a7889a3.png)
![issue1](https://cloud.githubusercontent.com/assets/651980/6923583/8e6e3844-d7a0-11e4-8e37-5b5e800882ad.png)

danielpgross:
@buske, are you able to provide more details about the cause of confusion here? Specific feedback from these users would be very useful.

Based on what I see now, some improvements might be to:
- provide clearer visual separation between root nodes on the hierarchy tree
- automatically check off all ancestor nodes when a leaf node is checked (this actually might apply to the tree in "Clinical symptoms and physical findings" as well)

buske:
If I recall, Veronika's suggestion was some sort of line indicator like what is used in old file system viewers:

![image](http://i.imgur.com/ZBOKk.png)

Alternatively, checking off ancestor nodes could be possible, but that doesn't fit the radio button UI, and I think it will cause more trouble than good when involving the Clinical symptoms section. Further, it requires the user to actually click something in order to understand it. I'd suggest going for a visual cue.

marta-:
Indeed, that was the suggestion. However we should not introduce a new hierarchy style, but remain consistent with the look from the ontology browser.

danielpgross:
To remain consistent with the ontology browser but still introduce a visual cue for the hierarchical nature of the information, it looks like the only option is to use the arrow symbols used by the ontology browser:
![p0000001___phenotype_database](https://cloud.githubusercontent.com/assets/4006062/7163917/c0850a76-e36a-11e4-906e-ef7f388a5824.png)

What do you think of this solution?

PR #1606, opened 2015-04-21
Issue #1486: Clarify hierarchy structure of inheritance terms

danielpgross:
Done

### DIFF
--- a/components/ontology-access/hpo/ui/src/main/resources/PhenoTips/HPOBranchDisplayer.xml
+++ b/components/ontology-access/hpo/ui/src/main/resources/PhenoTips/HPOBranchDisplayer.xml
@@ -76,7 +76,15 @@ $xwiki.jsx.use('PhenoTips.HPOBranchDisplayer')##
   #end
   #foreach($orderedItemId in $order)
     #set ($foreach.item = $macro.childrenMap.get($orderedItemId))
-    #if ("$!{foreach.item}" != '' &amp;&amp; "$!foreach.item.id" != '')&lt;li class="term-entry"&gt;#_displayItem("${foreach.item.id}", "${foreach.item.name}", $foreach.item)#_displayChildren($foreach.item.id)&lt;/li&gt;#end
+    #if ("$!{foreach.item}" != '' &amp;&amp; "$!foreach.item.id" != '')
+      #set ($childrenHtml = "#_displayChildren($foreach.item.id)")
+      #if ("$childrenHtml" != "")
+        #set ($childrenExpandSymbol = '&lt;span class="expand-tool locked">▼&lt;/span>')
+      #else
+        #set ($childrenExpandSymbol = '&lt;span class="expand-tool locked disabled">▼&lt;/span>')
+      #end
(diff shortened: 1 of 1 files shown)


danielpgross:
p0000001___phenotype_database

buske:
So much clearer!

components/ontology-access/hpo/ui/src/main/resources/PhenoTips/HPOBranchDisplayer.xml
```diff
@@ -76,7 +76,15 @@ $xwiki.jsx.use('PhenoTips.HPOBranchDisplayer')##
...
     #set ($foreach.item = $macro.childrenMap.get($orderedItemId))
-    #if ("$!{foreach.item}" != '' &amp;&amp; "$!foreach.item.id" != '')&lt;li class="term-entry"&gt;#_displayItem("${foreach.item.id}", "${foreach.item.name}", $foreach.item)#_displayChildren($foreach.item.id)&lt;/li&gt;#end
+    #if ("$!{foreach.item}" != '' &amp;&amp; "$!foreach.item.id" != '')
+      #set ($childrenHtml = "#_displayChildren($foreach.item.id)")
+      #if ("$childrenHtml" != "")
+        #set ($childrenExpandSymbol = '&lt;span class="expand-tool locked">▼&lt;/span>')
+      #else
+        #set ($childrenExpandSymbol = '&lt;span class="expand-tool locked disabled">▼&lt;/span>')
+      #end
+      &lt;li class="term-entry">$!childrenExpandSymbol#_displayItem("${foreach.item.id}", "${foreach.item.name}", $foreach.item)$!childrenHtml&lt;/li>
```
sdumitriu: `>` should also be escaped as `&gt;`

sdumitriu:
I find it a bit weird that for radio buttons `Unknown` is pulled to the left:

components/ontology-access/hpo/ui/src/main/resources/PhenoTips/HPOBranchDisplayer.xml
danielpgross: Ok, I will change it. FYI, this is the output from the xar export.

sdumitriu: Yes, but we're supposed to run `mvn xar:format ; mvn xar:format` in the module with the changes before committing. That's an XWiki annoyance, which I think might be fixed in a more recent version.

danielpgross: Ah I see, thanks.

danielpgross:
There's room to say that ```Unknown``` is different from the rest and should therefore appear separate, but for now I'll align it with the rest of the items.

danielpgross:
I see now that there already is a small ```margin-bottom``` applied to the ```Unknown``` item to separate it, and I agree that this looks much better:
p0000001___phenotype_database

Issue #1389, opened 2015-01-27
Arrow should not be shown for leaf terms in ontology browser


buske:
Expansion arrows are shown for leaf terms in the ontology browser (at least on the playground):

![image](https://cloud.githubusercontent.com/assets/4251264/5928090/fb33f810-a645-11e4-8be7-a3f221b740f9.png)

PR #1614, opened 2015-04-22
Issue #1389: "Arrow should not be shown for leaf terms in ontology browser"

veronikaslc:
Fixed


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -2464,7 +2464,7 @@ return PhenoTips;
             memo.count = this.countDescendents(this._getDataFromResponse(response));
             //Event.fire(element, 'obrowser:count:done', memo);
           }
-          if ((memo.count === 0) || (doPopulate &amp;&amp; !element.hasClassName('root') &amp;&amp; !element.down('.descendents .entry, .error'))) {
+          if ((memo.count === 0) || (doPopulate  &amp;&amp; !element.down('.descendents .entry, .error'))) {
```
sdumitriu: Theres a double space before the `&&`, replace it with only one.

PR #1638, opened 2015-05-01
Added a message into translations needed for PhenomeCentral

antonkatz:


sdumitriu:
This PR might no longer be needed, there is a new consent implementation.

Anyway, there are unaddressed comments: the translation should not contain HTML code, proper escaping must be applied where the translation is used.

Issue #1531, opened 2015-03-31
The patient JSON should include authorship information


allasm:
These include the first and the last author, and the patient record creation date and last modification date.

Currently "export multiple patients to JSON"  exports more fields than ExportPatient, since some of the fields are manually added in a velocity script ExportJSON on top of what  Patient.toJSON() generates:

...
#set ($JSON = $patient.toJSON($displayedFields))
#if ($enabledFields.contains('referrer'))
  #set ($discard = $JSON.put("referrer", ${xwiki.getUserName($patientDoc.creator, false)}))
#end
....

All of those fields should be included by the toJSON() method (assuming $displayedFields is either empty/null or has those fields, of course)

PR #1644, opened 2015-05-04
Issue #1531: "Patient toJSON() should include more fields"

veronikaslc:
Fixed

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/MetaDataController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/MetaDataController.java
@@ -47,7 +47,7 @@ import net.sf.json.JSONObject;
  * Exposes the metainfo about the document.
  *
  * @version $Id$
- * @since 1.0M10
+ * @since 1.2RC1
  */
 @Component(roles = { PatientDataController.class })
 @Named("metadata")
@@ -85,7 +85,7 @@ public class MetaDataController extends AbstractSimpleController implements Pati
 
             Map<String, String> result = new LinkedHashMap<String, String>();
 
-            SimpleDateFormat format = new SimpleDateFormat("yyyy.MM.dd");
+            SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
 
             result.put(DOCUMENT_NAME, doc.getDocumentReference().getName());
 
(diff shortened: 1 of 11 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/RejectedGeneListController.java
```diff
@@ -155,13 +163,16 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
         while (iterator.hasNext()) {
             Map<String, String> item = iterator.next();
 
-            if (StringUtils.isBlank(item.get(COMMENTS_KEY))
-                || (selectedFieldNames != null
+            if (!StringUtils.isBlank(item.get(GENE_KEY))) {
```
sdumitriu: This is a different issue.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/MetaDataController.java
```diff
@@ -0,0 +1,164 @@
...
+
+import com.xpn.xwiki.doc.XWikiDocument;
+
+import net.sf.json.JSONObject;
+
+/**
+ * Exposes the metainfo about the document.
+ *
+ * @version $Id$
+ * @since 1.0M10
```
sdumitriu: Wrong `@since` version.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/MetaDataController.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    public PatientData<String> load(Patient patient)
+    {
+        try {
+            XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
+
+            Map<String, String> result = new LinkedHashMap<String, String>();
+
+            SimpleDateFormat format = new SimpleDateFormat("yyyy.MM.dd");
```
sdumitriu: Don't use this custom format, use the standard ISO date-time one.

veronikaslc: I took this custom date format from DataToCellConverter.java, where it is used for the same fields. Should we also change format there to standard ISO date-time just to be consistent?

sdumitriu:
@allasm Indeed, you're right. Still, the ISO date format should be used instead of that dotted thing.

PR #1650, opened 2015-05-07
Issue #1649: Segmented bar should be a widget

mjshepherd:
Put the code in widgets.xml, wasn't sure where it should go.


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+          displayPercentValueFlag : false
+        };
+        container.insert(new PhenoTips.widgets.SegmentedBar(score, barOptions).generateSegmentedBar());
+
+    @param value The value that this bar is meant to display. Must be between 0 and 1.
+
+    @param options Options for styling the segmented bar, all values are optional:
+
+    {
+      segments: The number of segments in the bar. Default is 5.,
```
sdumitriu: The `.,` at the end looks weird, choose one.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+      fill: The fill color for the segments. Default is #777.,
+      border: The border color of the segments. Default is #000.,
```
sdumitriu: `000` or `777`? The code says the latter.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+      segmentHeight: The height of each individual segment. Default is 5px.,
```
sdumitriu: Be more specific. Does it take a number (which is treated as px), only a px value, or any css value? I wouldn't know what to pass based only on this description.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+        };
+        container.insert(new PhenoTips.widgets.SegmentedBar(score, barOptions).generateSegmentedBar());
+
+    @param value The value that this bar is meant to display. Must be between 0 and 1.
+
+    @param options Options for styling the segmented bar, all values are optional:
+
+    {
+      segments: The number of segments in the bar. Default is 5.,
+      fill: The fill color for the segments. Default is #777.,
```
sdumitriu: Are only `#abc` values allowed, or any CSS color specification that browsers support, like `hsla(120, 100%, 50%, 1)`?

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+      displayPercentValueFlag: Option to display the percentage value after the bar. Default is true
+    }
+
+    */
+    initialize : function(value, options) {
+      if (value &gt; 1 || value &lt; 0) {
+        console.log("Invalid segmented bar value");
+      }
+      this.value = value;
+      this.options = options || {};
```
sdumitriu: Here's a shorter way of doing this: https://github.com/mjshepherd/phenotips/blob/Issue-%231649/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml#L2323-L2360

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+      border: The border color of the segments. Default is #000.,
+      segmentHeight: The height of each individual segment. Default is 5px.,
+      segmentWidth: The width of each individual segment. Default is equal to either the specified height or the default height.,
+      displayPercentValueFlag: Option to display the percentage value after the bar. Default is true
+    }
+
+    */
+    initialize : function(value, options) {
+      if (value &gt; 1 || value &lt; 0) {
+        console.log("Invalid segmented bar value");
```
sdumitriu: And if it's invalid, we just ignore this problem and continue displaying it?

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+      this.fill = this.options.fill || this.DEFAULT_FILL;
+      this.border = this.options.border || this.DEFAULT_BORDER;
+      this.segmentHeight = this.options.segmentHeight || this.DEFAULT_SEGMENT_HEIGHT;
+      this.segmentWidth = (this.options.segmentWidth || this.options.segmentHeight) || this.DEFAULT_SEGMENT_HEIGHT;
+      this.displayPercentValueFlag = (this.options.displayPercentValue != undefined) ? this.options.displayPercentValue : this.DEFAULT_DISPLAY_PERCENT_VALUE_FLAG;
+    },
+    generateSegmentedBar: function() {
+      var bar = new Element('div', {
+        'class' : 'segmented-bar',
+        'style' : 'color: ' + this.border + ';',
```
sdumitriu: Instead of setting the style inline as an attribute, you can call [`setStyle({...})`](http://api.prototypejs.org/dom/Element/prototype/setStyle/), which has the advantage that it accepts an explicit map instead of one string.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+    </class>
+    <property>
+      <cache>default</cache>
+    </property>
+    <property>
+      <code>var PhenoTips = (function(PhenoTips) {
+  var widgets = PhenoTips.widgets = PhenoTips.widgets || {};
+  widgets.SegmentedBar = Class.create({
+    DEFAULT_SEGMENTS : 5,
+    DEFAULT_FILL : '#777',
```
sdumitriu: I wonder if [`currentColor`](http://www.w3.org/TR/css3-color/#currentcolor) would work as a better default. It might make it work correctly without the need to specify an explicit color.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+        'class' : 'segmented-bar',
+        'style' : 'color: ' + this.border + ';',
+        'title': Math.round(this.value * 100) + '%' || ''
+      });
+      var valueUnit = 1 / this.segments;
+      for (var i = 0; i &lt; this.segments; ++i) {
+        var segmentFill = Math.round(100 * Math.min(Math.max((this.value - i * valueUnit) / valueUnit, 0), 1));
+        var segment = new Element('span', {
+          'class': 'segmented-unit',
+          'style': 'height: ' + this.segmentHeight +';' +
```
sdumitriu: `setStyle` here as well, and further down.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+                    'width: ' + this.segmentWidth + ';' +
+                    'border-color:' + this.border + ';'
+        }).update(
```
sdumitriu: It is possible to do this without the need for an inner span.

If we don't separate between the border and fill colors, it's enough to set border-left-width to 1+box width * segmentFill.

If we do want different border and fill colors, then this will work:

```
.setStyle({
  backgroundClip: 'content-box', // This goes in the CSS below
  boxSizing: 'border-box', // This goes in the CSS below
  padding: 0, // This goes in the CSS below
  backgroundColor: this.fill,
  borderColor: this.border,
  paddingRight: (1-segmentFill) * this.segmentWidth,
  height: this.segmentHeight + 1 + 'px', // Assuming the width is just a value without a unit, otherwise use:
  width: 'calc(' + this.segmentWidth + '+1px)'
});

File: components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml [hunk 11]
```diff
@@ -4693,6 +4693,177 @@ document.observe("xwiki:dom:loaded", function() {
...
+})(PhenoTips || {});</code>
+    </property>
+    <property>
+      <name>Segmented Bar</name>
+    </property>
+    <property>
+      <parse/>
+    </property>
+    <property>
+      <use>currentPage</use>
```
sdumitriu: `onDemand`

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -6116,4 +6287,109 @@ input.xwiki-free-multiselect-value {
...
+}</code>
+    </property>
+    <property>
+      <name>Segmented Bar</name>
+    </property>
+    <property>
+      <parse/>
+    </property>
+    <property>
+      <use>currentPage</use>
```
sdumitriu: `onDemand`

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
mjshepherd: I don't understand how this is shorter. Since each option is independently optional, each key must be checked individually.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
mjshepherd: why `+1px` to all lengths?

sdumitriu: `box-sizing: border-box` means that the `width` includes the padding and border as well. Actually that should be `+2px`... Anyway, since we're using the padding as the fill, a 100% fill plus the border should fit in the whole box. Either we say that the desired box width is the interior size, in which case we add the border width to the `width` property, or we say that the desired box width is the exterior size, in which case we subtract the border width to get the maximum fill width. I think the first option is better.

mjshepherd: This is causing me some issues in chrome.  It looks fine normally but doesn't display properly at certain zoom levels

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4693,6 +4693,182 @@ document.observe("xwiki:dom:loaded", function() {
...
+    */
+    initialize : function(value, options) {
+      this.value = value;
+      this.options = options || {};
+      this.segments = this.options.segments || this.DEFAULT_SEGMENTS;
+      this.fill = this.options.fill || this.DEFAULT_FILL;
+      this.border = this.options.border || this.DEFAULT_BORDER;
+      this.segmentHeight = this.options.segmentHeight || this.DEFAULT_SEGMENT_HEIGHT;
+      this.segmentWidth = (this.options.segmentWidth || this.options.segmentHeight) || this.DEFAULT_SEGMENT_HEIGHT;
+      this.displayPercentValueFlag = (this.options.displayPercentValue != undefined) ? this.options.displayPercentValue : this.DEFAULT_DISPLAY_PERCENT_VALUE_FLAG;
```
meatcar: Do comparisons with undefined using 
```js
typeof displayPercentValue !== 'undefined'
```

This way, if someone accidentaly defined `var undefined = "blah"`, it won't mess with your code.

meatcar:
:+1:

PR #1672, opened 2015-05-21
Issue #1649: minor fix

mjshepherd:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -6354,7 +6354,7 @@ input.xwiki-free-multiselect-value {
   width: auto;
 }
 .segmented-bar .segmented-unit {
-  border: 0.1ex solid currentColor;
+  border: 0.2ex solid currentColor;
```
sdumitriu: This doesn't really fix the issue. The border will still disappear when decreasing font size, and now it's too thick. I guess `1px` is better.

Don't forget to also update the `border-radius` on the next line.

Issue #1678, opened 2015-05-26
Broken Hide button is displayed for growth charts in view mode


tmarathe:
While viewing a patient record with a growth chart the 'Hide' button for the growth chart produces no action.

sdumitriu:
That button should be removed, in view mode everything is displayed

PR #1684, opened 2015-05-27
Issue #1678: Broken Hide button is displayed for growth charts in vie…

mjshepherd:
…w mode

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -3879,6 +3879,18 @@ h1, h2, h3, h4 {
 }
 /** --------------------------------------------- **/
 /** Charts **/
+.growth-charts-section.collapsed #charts,
+.growth-charts-section.collapsed .expand-tools .hide,
+.growth-charts-section .expand-tools .show {
+  display: none;
+}
+.growth-charts-section.collapsed .expand-tools .show {
+  display: inline;
+}
+.growth-charts-section h3.expandable {
+  cursor: pointer;
+}
+
 .chart-wrapper {
   box-shadow: 0 0 6px $theme.borderColor;
   display:inline-block;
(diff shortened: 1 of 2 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -4089,17 +4101,6 @@ form .extradata-list tr:hover td {
   text-align: center;
   text-decoration: none !important;
 }
-.growth-charts-section.collapsed #charts,
```
sdumitriu: Why did you move this?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__charts.xml
```diff
@@ -116,9 +116,13 @@
 #set ($targetClassName = 'PhenoTips.MeasurementsClass')
 #if ($xcontext.action != 'export' &amp;&amp; ($xcontext.action == 'edit' || $doc.getObject($targetClassName)))
   (% class="growth-charts-section" %)(((
-    === $services.localization.render('phenotips.UIXField.growthCharts.heading') ===
-
-    #__expandTools()
+    #if($xcontext.action != 'view')
+      (% class="expandable" %)
+      === $services.localization.render('phenotips.UIXField.growthCharts.heading') ===
```
sdumitriu: Since this heading is present in both branches of the `#if`, why not put it only once before the test?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__charts.xml
```diff
@@ -116,9 +116,13 @@
 #set ($targetClassName = 'PhenoTips.MeasurementsClass')
 #if ($xcontext.action != 'export' &amp;&amp; ($xcontext.action == 'edit' || $doc.getObject($targetClassName)))
   (% class="growth-charts-section" %)(((
-    === $services.localization.render('phenotips.UIXField.growthCharts.heading') ===
-
-    #__expandTools()
+    #if($xcontext.action != 'view')
```
sdumitriu: That's not correct, there are more options than `view` and `edit`, and the only one where it makes sense to have it is in edit mode. For example, there's also `export`, and we don't want to have the `hide` button in the PDF.

Also, leave a space in `#if (`.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__charts.xml
```diff
@@ -116,9 +116,13 @@
...
+      (% class="expandable" %)
```
sdumitriu: I see why you put the heading in two places, you have optional content both before and after it. I guess it's fine to put the heading in both branches of the `#if` then.

PR #1686, opened 2015-05-27
Prioritize patient specificity scorers

danielpgross:


components/specificity-meter/api/src/main/java/org/phenotips/data/internal/MonarchPatientScorer.java
```diff
@@ -165,6 +165,10 @@ public double getScore(Patient patient)
         return -1;
     }
 
+    public int getScorerPriority() {
+        return 2;
```
sdumitriu: I would space these out, to allow something to come between monarch and omimInformationContent. Let's say 50 for this one, 10 for omimInformationContent.

components/specificity-meter/api/src/main/java/org/phenotips/data/internal/PatientScorerComparator.java
```diff
@@ -0,0 +1,35 @@
+package org.phenotips.data.internal;
```
sdumitriu: Missing license header.

components/specificity-meter/api/src/main/java/org/phenotips/data/PatientScorer.java
```diff
@@ -45,4 +45,11 @@
      * @return a score between {@code 0} and {@code 1}, or {@code -1} if the score cannot be computed by this scorer
      */
     double getScore(Patient patient);
+
+    /**
+     * Get the priority of this scorer. Higher priority (larger integer) scorers will be used first.
+     *
+     * @return a score greater than {@code 0}
```
sdumitriu: Since this is a new method, it should have a `@since`.

danielpgross:
@sdumitriu are you still planning to fix the test issues here?

sdumitriu:
Yes, and I know how, I just didn't get around to it.

Or, if you want to do it yourself, see [the changes in this file](https://github.com/phenotips/phenotips/commit/9788856895395b457835b757e6de8ead837fc00c#diff-fd854e7f4b49f87fb844f05f0c4efb1a).

danielpgross:
Ok, thanks. In that case, I'll leave it to you -- I just wanted to make sure we're on the same page.

danielpgross:
This thread is no longer being worked on.

Issue #1677, opened 2015-05-25
Case solved info button text is broken


buske:
- [ ] Info button only shows up after "Case solved" is selected, but should always show up
- [ ] Info text currently displays "PhenoTips.PatientClass_solvedCase_hint" instead of actual hint text

PR #1694, opened 2015-05-28
Issue #1677: Case solved info button text is broken

mjshepherd:


components/patient-data/impl/src/main/resources/ApplicationResources.properties
```diff
@@ -14,6 +14,8 @@ PhenoTips.PatientClass_omim_id_hint=Generally, a disorder is used to annotate th
...
 PhenoTips.InvestigationClass_hint=List genes considered as candidates for the patient's disease.
 PhenoTips.RejectedGenesClass_hint=List genes that have been investigated and ruled out as candidates for the patient's disease.
+PhenoTips.PatientClass_solvedCase_hint=Check this box if the molecular diagnosis for this patient is known. Pubmed ID may be entered for any published results as well as the solved gene and resolution notes.
```
sdumitriu: I'm not sure `solved gene` is the right term.

buske: "Check this box is the molecular diagnosis for the patient is known."
Leave the rest off. We can add a hint for those fields later, if we'd like, but the solved gene field will be gone soon anyway.

PR #1697, opened 2015-05-29
Issue 1662 - Added a check that death date is not before birth date

itaiGershtansky:


components/patient-data/impl/src/main/resources/ApplicationResources.properties
```diff
@@ -1,232 +1,233 @@
...
-PhenoTips.PatientConsent_genetic=I confirm that consent has been obtained to share this patient's genetic data on restricted access databases.
-PhenoTips.PatientConsent_share_history=I confirm that consent has been obtained to share this patient's medical and family history on restricted access databases.
-PhenoTips.PatientConsent_share_images=I confirm that consent has been obtained to share this patient's medical images/photos on restricted access databases.
-PhenoTips.PatientConsent_matching=This patient is matchable through the MatchmakerExchange and you may receive occasional match notifications (unless you set the visibility to Private).
```
allasm: Hmm, why is git showing so many changes? Did you use other line endings? This doesn't look like a good idea as it is really hard to check what really has been changed....

PR #1699, opened 2015-05-29
Tests for MeasurementAgeUpdater

DeanWay:


components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,147 @@
...
+        MockitoAnnotations.initMocks(this);
+        when(this.source.getXObject((EntityReference) Matchers.any())).thenReturn(this.patientRecordObj);
+
+        Field age = ReflectionUtils.getField(MeasurementAgeUpdater.class, "AGE_PROPERTY_NAME");
+        Field date = ReflectionUtils.getField(MeasurementAgeUpdater.class, "DATE_PROPERTY_NAME");
+
+        age.setAccessible(true);
+        date.setAccessible(true);
+
+        this.DATE_PROPERTY_NAME = (String)date.get(null);
```
sdumitriu: That's really convoluted. A better test would be to just define it here again, and check that the same string is indeed used.

PR #1717, opened 2015-06-02
Issue 1662

itaiGershtansky:
Added a validity check that death date is not before birth date

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -353,10 +353,73 @@ document.observe('xwiki:dom:loaded', function() {
   return XWiki;
 }(XWiki || {}));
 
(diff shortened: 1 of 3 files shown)


sdumitriu:
Very nice pull request.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -352,6 +352,101 @@ document.observe('xwiki:dom:loaded', function() {
...
+   if (birth.year &gt; death.year) {
+      return false;
+   }
+   if (used_decade) {// here birth.year==death.year
+     return true;
+   }
+
+   //years are the same, compare months
+   if (typeof birth.month == "undefined" || typeof death.month == "undefined") {
+      //dates cannot be compared
```
sdumitriu: We should assume fair play, if the year is the same but one of the dates doesn't have better precision, then we can say the dates are compatible. Same for missing days.

PR #1724, opened 2015-06-04
Issue 1722

tmarathe:


mjshepherd:
The issue will still appear on page load and when viewing the patient document. You'll also need to edit phenotips/components/patient-measurements/ui/src/main/resources/PhenoTips/MeasurementsClass.xml

allasm:
Instead of repeating the same code 10+ times maybe we can make a service for number suffixes and use that (in a way similar to using translations, which would translate 20 to "20th", 21 to "21st" etc)?

sdumitriu:
@allasm Yes, that's a good idea. Plus, it needs to take into account the language, since in any other language the suffixes won't be right.

PR #1725, opened 2015-06-04
PDF/JSON is opened in a new tab

itaiGershtansky:


tmarathe:
This works while exporting from the edit page, but the issue persists when exporting from the patient document page. Additionally, the new window gets blocked as a popup in Firefox 38.0.5 and Chrome.

itaiGershtansky:
1. Fixed exporting from patient document page: https://github.com/phenotips/phenotips/pull/1728
2. If the browser blocks pop-ups, I don't think we want (and can?) override this.

sdumitriu:
For 2, try this: http://stackoverflow.com/a/19666946/620249

Issue #1452, opened 2015-03-03
Rename "Report Name" to "External Identifier" in the data tables


sdumitriu:

buske:
internal id -> "Identifier"
external id -> "External identifier"

PR #1727, opened 2015-06-04
issue #1452: Swap 'Report Name' and 'Identifier' labels for the data …

itaiGershtansky:
…tables

Done


sdumitriu:
Don't swap the order of the columns, just their labels.

PR #1732, opened 2015-06-05
Add build step for generating files required by local Monarch scorer

danielpgross:
Work in progress


danielpgross:
This thread is no longer being worked on.

Issue #1658, opened 2015-05-09
Add ICSI (intra-cytoplasmic sperm injection) as an alternative to IVF


sdumitriu:

itaiGershtansky:
Fixed together with issue #1655, in pull request #1734

PR #1734, opened 2015-06-05
Issue #1655: New 'Assisted reproduction' questions: donor sperm or egg

itaiGershtansky:
Question added. Property added to patient data export


sdumitriu:
The checkstyle fails.

sdumitriu:
These should actually be two questions: _Donor sperm_ and _Donor egg_.

itaiGershtansky:
Fixed issue #1658 as well

PR #1737, opened 2015-06-08
DefaultPushServerGetPatientIDResponse Tests

JKereliuk:


components/patient-data-sharing/push-api/pom.xml
```diff
@@ -117,4 +117,7 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <properties>
+    <coverage.instructionRatio>0</coverage.instructionRatio>
```
sdumitriu: `0` is not good.

JKereliuk: I have few pull requests open for tests in push-api and I had to keep changing it to make it build properly. This will be changed if tests are merged into master based on what other tests are accepted.

PR #1738, opened 2015-06-08
Develop

jakennedy:
Hi Sergiu, et. al., 

I am creating this pull request so you can look at the changes that were made to support storing attachments in S3 for Harvard's UDN project. We reused an existing XWiki contrib module [xwiki-platform-cloud](https://github.com/xwiki-contrib/xwiki-platform-cloud) and made some changes to make it work with 5.4.x. 

The module was added as a subtree to our fork of the Phenotips project because we are only building one component. However, for full integration it would be better if the module itself was forked, improved, and added as a Phenotips dependency in Maven. 

We would like the Phenotips team to add cloud integration to the project so we don't have to maintain this fork going forward.


davidbernick:
You can ignore the /deploy dir as that's for our CI/CD environments.

jakennedy:
Latest changes should resolve all issues.

davidbernick:
@sdumitriu can we get feedback on whether this will be acceptable?

marta-:
Yes. What will happen is:
- we'll squash the changes to reflect cleanly what was modified wrt the code developed at xwiki-contrib
- we will commit the code to xwiki-contrib so that the fix be there in the original code
- we will then fork it, mentain it on our side and contribute back any relevant fixes and developments to stay in sync with the origin

davidbernick:
Terrific. Any timeline?

PR #1745, opened 2015-06-09
MeasurementAgeUpdater test

DeanWay:


DeanWay:
Nominating @JKereliuk and @mjshepherd for review

components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Before
+    public void setUp() throws NoSuchFieldException, IllegalAccessException
+    {
+        MockitoAnnotations.initMocks(this);
+        when(this.source.getXObject((EntityReference) Matchers.any())).thenReturn(this.patientRecordObj);
+
+        objects = new LinkedList<>();
+        objects.add(measurement);
+        when(this.source.getXObjects((EntityReference)Matchers.any())).thenReturn(objects);
```
mjshepherd: Matcher.<EntityReference>any()

components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Date birthDate = dateFormat.parse(birthDateString);
+        Date measureDate = dateFormat.parse(measureDateString);
+
+        when(this.patientRecordObj.getDateValue(Matchers.matches("date_of_birth"))).thenReturn(birthDate);
+        when(this.measurement.getDateValue(DATE_PROPERTY_NAME)).thenReturn(measureDate);
+
+        this.mocker.getComponentUnderTest().onEvent(event, source, data);
+
+        verify(this.measurement).setFloatValue(eq(AGE_PROPERTY_NAME), eq(365/30.4375f));
+    }
```
mjshepherd: Maybe add tests for multiple measurement objects and no objects. Won't increase coverage but enforces expected behaviour

JKereliuk: To continue on this, do that and verify that certain methods are called multiple times with the correct arguments.

mjshepherd:
More dates are good. But you should also check with multiple measurement objects. The MeasurementAgeUpdater retrieves all the measurement objects from the record and updates them all.

DeanWay:
@mjshepherd Added tests for multiple and zero measurement objects

mjshepherd:
+1

JKereliuk:
+1 @sdumitriu your turn!

components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Tests for the {@link MeasurementAgeUpdater} component.
+ *
+ */
+public class MeasurementAgeUpdaterTest
+{
+
+    @Rule
+    public MockitoComponentMockingRule<MeasurementAgeUpdater> mocker =
```
sdumitriu: The parameter should be the interface with the `@Role` annotation, in this case `EventListener`:

    public MockitoComponentMockingRule<EventListener> mocker =
        new MockitoComponentMockingRule<EventListener>(MeasurementAgeUpdater.class);

components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,186 @@
...
+            new MockitoComponentMockingRule<>(MeasurementAgeUpdater.class);
```
sdumitriu: Our [codestyle](https://github.com/phenotips/build-tools/tree/master/codestyle-resources/src/main/resources) is to use one level of indentation. Although we don't enforce it with checkstyle for test code, it should still be respected as much as possible.

components/patient-measurements/api/src/test/java/org/phenotips/measurements/internal/MeasurementAgeUpdaterTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @Test
+    public void returnsNormallyWhenGivenNoMeasurementObjects() throws Exception {
+        when(this.source.getXObjects(any(EntityReference.class))).thenReturn(null);
+        NullPointerException exceptionWhenListIsNull = null;
+        try {
+            this.mocker.getComponentUnderTest().onEvent(event, source, data);
+        } catch(NullPointerException e){
+            exceptionWhenListIsNull = e;
```
sdumitriu: You don't need to do this. If there is an exception thrown, the test will automatically fail.

If, however, you do want to test that calling a method will throw an exception, then you would annotate the method with:

    @Test(expected=NullPointerException.class)

sdumitriu:
You can install EclEmma to check which lines are covered by tests and which aren't. I think you're missing the case when the event's source doesn't have a patient object at all.

DeanWay:
@sdumitriu test coverage is now %100

Issue #1442, opened 2015-02-24
Record locking


marta-:

buske:
Allow doctors to "lock" a record to prevent any further editing (from them or any other user) until they "unlock" it again.

meatcar:
@DeanWay @JKereliuk @tmarathe have you guys tested this, is it working properly?

JKereliuk:
I tested this a while ago and it appeared to be working correctly.

tmarathe:
I tested record locking also, it is working correctly.

PR #1749, opened 2015-06-09
Issue #1442: Patient records can now be locked

mjshepherd:


mjshepherd:
@JKereliuk @tmarathe @DeanWay This is going to need some serious testing before merging. Adding a lock essentially removes edit rights from all users. It may interfere with other features. Also I dropped the coverage for patient-access-rules by a bit...

mjshepherd:
@antonkatz @veronikaslc Will you review?

antonkatz:
Yup. Tomorrow.

components/security/authorization/pom.xml
```diff
@@ -47,6 +47,18 @@
       <artifactId>xwiki-platform-users-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>phenotips-constants</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
```
antonkatz: Where is this guy used?

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param patientId The record to be locked
+     * @param lock "true" to create a lock, "false" to remove
+     */
+    public void setRecordLock (String patientId, Boolean lock) throws XWikiException
+    {
+        if (patientId == null || lock == null) {
+            return;
+        }
+
+        XWikiContext context = (XWikiContext) this.execution.getContext().getProperty("xwikicontext");
```
antonkatz: Sergiu might comment that this should be done through injecting `Provider<Context>` (if my memory serves me right).

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,108 @@
...
+    /** Allows checking of access rights on a patient*/
+    @Inject
+    PermissionsManager pm;
+
+    /**
+     * Sets or removes a lock from a record. Locks remove edit rights from all users.
+     * @param patientId The record to be locked
+     * @param lock "true" to create a lock, "false" to remove
+     */
+    public void setRecordLock (String patientId, Boolean lock) throws XWikiException
```
antonkatz: Not a good idea for a script service to throw exceptions. It'll go straight to the user.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,387 @@
...
+  <version>1.1</version>
+  <title>Patient Lock</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>false</hidden>
+  <content>{{velocity}}
+#set ($patientId = $!{request.getParameter('patientId')})##The patient id to be locked or unlocked
+#set ($lockingRequest = $!{request.getParameter('lock')})##The locking request. True to lock, false to unlock.
+#if ("$!{patientId}" != ' ' &amp;&amp; "$!{lockingRequest}" != ' ')
```
antonkatz: Why `' '` instead of `''`

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,387 @@
...
+
+    button.observe('click', function(event) {
+      event.stop();
+      (!activeLockingRequest &amp;&amp; new Ajax.Request(
+        button.href, {
+          method: "POST",
+          onCreate: function() {
+            activeLockingRequest = true;
+          },
+          onComplete: function() {
```
antonkatz: Does the user need to know if this fails?

components/security/authorization/src/main/java/org/phenotips/security/authorization/internal/LockedAuthorizationModule.java
```diff
@@ -0,0 +1,76 @@
...
+
+        try {
+            XWikiDocument doc = context.getWiki().getDocument(document, context);
+            BaseObject lock = doc.getXObject(this.lockClassReference);
+            if (lock != null && !access.isReadOnly()) {
+                return false;
+            }
+        } catch (XWikiException e) {
+            return null;
+        } catch (NullPointerException e) {
```
antonkatz: You can put the two together ```(XWikiException | NullPointerException e)```.

components/security/authorization/pom.xml
mjshepherd: My test was failing because it couldn't find the class. I'm honestly not sure why I had to include the dependency even though I never explicitly use it.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
mjshepherd: I was thinking about that. If the locking/unlocking fails then it will be noticeable. I could alternatively alert the user with a red notification but it would probably say something like "Failed to lock/unlock the record" which is marginally better I suppose. What do you think?

antonkatz: Hm... seems like this functionality is important enough that the user should not walk away thinking that the record is unmodifiable when it still is.

mjshepherd: Fair, will add

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+  <title>Patient Lock</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>false</hidden>
+  <content>{{velocity}}
+#set ($patientId = $!{request.getParameter('patientId')})##The patient id to be locked or unlocked
+#set ($lockingRequest = $!{request.getParameter('lock')})##The locking request. True to lock, false to unlock.
+#set ($resultSuccessful = false)
+#if ("$!{patientId}" != ' ' &amp;&amp; "$!{lockingRequest}" != ' ')
```
antonkatz: These guys are back. I saw you commit the changes though.

components/security/authorization/pom.xml
antonkatz: Strange. I think these are REST dependencies.

antonkatz:
+1, except for the spaces in the velocity code.

veronikaslc:
Joining with +1 as well.

antonkatz:
@sdumitriu the ball's in your court :)

JKereliuk:
I created a new record and locked it, then unlocked it and made it open. After creating a new user and trying to edit it I got the following message 
forceedit

After clicking force edit I could still edit but it was a strange message since it was indeed unlocked.

antonkatz:
That's a different kind of lock. It means that the document was being
edited by 'a b', and it was never registered that the 'a b' was done
editing the document. It's a concurrent edit lock.

components/security/authorization/pom.xml
sdumitriu: This is a [runtime dependency of the `XWiki` class](https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-7.0.1/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java#L64), and since it is marked as a `provided` dependency of oldcore, it is not transitively included in the classpath.

components/patient-access-rules/api/pom.xml
```diff
@@ -30,7 +30,7 @@
 
   <properties>
     <checkstyle.skip>true</checkstyle.skip>
-    <coverage.instructionRatio>0.69</coverage.instructionRatio>
+    <coverage.instructionRatio>0.67</coverage.instructionRatio>
```
sdumitriu: It's really frowned upon to decrease the coverage. If there's new code, it should have new tests.

Anyway, all this code shouldn't be moved into its own module, in `components/record-locking`, complete with `api`, `impl`, `ui`, `pageobjects` and `tests` submodules.

mjshepherd: Should?

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,115 @@
...
+    @Inject
+    PermissionsManager pm;
+
+    /**
+     * Sets or removes a lock from a record. Locks remove edit rights from all users.
+     * @param patientId The record to be locked
+     * @param lock "true" to create a lock, "false" to remove
+     * @return true if successful false if otherwise
+     */
+    public Boolean setRecordLock (String patientId, Boolean lock)
```
sdumitriu: I would split this into two different methods, `lock` and `unlock`, and they should receive a `Patient` object instead of a string.

sdumitriu: If you don't explicitly allow `null` for a `Boolean` parameter or return value, then use the primitive `boolean` type instead.

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (patientId == null || lock == null) {
+            return Boolean.FALSE;
+        }
+
+        XWikiContext context = (XWikiContext) this.execution.getContext().getProperty("xwikicontext");
```
sdumitriu: Use `Provider<XWikiContext>` instead.

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,115 @@
...
+        Patient patient = this.pr.getPatientById(patientId);
+        if (patient == null) {
+            return Boolean.FALSE;
+        }
+
+        try {
+            DocumentReference patientDocumentReference = patient.getDocument();
+            XWikiDocument patientDocument = xwiki.getDocument(patientDocumentReference, context);
+            PatientAccess patientAccess = this.pm.getPatientAccess(patient);
+            Boolean hasLockingPermission = patientAccess.hasAccessLevel(new ManageAccessLevel());
```
sdumitriu: Don't instantiate components, `@Inject @Named('manage') private AccessLevel requiredAccess` an instance instead.

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,115 @@
...
+            if (hasLockingPermission) {
+                BaseObject previousLockObject = patientDocument.getXObject(lockClassReference);
+
+                if (lock && previousLockObject == null) {
+                    patientDocument.createXObject(lockClassReference, context);
+                } else if (!lock && previousLockObject != null) {
+                    patientDocument.removeXObjects(lockClassReference);
+                }
+
+                xwiki.saveDocument(patientDocument, context);
```
sdumitriu: Add a comment and set this as a minor edit:

    xwiki.saveDocument(patientDocument, (lock ? "Locked" : "Unlocked") + " patient record", true, context);

components/patient-access-rules/api/pom.xml
sdumitriu: Yes, I started a different phrase and changed it.

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/RecordLockingService.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * A service to add or remove a lock object to a record. The lock object will remove edit rights for all users on the
+ * document.
+ * @version $Id$
+ * @since 1.3
+ */
+@Component
+@Named("recordLocking")
+@Singleton
+public class RecordLockingService implements ScriptService
```
sdumitriu: A script service should not contain real code, except for rights checking. Add a real component for this.

A script service is not easy to use from other Java code, that's why a real component is needed.

mjshepherd: No problem.  But regarding your comment above about the script service having two methods and taking a patient object. If I implement that then there will need to be velocity code which processes the string representation of the patient before calling the script service. Basically there will just be more velocity code. Wouldn't it be better to have the script service deal with the raw request parameter values rather than putting that logic in velocity?
The component will have two methods.

sdumitriu: You can add two sets of methods in the script service, one that receives a string and one that receives a `Patient`. The real component should only have the ones with `Patient`.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+#set ($lockObject = $doc.getObject($lockDocName))
+#set ($patient = $services.patients.getPatientById($doc.documentReference))
+#set ($patientAccess = $services.permissions.getPatientAccess($patient))
+#set ($patientLockingService = $xwiki.getDocument($lockDocName))
+#set ($lockingServiceRequestParams = "patientId=" + $patient.getId())
+{{html}}
+##Macro used to create help button.
+#macro (_createLockButton $link $icon $text)&lt;span class='buttonwrapper'&gt;&lt;a id="record-locking-button" class="button" href="${link}"&gt;&lt;span class = "fa fa-${icon}"&gt;&lt;/span&gt; ${text}&lt;/a&gt; &lt;span class="fa fa-question-circle xHelpButton" title="$services.localization.render('phenotips.patientLock.action.hint')"&gt;&lt;/span&gt;&lt;/span&gt;#end
+&lt;div class="locking-container"&gt;
+#if ("$!{lockObject}" == '')
```
sdumitriu: The logic for detecting if a patient record is locked or not shouldn't be in velocity. Add an `boolean isLocked(patient)` method in the API.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+$xwiki.ssx.use('PhenoTips.PatientLock')##
+#set ($lockDocName = 'PhenoTips.PatientLock')
+#set ($lockObject = $doc.getObject($lockDocName))
+#set ($patient = $services.patients.getPatientById($doc.documentReference))
+#set ($patientAccess = $services.permissions.getPatientAccess($patient))
+#set ($patientLockingService = $xwiki.getDocument($lockDocName))
+#set ($lockingServiceRequestParams = "patientId=" + $patient.getId())
+{{html}}
+##Macro used to create help button.
+#macro (_createLockButton $link $icon $text)&lt;span class='buttonwrapper'&gt;&lt;a id="record-locking-button" class="button" href="${link}"&gt;&lt;span class = "fa fa-${icon}"&gt;&lt;/span&gt; ${text}&lt;/a&gt; &lt;span class="fa fa-question-circle xHelpButton" title="$services.localization.render('phenotips.patientLock.action.hint')"&gt;&lt;/span&gt;&lt;/span&gt;#end
```
sdumitriu: Every operation that alters data should be protected against CSRF.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+    </property>
+    <property>
+      <code>(function () {
+  var init = function() {
+    var button = $('record-locking-button');
+    var activeLockingRequest = false; //Prevent multiple requests being active concurrently
+
+    button.observe('click', function(event) {
+      event.stop();
+      (!activeLockingRequest &amp;&amp; new Ajax.Request(
```
sdumitriu: If all you're doing is making a request then reloading, why don't you just let the link be followed?

However, since this is a dangerous operation, it should be confirmed.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+#set ($patientLockingService = $xwiki.getDocument($lockDocName))
+#set ($lockingServiceRequestParams = "patientId=" + $patient.getId())
+{{html}}
+##Macro used to create help button.
+#macro (_createLockButton $link $icon $text)&lt;span class='buttonwrapper'&gt;&lt;a id="record-locking-button" class="button" href="${link}"&gt;&lt;span class = "fa fa-${icon}"&gt;&lt;/span&gt; ${text}&lt;/a&gt; &lt;span class="fa fa-question-circle xHelpButton" title="$services.localization.render('phenotips.patientLock.action.hint')"&gt;&lt;/span&gt;&lt;/span&gt;#end
+&lt;div class="locking-container"&gt;
+#if ("$!{lockObject}" == '')
+  #if ($patientAccess.hasAccessLevel('manage'))
+    #set ($lockingServiceRequestParams = $lockingServiceRequestParams + "&amp;lock=true")
+    #_createLockButton("$patientLockingService.getURL('view', $lockingServiceRequestParams)", 'lock', "$services.localization.render('phenotips.patientLock.action.lock')")
```
sdumitriu: Instead of leading out to a different document, I would use this as a sheet for the patient record:

    $doc.getURL('view', "sheet=${lockDocName}")

sdumitriu: Then you don't need to pass the patient ID as a parameter, just use `$doc`.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+#set ($patientId = $!{request.getParameter('patientId')})##The patient id to be locked or unlocked
+#set ($lockingRequest = $!{request.getParameter('lock')})##The locking request. True to lock, false to unlock.
+#set ($resultSuccessful = false)
+#if ("$!{patientId}" != '' &amp;&amp; "$!{lockingRequest}" != '')
+  #set ($resultSuccessful = $services.recordLocking.setRecordLock($patientId, $lockingRequest))
+#end
+#if ($resultSuccessful)
+  $response.setStatus(200)
+#else
+  $response.setStatus(500)
```
sdumitriu: `500` is an internal server error, specifically an error in the code. This branch is not usually an unknown server error, it's a bad request from the client. Maybe the return type should be changed from a simple boolean to an enum to indicate what exactly is wrong: unauthorized request, missing/wrong parameters, actual internal error.

tmarathe:
I created a new record from account A, locked it, and added account B as a collaborator. When the permission level for B is set to "Can view and modify patient data", B is only able to view the locked record. However if the permission level for B is set to "Can view and modify patient data and manage collaborators", B is able to unlock the record, edit it, etc. Is this expected behaviour? If so, it should be made clear that you are adding a new manager of the record.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
```diff
@@ -0,0 +1,397 @@
...
+  <hidden>false</hidden>
+  <content>{{velocity}}
+#set ($patientId = $!{request.getParameter('patientId')})##The patient id to be locked or unlocked
+#set ($lockingRequest = $!{request.getParameter('lock')})##The locking request. True to lock, false to unlock.
+#set ($resultSuccessful = false)
+#if ("$!{patientId}" != '' &amp;&amp; "$!{lockingRequest}" != '')
+  #set ($resultSuccessful = $services.recordLocking.setRecordLock($patientId, $lockingRequest))
+#end
+#if ($resultSuccessful)
+  $response.setStatus(200)
```
sdumitriu: What happens if the user middle-click on the lock button? He'll end up in a strange place (the header is different in the PhenoTips space), with an empty document. If this isn't an AJAX request, then redirect back to the patient record.

components/patient-access-rules/ui/src/main/resources/PhenoTips/PatientLock.xml
sdumitriu: If you're in the middle of an edit and click this button, then you'll lose any unsaved changes. Display the whole lock UI only in view mode.

sdumitriu:
@tmarathe Yes, that's the intended behavior.

One possible improvement would be to change the name of that access level, since more than just collaborators can be managed.

tmarathe:
If I lock the record while in Edit Mode, I get the following error screen:
screen shot 2015-06-11 at 11 06 37 am

If you return to the record, it has been locked successfully, despite the error message.

mjshepherd:
@antonkatz @veronikaslc you'll need to give this a second look

components/record-locking/api/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.phenotips</groupId>
+    <artifactId>patient-record-locking</artifactId>
+    <version>1.2-SNAPSHOT</version>
+  </parent>
+  <artifactId>phenotips-record-locking-api</artifactId>
+  <name>PhenoTips - Record Locking - API</name>
+  <properties>
+    <checkstyle.skip>true</checkstyle.skip>
```
antonkatz: cheater! :)

components/record-locking/api/src/main/java/org/phenotips/recordLocking/PatientRecordLockManager.java
```diff
@@ -0,0 +1,52 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.recordLocking;
+
+import org.phenotips.data.Patient;
+
+import org.xwiki.component.annotation.Role;
+
+/** A component to add, remove and handle locks on patient records. A record lock removes edit rights for a patient
```
antonkatz: new line

components/record-locking/api/pom.xml
sdumitriu: That's against the rules.

components/record-locking/api/src/main/java/org/phenotips/recordLocking/internal/authorization/LockedAuthorizationModule.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return 110;
+    }
+
+    @Override public Boolean hasAccess(User user, Right access, DocumentReference document)
+    {
+        XWikiContext context = contextProvider.get();
+
+        try {
+            XWikiDocument doc = context.getWiki().getDocument(document, context);
```
antonkatz: Why aren't you using the ```PatientRecordLockManager```

components/record-locking/api/src/main/java/org/phenotips/recordLocking/script/RecordLockingService.java
```diff
@@ -0,0 +1,96 @@
...
+    @Inject
+    private PatientRepository pr;
+
+    /**
+     * Locks the patient record.
+     * @param patient The patient to be locked
+     * @return A {@link HttpStatus} indicating the status of the request.
+     */
+    public int lockPatient(Patient patient) {
+        return lockManager.lockPatientRecord(patient) ? HttpStatus.SC_OK : HttpStatus.SC_BAD_REQUEST;
```
antonkatz: ```this.lockManager```

components/record-locking/api/src/main/java/org/phenotips/recordLocking/script/RecordLockingService.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Unlocks the patient record.
+     * @param patient The patient to be unlocked
+     * @return A {@link HttpStatus} indicating the status of the request.
+     */
+    public int unlockPatient(Patient patient) {
+        return lockManager.unlockPatientRecord(patient) ? HttpStatus.SC_OK : HttpStatus.SC_BAD_REQUEST;
```
antonkatz: same here

components/record-locking/api/src/test/java/org/phenotips/recordLocking/internal/DefaultPatientRecordLockManagerTest.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @Test
+    public void returnsFalseAfterXWikiExceptionWhileRetrievingDocument() throws ComponentLookupException, XWikiException
+    {
+        Mockito.doThrow(new XWikiException()).when(xwiki).getDocument(patientDocumentReference, context);
+        PatientRecordLockManager mockedLockManager = this.mocker.getComponentUnderTest();
+        Assert.assertFalse(mockedLockManager.lockPatientRecord(patient));
+    }
+
```
antonkatz: How come you don't test for normal/expected locking and unlocking?

components/record-locking/api/src/test/java/org/phenotips/recordLocking/internal/authorization/LockedAuthorizationModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+    {
+        MockitoAnnotations.initMocks(this);
+        ParameterizedType cpType = new DefaultParameterizedType(null, Provider.class, XWikiContext.class);
+        this.contextProvider = this.mocker.getInstance(cpType);
+
+        Mockito.doReturn(context).when(contextProvider).get();
+        Mockito.doReturn(xwiki).when(context).getWiki();
+
+
+    }
```
antonkatz: so much space

components/record-locking/api/src/test/java/org/phenotips/recordLocking/internal/DefaultPatientRecordLockManagerTest.java
mjshepherd: I do, ``canLockPatient`` and ``canUnlockPatient``

antonkatz: Right... pardon

Issue #1710, opened 2015-06-01
Need "metabolic" and "endocrine" section for many HPO terms that currently fall under the heading of "other"


CourtneyHum:
![image](https://cloud.githubusercontent.com/assets/11950131/7922415/1b42e602-087a-11e5-995a-4753e5af5198.png)

sdumitriu:
Looks like a regression of #580. @antonkatz can you look at it?

antonkatz:
There's a pull request for this issue that seems to have no unresolved comments with it.

PR #1757, opened 2015-06-10
Issue #1710: Need "metabolic" and "endocrine" section for many HPO te…

antonkatz:
…rms that currently fall under the heading of "other".

### DIFF
--- a/components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
+++ b/components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
@@ -188,7 +188,7 @@ public class SolrVocabularyTerm implements VocabularyTerm
             return 0;
         }
 
-        long distance = Integer.MAX_VALUE;
+        long distance = Long.MAX_VALUE;
 
         Map<String, Integer> myLevelMap = new HashMap<String, Integer>();
         myLevelMap.put(getId(), 0);
@@ -200,17 +200,17 @@ public class SolrVocabularyTerm implements VocabularyTerm
         Set<VocabularyTerm> otherCrtLevel = new HashSet<VocabularyTerm>();
         otherCrtLevel.add(other);
 
-        for (int l = 1; l <= distance && !myCrtLevel.isEmpty() && !otherCrtLevel.isEmpty(); ++l) {
+        for (int l = 1; l <= distance && (!myCrtLevel.isEmpty() || !otherCrtLevel.isEmpty()); ++l) {
             distance = Math.min(distance, processAncestorsAtDistance(l, myCrtLevel, myLevelMap, otherLevelMap));
             distance = Math.min(distance, processAncestorsAtDistance(l, otherCrtLevel, otherLevelMap, myLevelMap));
         }
-        return distance == Integer.MAX_VALUE ? -1 : distance;
+        return distance == Long.MAX_VALUE ? -1 : distance;
     }
 
     private long processAncestorsAtDistance(int localDistance, Set<VocabularyTerm> sourceUnprocessedAncestors,
         Map<String, Integer> sourceDistanceMap, Map<String, Integer> targetDistanceMap)
     {
-        long minDistance = Integer.MAX_VALUE;
+        long minDistance = Long.MAX_VALUE;
         Set<VocabularyTerm> nextLevel = new HashSet<VocabularyTerm>();
         for (VocabularyTerm term : sourceUnprocessedAncestors) {
             for (VocabularyTerm parent : term.getParents()) {
--- a/components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
+++ b/components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
@@ -24,6 +24,7 @@ import org.xwiki.component.annotation.Component;
 import java.text.MessageFormat;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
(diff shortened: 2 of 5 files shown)


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
```diff
@@ -200,17 +200,17 @@ public long getDistanceTo(final VocabularyTerm other)
         Set<VocabularyTerm> otherCrtLevel = new HashSet<VocabularyTerm>();
         otherCrtLevel.add(other);
 
-        for (int l = 1; l <= distance && !myCrtLevel.isEmpty() && !otherCrtLevel.isEmpty(); ++l) {
+        for (int l = 1; l <= distance && (!myCrtLevel.isEmpty() || !otherCrtLevel.isEmpty()); ++l) {
```
antonkatz: @sdumitriu this one I'm not sure about. But to me it seems like ```&&``` was a bug.

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
```diff
@@ -103,6 +117,51 @@ protected int getSolrDocsPerBatch()
...
+        List<VocabularyTerm> sortedCategories = this.sortTermsByLevel(categoriesSet);
+        List<String> sortedCategoriesIds = new LinkedList<>();
+        for (VocabularyTerm category : sortedCategories) {
+            sortedCategoriesIds.add(category.getId());
+        }
+        json.put("term_category", sortedCategoriesIds);
+        return json;
+    }
+
+    private List<VocabularyTerm> sortTermsByLevel(Collection<VocabularyTerm> terms)
```
antonkatz: Or the scala equivalent ```terms.toSeq.map {t => t -> t.getDistanceTo(getRootTerm)} sortBy(_._2) unzip(1)```

antonkatz:
@JKereliuk This really needs to be tested. I made some potentially significant changes.
I nominate @mjshepherd @meatcar for review

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
meatcar: why dont you add this as a comment to the code, so it's easeir to read? :stuck_out_tongue:

meatcar:
looks good to me. :+1:

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
sdumitriu: Yes, you're right. But now you need to add a test that captures this case.

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
```diff
@@ -103,6 +117,51 @@ protected int getSolrDocsPerBatch()
...
+    {
+        List<VocabularyTerm> sortedTerms = new LinkedList<>();
+        Map<Long, VocabularyTerm> levelMap = new HashMap<>();
+        for (VocabularyTerm term : terms) {
+            levelMap.put(term.getDistanceTo(this.getRootTerm()), term);
```
mjshepherd: Couldn't all this code be simplified using a comparator?

antonkatz: You're right, but I'm not sure how performant ```getDistanceTo()``` is. A comparator would make unnecessary calls.

mjshepherd: Maybe compromise. You could make the level map (but swap the keys and values) then sort your list based on each terms value in the map. Then your level map acts as a cache for the ``getDistanceTo`` method.

antonkatz: Haven't thought of that. That's what I should've done. Do you think I should rewrite it?

mjshepherd: That depends, do you want a +1? :)

antonkatz: Darn :) The things I'll do for a plus one.

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
antonkatz: Why don't we just write in Scala? :P

antonkatz:
@mjshepherd ready to give me that :+1: ?

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
```diff
@@ -103,6 +118,54 @@ protected int getSolrDocsPerBatch()
...
+    private List<VocabularyTerm> sortTermsByLevel(Collection<VocabularyTerm> terms)
+    {
+        List<VocabularyTerm> sortedTerms = new LinkedList<>();
+        final Map<VocabularyTerm, Long> levelMap = new HashMap<>();
+        for (VocabularyTerm term : terms) {
+            levelMap.put(term, term.getDistanceTo(this.getRootTerm()));
+            sortedTerms.add(term);
+        }
+        sortedTerms.sort(new Comparator<VocabularyTerm>()
+        {
```
mjshepherd: Jenkins build is failing. I think because List.sort() is not implemented until Java 8. I don't actually know what our current status on java 8 is. If we can't use it then just use Collections.sort()

sdumitriu: We require Java 7, so the build must pass with that JDK. Indeed, replace this with `Collections.sort`.

mjshepherd:
:+1: @sdumitriu

sdumitriu:
I don't like this approach at all, since it explicitly adds a dependency on HPO as the symptom vocabulary, a custom method for this vocabulary, and too much code in general.

Issue #1654, opened 2015-05-08
Add "Other" as an option for the patient's sex


sdumitriu:

allasm:
TODO: also update push functionality, matchmaking and pedigree handling of the "sex" field values

marta-:
@CourtneyHum Please list the possible options for "other" patient sex here.

itaiGershtansky:
From looking at the code, I'm guessing there might be some problems when a patient with sex "other" is a parent.

CourtneyHum:
I am proposing a greater number of options in the field “Sex” based on a literature review and consultation with Riyana Babul-Hirji, CCGC, and Dr. David Chitayat who work in the Multidisciplinary Urogenital Clinic at Sick Kids. 

In 2006, the Lawson Wilkins Pediatric Endocring Society and the European Society for Paediatric Endocrinology consensus group proposed the new Chicago classification of Disorders of Sex Development (DSD). The nomenclature is based on the primary genetic defect and includes three broad categories. 

![image](https://cloud.githubusercontent.com/assets/11950131/8337061/ae3cd374-1a76-11e5-8d1f-97f3383cac64.png)

There are several papers that reference this Chicago classification and I have included above a table that clearly demonstrates the categories. 

It would seem ideal to include the three categories as options under a “sex” label of “other”.

The issue of determining corresponding growth chart, however, is not simple when using the three categories outlined in the Chicago classification. The rule used by Dr. Chitayat is with the presence of a Y chromosome, a male growth chart is used, except for individuals with complete androgen insensitivity syndrome (CAIS).

Thus, when “Sex chromosome DSD” is selected and the karyotype entered in the free text field, there would need to be a way to detect the presence or absence of a Y chromosome, or a way for the clinician to indicate if male or female growth charts should be used.

When “46, XY DSD” is selected, the individuals’ measurements should be plotted on a male growth chart, except when CAIS entered in the free text field. Alternatively, CAIS could be an option that appears when 46, XY DSD is selected.

When “46, XX DSD” is selected, a female growth chart is always used.

[...]

PR #1788, opened 2015-06-23
Issue 1654#2

itaiGershtansky:
@allasm, @danielpgross, please review

### DIFF
--- a/components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
+++ b/components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
@@ -823,7 +823,6 @@
         return new Element('dd').insert(id);
     },
     __displaySex : function(sex) {
-        console.log("push patient");
         return new Element('dd').insert(
            (sex == "F") ? "$services.localization.render('phenotips.PushPatient.female')" :
            (sex == "M") ? "$services.localization.render('phenotips.PushPatient.male')" :
--- a/components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
+++ b/components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
@@ -823,7 +823,12 @@
         return new Element('dd').insert(id);
     },
     __displaySex : function(sex) {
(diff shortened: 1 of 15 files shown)


danielpgross:
:+1:

On a non-technical UI point, I think it's worth considering the symbol used for "Other". The diamond symbol seems to consistently represent "unknown gender" or "unspecified gender" in pedigree drawing standards, but "other" has a distinctly different meaning, like just like "null" is different from "undefined". Maybe there should be an "O" overlaid onto the diamond. +@sdumitriu and @allasm for feedback.

marta-:
@CourtneyHum any suggestions on the symbol to use in the pedigree when the gender is "Other"?

CourtneyHum:
If going by the standard pedigree nomenclature 

"The Pedigree Standardization Work Group continues to recommend that the male or female symbol be used to define the phenotypic gender, with the karyotype noted below the symbol, when known...A diamond can be used when it is not clinically relevant to assign gender"

image

CourtneyHum:
I don't know if this is possible, but could the default be a diamond for someone with gender "other" but either male or female selected in the pedigree editor by the clinician, without changing in the Patient Information section?

marta-:
Could the different options listed in https://github.com/phenotips/phenotips/issues/1654#issuecomment-114951789 be each mapped to a gender symbol?

CourtneyHum:
image

It's not straightforward. 
1. a) i) female
        ii) either
        iii) either
        iv) either
    b) i) either

etc...

With most of these DSD's there is a spectrum of severity depending on the effect of the mutation so the decision to raise an individual as a male or female is difficult and not always predictable based on the condition.

CourtneyHum:
@bgallinger I'd love to discuss this with you because the issue is getting more and more complicated

CourtneyHum:
In The Practical Guide to the Genetic Family History

"A diamond can also represent persons with congenital disorders of sex development in which chromosomal, gonadal, or anatomic sex is atypical."

marta-:
So should there be a note under the diamond stating what kind of "other" it is? would that make things clearer for everyone?

CourtneyHum:
Yes the user should write in the karyotype and the diagnosis

image

CourtneyHum:
Other
- karyotype ____________ (anything with a Y is coded as male for growth charts)
- phenotypic sex M | F | Neither (to determine pedigree symbol)

Issue #1640, opened 2015-05-04
Vocabularies should auto-generate their Solr configuration at startup


sdumitriu:
This is needed so that new ontologies can be easily installed as extensions, without the need to manually modify the Solr configuration on disk.

At startup, an ontology component copies the configuration from inside the jar into the solr directory, and calls `org.apache.solr.core.CoreContainer.create(CoreDescriptor)` to register it.

veronikaslc:
To sum up what has to be done:
1. Instead of having all the configuration files for the ontologies in resources/solr-configuration, each vocabulary should have its own configuration (but not the actual index files) in its src/main/resources 
2. Implement SolrVocabularyResourceManager.java initialize() method to prepare and register the core, if it doesn't exist already, by: 
 -- Copy the configuration from inside the jar into the solr directory <permanent directory>/solr/<core name>
 -- Call org.apache.solr.core.CoreContainer.create(CoreDescriptor) to register it.

PR #1790, opened 2015-06-24
Issue #1640: 

veronikaslc:
@meatcar, @antonkatz if you can review, there is one file changed - DefaultSolrVocabularyResourceManager.java, and solr configuration files are moved from resources/solr-configuration to the corresponding vocabularies components src/main/resources folders.


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
```diff
@@ -59,16 +73,84 @@
...
+            CoreDescriptor dcore =
+                new CoreDescriptor(container, vocabularyName, solrHome.toPath().resolve(vocabularyName).toString());
+
+            this.core = new EmbeddedSolrServer(container, vocabularyName);
+
+            this.score = container.create(dcore);
+
             this.cache = this.cacheFactory.createNewLocalCache(new CacheConfiguration());
         } catch (RuntimeException ex) {
             throw new InitializationException("Invalid Solr core: " + ex.getMessage());
```
meatcar: you should pass the caught exception into the new exception like this

```java
thrown new Exception("message", e);
```

This way you can preserve the stack trace, and make debugging easier.

components/vocabularies/chebi/api/src/main/resources/chebi/conf/solrconfig.xml
```diff
@@ -1,4 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+-->
+
 <!--
```
meatcar: There's a double licence in this file. We only do AGPLv3 now, just run `mvn license:remove` followed by `mvn license:format` to update the licenses.

components/vocabularies/chebi/api/src/main/resources/chebi/conf/schema.xml
```diff
@@ -1,4 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+-->
+
 <!--
```
meatcar: There's a double licence in this file. We only do AGPLv3 now, just run `mvn license:remove` followed by `mvn license:format` to update the licenses.

components/vocabularies/ethnicity/api/src/main/resources/ethnicity/conf/schema.xml
```diff
@@ -1,4 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+-->
+
 <!--
```
meatcar: There's a double licence in this file. We only do AGPLv3 now, just run `mvn license:remove` followed by `mvn license:format` to update the licenses.

components/vocabularies/ethnicity/api/src/main/resources/ethnicity/conf/solrconfig.xml
```diff
@@ -1,4 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+-->
+
 <!--
```
meatcar: There's a double licence in this file. We only do AGPLv3 now, just run `mvn license:remove` followed by `mvn license:format` to update the licenses.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
```diff
@@ -59,16 +73,84 @@
...
+
+            if (directoryListing == null) {
+                return;
+            }
+
+            for (File file : directoryListing) {
+                JarFile jar = new JarFile(file);
+                String jarName = jar.getName().substring(jarsHome.length());
+                if ("jar".equals(FilenameUtils.getExtension(file.toPath().toString()))
+                    && jarName.startsWith("vocabulary-" + vocabularyName)) {
```
meatcar: Do we ever have multiple jars with the same vocabulary name, from which we want to copy resources? Maybe its better to be more selective, and extract the jar location from specific classes, so we don't accidentally extract some files from a jar that matched this pattern? For example, if we have two versions of vocabularies.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
meatcar: In general, you don't need to catch runtime exceptions.

antonkatz:
There's a file rename ``` ...c/main/resources/chebi/conf/stopwords.txt → ...rc/main/resources/omim/conf/stopwords.txt```. Is the ```stopwords``` file from ```chebi``` supposed to be in ```omim```?

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
```diff
@@ -59,16 +73,84 @@
...
+    @Inject
+    private Environment environment;
+
     @Override
     public void initialize(String vocabularyName) throws InitializationException
     {
+        // Get jars home path to where the jar resources are stored
+        Class<DefaultSolrVocabularyResourceManager> clazz = DefaultSolrVocabularyResourceManager.class;
+        String jarPath = clazz.getProtectionDomain().getCodeSource().getLocation().getPath();
+        String jarsHome = jarPath.substring(0, jarPath.lastIndexOf('/'));
```
antonkatz: This is not nice. Should be done though Java's file API.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
```diff
@@ -59,16 +73,84 @@
...
+     *
+     * @param jar the jar file
+     * @param vocabularyName the name of the vocabulary being managed
+     * @param dest the destination
+     * @throws IOException
+     */
+    private static void copyConfigsFromJar(JarFile jar, String vocabularyName, File dest) throws IOException
+    {
+        // return if vocabulary directory already exists
+        if (Files.isDirectory(dest.toPath().resolve(vocabularyName))) {
```
antonkatz: Why?

sdumitriu:
@antonkatz stopwords rename: it's the same file, shared between all cores.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
sdumitriu: The assumption is that if the core files have already been extracted during the previous startup, then it's not necessary to re-copy them over.

However, given that it's not that costly to copy a few files, we should copy them each time. This also ensures that after an upgrade, the new version of the files will be used.

So yes, remove that condition.

components/vocabularies/chebi/api/src/main/resources/chebi/conf/schema.xml
veronikaslc: unfortunately it doesn't remove unnecessary license, have to do manually

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
veronikaslc: @sdumitriu - is there a way to dynamically inject/get a vocabulary class somehow to get its jar path? Injecting component gives only corresponding ScriptService.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java
```diff
@@ -59,16 +73,75 @@
...
 
+    @Inject
+    private Environment environment;
+
     @Override
     public void initialize(String vocabularyName) throws InitializationException
     {
+        // Get jars home path to where the jar resources are stored
+        Class<DefaultSolrVocabularyResourceManager> clazz = DefaultSolrVocabularyResourceManager.class;
+        String jarPath = clazz.getProtectionDomain().getCodeSource().getLocation().getPath();
```
sdumitriu: That is indeed the way to find the jar where a class is defined. However, you're getting the jar wher the `DefaultSolrVocabularyResourceManager` is defined, not the one for the vocabulary you're trying to initialize. If you want to get the right jar, you will have to call this from the vocabulary class, not from here, and pass the jarPath to this `initialize` method. But I don't quite like that.

An alternative is to assume a fixed set of configuration files that must be copied, and only try to copy those instead of looking inside the jar, in which case you don't care what the jar is. Something along the line of:

```

    public static final List<String> CONFIG_FILES = Arrays.asList("schema.xml", "solrconfig.xml", "protwords.txt", "stopwords.txt", "synonyms.txt");
...
        for (String file : CONFIG_FILES) {
            InputStream in = this.getClass().getResourceAsStream("/" + vocabularyName + "/conf/" + file);
            if (in == null) {
                continue;
            }
            Files.copy(in, dest.toPath().resolve(file));
        }

File: components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultSolrVocabularyResourceManager.java [hunk 10]
```diff
@@ -45,12 +56,18 @@
 @InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
 public class DefaultSolrVocabularyResourceManager implements SolrVocabularyResourceManager
 {
+    /** List of config Solr files. */
+    public static final List<String> CONFIG_FILES = Arrays.asList("/conf/schema.xml", "/conf/solrconfig.xml",
+        "/conf/protwords.txt", "/conf/stopwords.txt", "/conf/synonyms.txt", "/core.properties.properties");
```
sdumitriu: Are you sure it's `core.properties.properties`, with two `.properties`?

veronikaslc: should be just core.properties.

veronikaslc:
Closing this PR, rebased this branch to a new PT-1698 for cleaning purposes.

PR #1792, opened 2015-06-25
Default domain object factory tests

JKereliuk:


JKereliuk:
Nominating @tmarathe  @DeanWay  @danielpgross to review!

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,254 @@
...
+        when(this.patient.getDocument()).thenReturn(null);
+        when(this.access.hasAccess(Right.VIEW, null, null)).thenReturn(false);
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(this.patient, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientDocumentExceptionReturnsNull() throws Exception {
+        DocumentReference documentReference = mock(DocumentReference.class);
+        when(this.patient.getDocument()).thenReturn(documentReference);
+        when(this.access.hasAccess(Right.VIEW, null, null)).thenReturn(true);
```
DeanWay: this should be changed to `when(this.access.hasAccess(Right.VIEW, null, documentReference)).thenReturn(true);` else you are not effectively setting the behaviour of `hasAccess()` for this test since `patient.getDocument()` will return a mock object, not null

DeanWay: Actually, keep this line how it is, do not create the mock DocumentReference and of set the return value of `patient.getDocument()` to null and change the next line to `when(this.documentAccessBridge.getDocument(Matchers.any())).thenThrow(Exception.class);`
Creating the mock is unnecessary since you are not using it to set any new behavior

tmarathe: @DeanWay Shouldn't it be `when(this.access.hasAccess(Right.VIEW, documentReference, null)).thenReturn(true);`? 'hasAccess` is called as `hasAccess(right, documentReference, entityReference)`.

DeanWay: @tmarathe DocumentReference is a subclass of EntityReference. The usage we are testing calls as follows `hasAccess(Right, DocumentReference, DocumentReference)` where `patient.getDocument()` is passed as the third argument

JKereliuk: Good catch!

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,255 @@
...
+        Object[] summary = {new Object(), new Object(), new Object(), new Object(), new Object(), new Object(), new Object()};
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(summary, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientFromSummaryNoAccess() throws Exception
+    {
+        when(this.patient.getDocument()).thenReturn(null);
+        when(this.access.hasAccess(Right.VIEW, null, null)).thenReturn(false);
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(this.patient, this.uriInfo));
```
DeanWay: This test is currently redundant with `cannotCreatePatientWithNoAccess()`
The first line of this method should be changed to `when(this.stringResolver.resolve(Matchers.anyString())).thenReturn(null);`
and the last line should be changed to `assertNull(this.mocker.getComponentUnderTest().createPatientSummary({some array of objects}, this.uriInfo));`

JKereliuk: You are correct, I forgot to update this after the copy paste probably :/

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,255 @@
...
+    public void createPatientFromSummaryWrongLength() throws Exception
+    {
+        Object[] summary = {new Object()};
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(summary, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientFromSummaryWrongObjectTypes() throws Exception
+    {
+        Object[] summary = {new Object(), new Object(), new Object(), new Object(), new Object(), new Object(), new Object()};
```
DeanWay: Perhaps change this line to `Object[] summary = new Object[7];` followed by `Arrays.fill(summary, new Object());` It's a little different in that the entire array will just be references to the same object, but that doesn't really matter and it's a little more apparent that you want this array to pass the test of length == 7

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+        PatientSummary patientSummary = mocker.getComponentUnderTest().createPatientSummary(this.patient, this.uriInfo);
+
+        assertEquals("id", patientSummary.getId());
+        assertEquals("externalid", patientSummary.getEid());
+        assertEquals("creator", patientSummary.getCreatedBy());
+        assertEquals("creator", patientSummary.getLastModifiedBy());
+        assertEquals("version", patientSummary.getVersion());
+        assertTrue(patientSummary.getCreatedOn() instanceof DateTime);
+        assertTrue(patientSummary.getLastModifiedOn() instanceof DateTime);
```
DeanWay: checking that that the return value is an instanceof DateTime here is unnecessary, since that is the return type of the method and will always be that or null. In other words this could just be `AssertNotNull()` but I think it would be better to check for the exact value that you set up: `creationDate`

danielpgross:
How about the case when the patient document doesn't exist?

JKereliuk:
@danielpgross this is a good point. In this case DocumentReferenceResolver would return null, so to not get a segfault we would require there be no access. @sdumitriu  Perhaps we should add in a check that returns null if the doc is null?

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+        when(this.patient.getDocument()).thenReturn(null);
+        when(this.access.hasAccess(Right.VIEW, null, null)).thenReturn(false);
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(this.patient, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientDocumentExceptionReturnsNull() throws Exception {
+        when(this.patient.getDocument()).thenReturn(null);
+        when(this.access.hasAccess(Right.VIEW, null, null)).thenReturn(true);
+        when(this.documentAccessBridge.getDocument(Matchers.any(DocumentReference.class))).thenThrow(Exception.class);
```
DeanWay: `doThrow(Exception.class).when(this.documentAccessBridge).getDocument(any(DocumentReference.class));` to avoid unchecked warning

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+    public void createPatientFromSummaryWrongObjectTypes() throws Exception
+    {
+        Object[] summary = new Object[7];
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(summary, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientFromSummaryNoAccess() throws Exception
+    {
+        Object[] summary = {new Object()};
```
DeanWay: `Object[] summary = new Object[7]` the null return value you are getting currently is due to the `summaryData.length != 7` condition failing

DeanWay: The `instanceof Date` conditions need to be met as well

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactoryTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+        doThrow(Exception.class).when(this.documentAccessBridge).getDocument(Matchers.any(DocumentReference.class));
+        assertNull(this.mocker.getComponentUnderTest().createPatientSummary(this.patient, this.uriInfo));
+    }
+
+    @Test
+    public void createPatientPerformsCorrectly() throws ComponentLookupException, URISyntaxException, Exception
+    {
+        XWikiDocument documentReference = mock(XWikiDocument.class);
+        DocumentReference patientDocument = mock(DocumentReference.class);
```
sdumitriu: `DocumentReference` are very stupid POJOs, you shouldn't mock them but instantiate a new one:

    new DocumentReference("wiki", "data", "P0000001")

JKereliuk: I will make another request with a fix

sdumitriu: No need, already fixed it.

PR #1799, opened 2015-06-26
issue#1654 Add 'Other' as an option for the patient's sex

itaiGershtansky:
When exporting in a BOADICEA format, treats 'other' sex the same as 'unknown' sex


allasm:
+1

Issue #1671, opened 2015-05-20
The pedigree editor fails to load when using PhenoTips behind a reverse proxy via mod_proxy_http


sdumitriu:
It works fine with `mod_proxy_ajp`, but there's a bug in the XWiki REST implementation when using `mod_proxy_http`: the REST system ignores the forwarded protocol/host/port and uses the internal ones, so the URLs are invalid. The pedigree code that deals with URLs found in REST responses should strip away the host and use only the path.

PR #1801, opened 2015-06-30
Issue #1671: The pedigree editor fails to load when using PhenoTips behind a reverse proxy via mod_proxy_http

veronikaslc:
Fixed. 
@allasm, @danielpgross - can you guys make a quick glance over little change in templateSelector.js?


danielpgross:
Are the ```pathname``` and ```origin``` properties available in IE9 and up? I looked [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement#Browser_compatibility), but the answer isn't clear.

allasm:
Looks like <a href="https://developer.mozilla.org/en-US/docs/Web/API/URLUtils/origin">origin</a> is not supported, but there seem to be a <a href="https://github.com/m7r/grunt-ngdocs/issues/106">workaround</a>

sdumitriu:
There's no need for origin, just remove the hostname completely and make path-only request, the browser will know to fill in the right protocol, host and port.

sdumitriu:
I don't think that using the `pathname` attribute is right, for two reasons:

- it's not clear which browsers (and which versions) actually implement it
- it discards the query string, if there is any

components/pedigree/resources/src/main/resources/pedigree/view/templateSelector.js
```diff
@@ -45,7 +45,9 @@ var TemplateSelector = Class.create( {
             pictureBox.update("Loading...");
             this.mainDiv.insert(pictureBox);
             var href = getSelectorFromXML(objects[i], "link", "rel", "http://www.xwiki.org/rel/properties").getAttribute("href");
-            new Ajax.Request(href, {
+            //should use only path when running PhenoTips behind a reverse proxy via mod_proxy_http
+            var path = href.substring(href.indexOf("/rest"));
```
sdumitriu: This discards the appname from the path. You should substring from the first `/` after `//`:

    var path = href.substring(href.indexOf("/", href.indexOf("//") + 2);

PR #1802, opened 2015-06-30
Patient data REST api tests (ConfigureISODatesInJson and DefaultPatientResourceImpl)

DeanWay:


components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,359 @@
...
+        doReturn(this.uri).when(this.uriInfo).getRequestUri();
+
+        ReflectionUtils.setFieldValue(this.patientResource, "logger", this.logger);
+        ReflectionUtils.setFieldValue(this.patientResource, "repository", this.repository);
+        ReflectionUtils.setFieldValue(this.patientResource, "access", this.access);
+        ReflectionUtils.setFieldValue(this.patientResource, "users", this.users);
+        ReflectionUtils.setFieldValue(this.patientResource, "uriInfo", this.uriInfo);
+    }
+
+    //----------------------------Get Patient Tests----------------------------
```
DeanWay: @sdumitriu  @marta- Are these kinds of organizational comments allowed in our style?

DeanWay:
@JKereliuk @tmarathe @veronikaslc please review 😄

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
sdumitriu: Yes.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,361 @@
...
+        doReturn(this.currentUser).when(this.users).getCurrentUser();
+        doReturn(null).when(this.currentUser).getProfileDocument();
+
+        doReturn(this.patient).when(this.repository).getPatientById(this.id);
+        doReturn(null).when(this.patient).getDocument();
+
+        this.uri = new URI(this.uriString);
+        doReturn(this.uri).when(this.uriInfo).getRequestUri();
+
+        ReflectionUtils.setFieldValue(this.patientResource, "logger", this.logger);
```
sdumitriu: You shouldn't have to do this. `MockitoComponentMockingRule` already sets mocks for the `@Inject`-ed fields, which you can access with `mocker.getInstance(PatientRepository.class)`, for example.

DeanWay: Will this work for the Logger? Logger is an interface but it is not an `@Role`, and I'm not sure where it is getting the implementation, since there is no `@Named` hint. When I tried `this.mocker.getInstance(Logger.class)` a ComponentLookupException was thrown

sdumitriu: For the logger there's a special method: `this.mocker.getMockedLogger()`

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/ConfigureISODatesInJsonTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Assert.assertTrue(testInstance.getEvents().get(0).getClass() == ApplicationStartedEvent.class);
+    }
+
+    @Test
+    public void verifyOnEventConfiguresProperly() throws ComponentLookupException {
+        converterList.add(converter);
+        converterList.add(jacksonConverter);
+        Engine.setInstance(engine);
+        Mockito.when(engine.getRegisteredConverters()).thenReturn(converterList);
+        Mockito.when(jacksonConverter.getObjectMapper()).thenReturn(objectMapper);
```
JKereliuk: To keep it the way the other tests are written you should statically import the when and verify methods instead of Mockito.when and Mockito.verify

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+    @SuppressWarnings("unchecked")
+    public void getPatientAddsLinksToPatientJSON() throws XWikiRestException {
+        doReturn(true).when(this.access).hasAccess(Right.VIEW, null, null);
+        doReturn(new JSONObject()).when(this.patient).toJSON();
+
+        Response response = this.patientResource.getPatient(this.id);
+
+        Assert.assertTrue(response.getEntity() instanceof JSONObject);
+
+        Map<String, Map<String, String>> json = (Map)response.getEntity();
```
JKereliuk: Why are you checking this as a map and not as a JSONObject?

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+        Assert.assertThat(json, hasValue(hasEntry("rel", Relations.SELF)));
+        Assert.assertThat(json, hasValue(hasEntry("href", this.uriString)));
+        Map<String, List<MediaType>> actualMap = (Map)response.getMetadata();
+        Assert.assertThat(actualMap, hasValue(hasItem(MediaType.APPLICATION_JSON_TYPE)));
+    }
+
+    @Test
+    public void getPatientAlwaysSendsLoggerMessageOnRequest() throws XWikiRestException {
+        doReturn(true).when(this.access).hasAccess(Right.VIEW, null, null);
+        doReturn(new JSONObject()).when(this.patient).toJSON();
```
JKereliuk: This is a small point, but to reduce the number of lines this has to go through you could just make this.repository.getPatientById(id) return null and the call would end faster with the log you are checking still happening

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+    public void updatePatientCantFindPatient() throws XWikiRestException {
+        doReturn(null).when(this.repository).getPatientById(anyString());
+        WebApplicationException ex = null;
+        try {
+            this.patientResource.updatePatient("", this.id);
+        } catch (WebApplicationException temp) {
+            ex = temp;
+        }
+        Assert.assertNotNull("updatePatient did not throw a WebApplicationException as expected " +
+            "when the patient could not be found", ex);
```
JKereliuk: Nice, I like the message. Tells what actually went wrong if it does.

sdumitriu: When you don't actually care about anything other than an exception is thrown, you can just use `@Test(expected=WebApplicationException.class)` above the method, and JUnit will test that such an exception is thrown.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
DeanWay: The `org.hamcrest.Matchers.hasEntry` and `hasValue` expect a Map. JSONObject implements the `java.uitl.Map` interface, but there will be an error unless the JSONObject is first cast to a Map

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
DeanWay: I wanted to test that the `WebApplicationException` was thrown here, but also check the status code passed to the exception and verify the logger message, because It seemed like was all essentially the same behaviour.  I started by using the `@Test(expected=WebApplicationException.class)` notation, but it seems that the test stops executing when the exception is thrown, making testing for the status code and the logger difficult

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+        this.repository = this.mocker.getInstance(PatientRepository.class);
+        this.access = this.mocker.getInstance(AuthorizationManager.class);
+        this.users = this.mocker.getInstance(UserManager.class);
+
+
+        doReturn(this.currentUser).when(this.users).getCurrentUser();
+        doReturn(null).when(this.currentUser).getProfileDocument();
+
+        doReturn(this.patient).when(this.repository).getPatientById(this.id);
+        doReturn(null).when(this.patient).getDocument();
```
sdumitriu: You should use real `DocumentReference`s for `patient.getDocument()` and `currentUser.getProfileDocument()`.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Test
+    public void deletePatientReturnsNoContentResponse() throws XWikiRestException {
+        Provider provider = mock(Provider.class);
+        XWikiContext context = mock(XWikiContext.class);
+        XWiki wiki = mock(XWiki.class);
+        doReturn(context).when(provider).get();
+        doReturn(wiki).when(context).getWiki();
+        ReflectionUtils.setFieldValue(this.patientResource, "xcontextProvider", provider);
```
sdumitriu: This shouldn't be here.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Test
+    public void deletePatientAlwaysSendsLoggerMessageOnRequest() throws XWikiRestException {
+        Provider provider = mock(Provider.class);
+        XWikiContext context = mock(XWikiContext.class);
+        XWiki wiki = mock(XWiki.class);
+        doReturn(context).when(provider).get();
+        doReturn(wiki).when(context).getWiki();
+        ReflectionUtils.setFieldValue(this.patientResource, "xcontextProvider", provider);
```
sdumitriu: This shouldn't be here.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Test
+    public void deletePatientCatchesXWikiException() throws XWikiRestException, XWikiException {
+        Provider provider = mock(Provider.class);
+        XWikiContext context = mock(XWikiContext.class);
+        XWiki wiki = mock(XWiki.class);
+        doReturn(context).when(provider).get();
+        doReturn(wiki).when(context).getWiki();
+        ReflectionUtils.setFieldValue(this.patientResource, "xcontextProvider", provider);
```
sdumitriu: This shouldn't be here.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
DeanWay: Do you mean that it should be moved to the setUp method, or that this line doesn't need to exist at all?

sdumitriu: Just like the other components injected, the xcontext provider is already mocked, so this isn't needed at all. If you're wondering how to access that mock, this should work: `this.mocker.getInstance(XWikiContext.class)`.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/ConfigureISODatesInJsonTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        converterList = new LinkedList<>();
+    }
+
+    @Test
+    public void checkConstruction() {
+        ConfigureISODatesInJson testInstance = new ConfigureISODatesInJson();
```
sdumitriu: Don't instantiate it yourself, use `mocker.getComponentUnderTest()`, since that will invoke the constructor for you.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/ConfigureISODatesInJsonTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        converterList = new LinkedList<>();
+    }
+
+    @Test
+    public void checkConstruction() throws ComponentLookupException {
+        ConfigureISODatesInJson testInstance = (ConfigureISODatesInJson)this.mocker.getComponentUnderTest();
```
sdumitriu: Why do you need to cast it? You're only using methods available in the `EventListener` interface.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,382 @@
...
+        Map<String, Map<String, String>> json = (Map<String, Map<String, String>>)response.getEntity();
+        Assert.assertThat(json, hasValue(hasEntry("rel", Relations.SELF)));
+        Assert.assertThat(json, hasValue(hasEntry("href", this.uriString)));
+        Map<String, List<MediaType>> actualMap = (Map)response.getMetadata();
+        Assert.assertThat(actualMap, hasValue(hasItem(MediaType.APPLICATION_JSON_TYPE)));
+        Assert.assertEquals(Status.OK.getStatusCode(), response.getStatus());
+    }
+
+    @Test
+    public void checkGetPatientAlwaysSendsLoggerMessageOnRequest()
```
sdumitriu: Logging is not an important part of the correct functionality of an implementation, so it doesn't have to be tested. Since it's not specified in the description of the interface, it can even be considered an implementation detail, which means that it shouldn't be tested at all.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -0,0 +1,382 @@
...
+
+        Response response = this.patientResource.deletePatient(this.id);
+
+        verify(wiki).getDocument(this.patientDocument, this.context);
+        verify(wiki).deleteDocument(patientXWikiDoc, this.context);
+        Assert.assertEquals(Status.NO_CONTENT.getStatusCode(), response.getStatus());
+    }
+
+    @Test
+    public void checkDeletePatientAlwaysSendsLoggerMessageOnRequest()
```
sdumitriu: All tests check something, so it's redundant to start their name with `check`.

PR #1810, opened 2015-07-08
DefaultPatientsResourceImpl tests

tmarathe:


tmarathe:
@DeanWay, @JKereliuk please review!

JKereliuk:
Update your license headers with ```mvn license:format```

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+        this.queries = this.mocker.getInstance(QueryManager.class);
+        this.uri = new URI("http://uri");
+
+        doReturn(this.uri).when(this.uriInfo).getBaseUri();
+        doReturn(this.uri).when(this.uriInfo).getRequestUri();
+        ReflectionUtils.setFieldValue(this.patientsResource, "uriInfo", this.uriInfo);
+
+        doReturn("00000001").when(this.patient).getId();
+        doReturn(this.currentUser).when(this.users).getCurrentUser();
+        doReturn(null).when(this.currentUser).getProfileDocument();
```
DeanWay: Return a real DocumentReference object here and mock the `currentResolver` to return a real DocumentReference object and check for them when mocking the behaviour of `hasAccess` instead of checking for null values or mocks. 
i.e: 
```java 
private DocumentReference userProfileDocument = new DocumentReference("wiki", "user", "00000001");
```

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @Before
+    public void setUp() throws ComponentLookupException, URISyntaxException {
+        MockitoAnnotations.initMocks(this);
+        Execution execution = mock(Execution.class);
+        ExecutionContext executionContext = mock(ExecutionContext.class);
+        ComponentManager compManager = this.mocker.getInstance(ComponentManager.class, "context");
+        when(compManager.getInstance(Execution.class)).thenReturn(execution);
+        doReturn(executionContext).when(execution).getContext();
+        doReturn(mock(XWikiContext.class)).when(executionContext).getProperty("xwikicontext");
```
DeanWay: Instead of returning a mock XWikiContext return the XWikiContext created by the `MockitoComponentMockingRule ` this also makes use of the `context` field you have written but are not currently using
here is how you do that:
```java
Provider<XWikiContext> provider = this.mocker.getInstance(XWikiContext.TYPE_PROVIDER);
this.context = provider.get();
...
doReturn(this.context).when(executionContext).getProperty("xwikicontext");
```

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+        verify(this.logger).error(eq("Could not process remote matching request: {}"), anyString(), anyObject());
+    }
+
+    @Test
+    public void addPatientAsJSON() throws XWikiRestException {
+        doReturn(true).when(this.access).hasAccess(eq(Right.EDIT), any(DocumentReference.class), any(EntityReference.class));
+        JSONObject json = new JSONObject();
+        Response response = this.patientsResource.addPatient(json.toString());
+        verify(this.logger).debug("Importing new patient from JSON via REST: {}", json.toString());
+    }
```
DeanWay: The only thing you are currently verifying in this test is a logger debug statement that will execute every time whenever the method is called. There is a lot of functionality that you are missing here. 
You need to mock the behaviour of `this.repository.createNewPatient()` and verify that that call occurred. As you have it written currently that method call will return null, and the following statement `patient.updateFromJSON(jsonInput);` will generate a `NullPointerException`. Once that change is made you should  perform some checks on the return value of the `addPatient` method in this test

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.phenotips.data.PatientRepository;
+import org.phenotips.data.rest.DomainObjectFactory;
+import org.phenotips.data.rest.PatientsResource;
+import org.phenotips.data.rest.model.Patient;
```
DeanWay: The class you are testing uses a different `Patient` class. Change to `import org.phenotips.data.Patient;`

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+        verify(this.logger).debug("Importing new patient from JSON via REST: {}", "");
+    }
+
+    @Test
+    public void addNullPatient() throws XWikiRestException {
+        doReturn(true).when(this.access).hasAccess(eq(Right.EDIT), any(DocumentReference.class), any(EntityReference.class));
+        Response response = this.patientsResource.addPatient(null);
+        Assert.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), response.getStatus());
+        verify(this.logger).error(eq("Could not process remote matching request: {}"), anyString(), anyObject());
+    }
```
DeanWay: Again, the Exception handling behaviour you are testing for here is occurring, but only because of a NullPointerException generated after `this.repository.createNewPatient();`. You need to mock the behaviour of this method call. Note that if you pass a null value to `addPatient()` the line `Patient patient = this.repository.createNewPatient();` will still execute since `JSONObject.fromObject(json);` will return a new JSONObject when passed a null value

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+        verify(this.logger).debug("Importing new patient from JSON via REST: {}", json.toString());
+    }
+
+    @Test
+    public void listPatientsNullOrderField() throws XWikiRestException {
+        WebApplicationException exception = new WebApplicationException();
+        try {
+            Patients result = this.patientsResource.listPatients(0, 30, null, "asc");
+        }
+        catch (WebApplicationException ex){
```
DeanWay: when `null` is pass for the `orderField` parameter it will simply be translated to "doc.name". No exception will be thrown

DeanWay: In fact, you should probably test that an exception *isn't* thrown, when the `orderField` or `order` parameters are null values.

sdumitriu: Yes, @DeanWay is right, there shouldn't be an exception when these are missing. Don't write tests that validate the current implementation, but think about the expected behavior and test that.

Tests are another way of providing a specification. The method name should be a short sentence describing what happens when calling a method within a specific environment/configuration. The mocks setup calls should clearly identify what that configuration is, and the assertions should show what's expected to happen — or NOT happen. The mocks should reflect a sane environment, even if the tested code path doesn't need every part of that environment.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
sdumitriu: And so the student becomes the master...

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,233 @@
...
+        doReturn(this.uri).when(this.uriInfo).getRequestUri();
+        ReflectionUtils.setFieldValue(this.patientsResource, "uriInfo", this.uriInfo);
+
+        doReturn("P00000001").when(this.patient).getId();
+        doReturn(this.currentUser).when(this.users).getCurrentUser();
+        doReturn(this.userProfileDocument).when(this.currentUser).getProfileDocument();
+    }
+
+    @Test
+    public void addPatientUserDoesNotHaveAccess() throws XWikiRestException {
```
DeanWay: `throws XWikiRestException` has been removed from the Class/Interface you are testing, remove them here as well

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,224 @@
...
+        doReturn(new ArrayList<Object[]>()).when(query).execute();
+        Patients result = this.patientsResource.listPatients(0, 30, "eid", "desc");
+        verify(this.queries).createQuery("select doc.fullName, p.external_id, doc.creator, doc.creationDate, doc.version, doc.author, doc.date"
+                + " from Document doc, doc.object(PhenoTips.PatientClass) p where doc.name <> :t order by "
+                + "p.external_id" + " desc", "xwql");
+    }
+
+    @Test
+    public void listPatientFailureHandling() throws QueryException {
+        WebApplicationException exception = new WebApplicationException();
```
DeanWay: I don't think creating a new `WebApplicationException` object is a good idea. Right now your tests are dependant on the implementation of the default constructor of `WebApplicationException`. One way that this could be problematic is if the status code is, by default, set to the same value you are testing for. Then you wouldn't be able to tell if it was this constructor, or the method you are testing that set that status code.
That's exactly what happens in this test. the `WebApplicationException` status code is set to `INTERNAL_SERVER_ERROR`, so this test will pass, whether or not the `WebApplicationException` is actually thrown.
Additionally, in the event that `listPatients()` doesn't throw a `WebApplicationException` under the condition that you have set up, this test will effectively give feedback that there was a `WebApplicationException` but it used the wrong status code. Which could be quite confusing.
To solve this problem you could either use the `@Test(expected = WebApplicationException.class)` notation and check the logger behaviour in another test, or use try/catch as you did here but initialize this value to `null`, set it to the caught exception and check whether it is still `null` afterward. This is the solution I used [here](https://github.com/phenotips/phenotips/pull/1802/files#diff-d4870c253487d0d0c680ad7578d14616R227)

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImplTest.java
```diff
@@ -0,0 +1,226 @@
...
+        doReturn(query).when(query).bindValue(anyString(), anyString());
+        doReturn(new ArrayList<Object[]>()).when(query).execute();
+        Patients result = this.patientsResource.listPatients(0, 30, "id", null);
+        verify(this.queries).createQuery("select doc.fullName, p.external_id, doc.creator, doc.creationDate, doc.version, doc.author, doc.date"
+                + " from Document doc, doc.object(PhenoTips.PatientClass) p where doc.name <> :t order by "
+                + "doc.name" + " asc", "xwql");
+    }
+
+    @Test
+    public void listPatientsNonDefaultBehaviour() throws WebApplicationException, QueryException {
```
DeanWay: `WebApplicationException` is a `RuntimeException` you don't need a `throws` clause for it

tmarathe:
@DeanWay take a look at the new tests?

DeanWay:
You should write some test cases for the usage of the `start` and `number` fields of `listPatients()`

tmarathe:
@DeanWay `listPatients` takes a start index and number of records to add, not a start and stop index.

PR #1824, opened 2015-07-14
Changed @inject for new AuthorizationModuleListProvider component

DeanWay:
@sdumitriu should this line be changed? Or will the current inject pick up the new implementation in `AuthorizationModuleListProvider `?


sdumitriu:
The code works correctly as it is.

Issue #1267, opened 2014-12-02
Clicking cancel on an empty patient should automatically delete it


buske:
Right now, if you create a new patient and immediately cancel it, you will be left with an empty patient record instead of no patient record. To help avoid littering the installation with empty patient records, empty records that are canceled should probably be automatically deleted (at least until the user modifies the record).

PR #1830, opened 2015-07-16
Issue #1267: Clicking cancel on an empty patient should automatically…

antonkatz:
… delete it.

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
@@ -92,7 +92,7 @@ public class PatientOnCancelDeleter extends AbstractEventListener
                 EntityReference defaultReference = new EntityReference(wiki.getDefaultPage(context),
                     EntityType.DOCUMENT, new EntityReference(wiki.getDefaultSpace(context), EntityType.SPACE));
                 defaultReference = currentResolver.resolve(defaultReference, EntityType.DOCUMENT);
-                String url = wiki.getURL(toDocumentReference(defaultReference), "view", context);
+                String url = wiki.getURL(new DocumentReference(defaultReference), "view", context);
                 context.getResponse().sendRedirect(url);
             }
         } catch (XWikiException ex) {
@@ -100,11 +100,6 @@ public class PatientOnCancelDeleter extends AbstractEventListener
         } catch (IOException ex) {
             logger.error("Could not redirect user appropriately. {}", ex.getMessage());
         }
-    }
-
-    protected DocumentReference toDocumentReference(EntityReference reference)
-    {
-        return new DocumentReference(reference);
     }
 
     private boolean isNew(XWikiDocument patient, XWikiContext context)
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
@@ -92,7 +92,7 @@ public class PatientOnCancelDeleter extends AbstractEventListener
                 EntityReference defaultReference = new EntityReference(wiki.getDefaultPage(context),
                     EntityType.DOCUMENT, new EntityReference(wiki.getDefaultSpace(context), EntityType.SPACE));
                 defaultReference = currentResolver.resolve(defaultReference, EntityType.DOCUMENT);
-                String url = wiki.getURL(new DocumentReference(defaultReference), "view", context);
+                String url = wiki.getURL(toDocumentReference(defaultReference), "view", context);
                 context.getResponse().sendRedirect(url);
             }
         } catch (XWikiException ex) {
@@ -100,6 +100,11 @@ public class PatientOnCancelDeleter extends AbstractEventListener
         } catch (IOException ex) {
             logger.error("Could not redirect user appropriately. {}", ex.getMessage());
         }
+    }
+
+    public DocumentReference toDocumentReference(EntityReference reference)
+    {
+        return new DocumentReference(reference);
     }
 
     private boolean isNew(XWikiDocument patient, XWikiContext context)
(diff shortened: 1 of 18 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
```diff
@@ -0,0 +1,129 @@
...
+            if (isNew(doc, context)) {
+                wiki.deleteDocument(doc, true, context);
+
+                EntityReference defaultReference = new EntityReference(wiki.getDefaultPage(context),
+                    EntityType.DOCUMENT, new EntityReference(wiki.getDefaultSpace(context), EntityType.SPACE));
+                defaultReference = currentResolver.resolve(defaultReference, EntityType.DOCUMENT);
+                String url = wiki.getURL(new DocumentReference(defaultReference), "view", context);
+                if (context.getResponse().isCommitted()) {
+                    context.getResponse().setHeader("Location", url);
+                    context.getResponse().flushBuffer();
```
antonkatz: @sdumitriu, the redirect does not work. It's as if the `Location` gets overwritten later on, or is already sent to the client. I have no more ideas to try. Could you fix it or tell someone how to fix it? I'm going away friday, and don't think I'll have the opportunity to fix this tomorrow. Thank you.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
```diff
@@ -0,0 +1,129 @@
...
+        try {
+            XWikiDocument doc = wiki.getDocument(patient.getDocument(), context);
+            if (isNew(doc, context)) {
+                wiki.deleteDocument(doc, true, context);
+
+                EntityReference defaultReference = new EntityReference(wiki.getDefaultPage(context),
+                    EntityType.DOCUMENT, new EntityReference(wiki.getDefaultSpace(context), EntityType.SPACE));
+                defaultReference = currentResolver.resolve(defaultReference, EntityType.DOCUMENT);
+                String url = wiki.getURL(new DocumentReference(defaultReference), "view", context);
+                if (context.getResponse().isCommitted()) {
```
sdumitriu: That's because you're missing a `!` here.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
```diff
@@ -0,0 +1,129 @@
...
+                    context.getResponse().setHeader("Location", url);
```
sdumitriu: Use `.sendRedirect(url)`, which sets the right status code, the `Location` header, and calls `flushBuffer()`.

antonkatz: It fails if the response `isCommitted`

sdumitriu: If the response is committed, then you can't change it in any way. But look again at the condition inside the `if`, you're missing the `!`.

sdumitriu: Actually the problem is that the `ActionExecutedEvent` is sent **after** the response has been processed, you should listen to `ActionExecutingEvent` instead.

sdumitriu:
@antonkatz Can you address the latest comments?

antonkatz:
Right, will do.

allasm:
Just found this pull request, is it still valid and if yes whats the status?

sdumitriu:
It's still valid, waiting for review.

marta-:
@veronikaslc @meatcar please review.

antonkatz:
@allasm this needs review, I'll rebase it right now

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientOnCancelDeleter.java
```diff
@@ -0,0 +1,124 @@
...
+    private boolean isNew(XWikiDocument patient, XWikiContext context)
+    {
+        // to be safe and not delete in case of error.
+        boolean hasVersion = false;
+        int versionCount = 0;
+        try {
+            List<String> versions = patient.getRevisions(new RevisionCriteria(), context);
+            for (String version : versions) {
+                XWikiRCSNodeInfo info = patient.getRevisionInfo(version, context);
+                if (!StringUtils.equalsIgnoreCase(info.getComment().trim(), AUTOSAVE_COMMENT) && !info.isMinorEdit()) {
```
allasm: What are minor edits? Can it be that there is one edit (document created) and then a bunch of minor edits, and then user clicks edit, cancel, and document gets deleted?

allasm: As discussed with @antonkatz the act of creating a patient adds a version, and clicking save adds another, so supposedly if ANY changes have been saved at least once there will be at least two non-minor versions. TODO: check what kind of edit is permission changes?

allasm: Another potential problem: User enters data, document gets auto-saved, user navigates away thinking everything is saved

problem #1: @antonkatz says navigating away triggers a cancel event, which may delete the document contradicting user expectations

problem #2: even if not, if the user later comes back, edits, cancels the document gets deleted at that point

allasm:
XwikiDocument.java is committed as one file, and it is hard to know what was the change compared to the original. We checked and main XWiki github still does not have the change, so it is important to separate the change so that it is easy(er) to re-apply it after a Phenotips XWiki upgrade

For the record, the change is:

3314c3431,3433
<         setComment(StringUtils.defaultString(comment));
---
>         if (comment != null) {
>             setComment(comment);
>         }

which fixes a bug where XWiki would use the comment from the previous edit in case current edit has no comment.

antonkatz:
There is code for VCF's in this pull request. I don't remember why it's there, my best guess is that it is supposed to prevent an extra save: right now, when a new patient document is created, there is code that executes and adds a VCF object to the document. 

https://github.com/phenotips/phenotips/blob/master/components/patient-data/ui/src/main/resources/PhenoTips/VCF.xml#L208

veronikaslc:
isn't working as expected, throwing various errors when cancelling or saving patient documents

allasm:
Is "throwing various errors" expected?

veronikaslc:
I guess no error is expected

Issue #1833, opened 2015-07-17
When death date is checked as unknown exported patient JSON still has the last entered date of death


allasm:
Related to #1662 - the date of death, if entered and saved before the "date of death is unknown" is checked is hidden from the view but is still present in the patient object in XWiki, and thus is still included in some of the export formats (e.g. JSON)

PR #1837, opened 2015-07-20
Issue #1833: Fixed death date value not being removed from xwiki patient object on save if unknown date of death is selected.

allasm:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientDeathdateUpdater.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        XWikiDocument doc = (XWikiDocument) source;
+
+        BaseObject patientRecordObj = doc.getXObject(Patient.CLASS_REFERENCE);
+        if (patientRecordObj == null) {
+            return;
+        }
+        boolean unknownDeathDate = getParameter("date_of_death_unknown", patientRecordObj.getNumber());
```
sdumitriu: Instead of checking for a parameter sent in the request, why not check the actual object? That way the death date will be cleared no matter how the patient is updated: REST, pushing from another server, updating through scripts...

    if (patientRecordObj.getIntValue("date_of_death_unknown") == 1)

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientDeathdateUpdater.java
```diff
@@ -0,0 +1,98 @@
...
+        if (unknownDeathDate)
+        {
+            patientRecordObj.setDateValue("date_of_death", null);
+            patientRecordObj.setDateValue("date_of_death_entered", null);
```
sdumitriu: `date_of_death_entered` is not a date property, but a string. Use `setStringValue` instead.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PatientDeathdateUpdater.java
allasm: Definitely a good idea. This was done this way basically because I did not know how/when exactly onEvent hooks work, and just copied the code from another updater (BirthDateUpdater - should we update that one in a similar manner as well?)

sdumitriu: No, since that one transforms fields that are only present in the request into a value that can be stored in the patient object.

PR #1840, opened 2015-07-21
Use equals() for string comparison

DeanWay:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/LifeStatusController.java
```diff
@@ -122,7 +122,7 @@ public void save(Patient patient)
             PatientData<Date> dates = patient.getData("dates");
 
             Integer deathDateUnknown = 0;
-            if (lifeStatus != null && lifeStatus.getValue() == DECEASED) {
+            if (lifeStatus != null && lifeStatus.getValue().equals(DECEASED)) {
```
sdumitriu: Put it the other way around: `DECEASED.equals(lifeStatus.getValue())`

PR #1845, opened 2015-07-22
writeJSON when passed collection of selectedFieldNames should only ad…

DeanWay:
…d the selectedFieldNames to the JSON


sdumitriu:
See line 110. Apgar scores go together, so they shouldn't be separated.

Issue #1274, opened 2014-12-05
Write tests for various patient data controllers


antonkatz:
It appears that there is not a single test for any of the controllers

PR #1847, opened 2015-07-22
Issue #1274

DeanWay:
Tests for DatesController and APGARController. Tests for the remaining PatientDataController implementations will be added in separate pull requests


DeanWay:
@JKereliuk @tmarathe please review

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/APGARControllerTest.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Test
+    public void loadDoesNotReturnNonIntegerStrings()
+    {
+        doReturn("STRING").when(this.data).getStringValue(anyString());
+
+        PatientData<Integer> result = this.controller.load(this.patient);
+
+        Assert.assertEquals(0, result.size());
```
tmarathe: Instead of this, `Assert.assertTrue(result.isEmpty())` might make what you're testing more clear.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/APGARControllerTest.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Test
+    public void writeJSONReturnsWhenGetDataReturnsNull() {
+        doReturn(null).when(this.patient).getData(DATA_NAME);
+        JSONObject json = new JSONObject();
+
+        this.controller.writeJSON(this.patient, json);
+
+        Assert.assertTrue(json.getJSONObject(DATA_NAME) == null || json.getJSONObject(DATA_NAME).isNullObject());
```
tmarathe: What's the point of this? Aren't both sides of the or statement the same?

sdumitriu: No, one isn't at all (`null`), the other is a special JSONObject that represents `null`.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/APGARControllerTest.java
DeanWay: result uses the `PatientData<T>` interface. `isEmpty()` is not a defined method for that interface

PR #1850, opened 2015-07-24
Issue #1274 simple value controllers

DeanWay:
Tests for PatientDataControllers that use SimpleValuePatientData

### DIFF
--- a/components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
+++ b/components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
@@ -1,0 +1,195 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


DeanWay:
@JKereliuk @tmarathe please review

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
```diff
@@ -0,0 +1,229 @@
...
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.phenotips.data.Patient;
+import org.phenotips.data.PatientData;
+import org.phenotips.data.PatientDataController;
+import org.phenotips.data.SimpleValuePatientData;
+import org.slf4j.Logger;
+import org.xwiki.bridge.DocumentAccessBridge;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.test.mockito.MockitoComponentMockingRule;
```
sdumitriu: Wrong [import order](https://github.com/phenotips/build-tools/blob/master/codestyle-resources/src/main/resources/eclipse.importorder).

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
```diff
@@ -0,0 +1,229 @@
...
+    private final String UNAFFECTED = "unaffected";
+
+    private final String DATA_NAME = "clinicalStatus";
+
+    @Before
+    public void setUp() throws Exception
+    {
+        MockitoAnnotations.initMocks(this);
+
+        this.controller = (ClinicalStatusController)this.mocker.getComponentUnderTest();
```
sdumitriu: Why do you need to cast it? You should only work with the interface, which you should always get fresh as `mocker.getComponentUnderTest()`.

PR #1857, opened 2015-08-07
Suppress cast warnings in AbstractComplexController

DeanWay:
Changed `AbstractComplexController` to avoid the following compiler warnings:
```
[WARNING] /phenotips/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java:[78,17] found raw type: com.xpn.xwiki.objects.BaseProperty
  missing type arguments for generic class com.xpn.xwiki.objects.BaseProperty<R>
[WARNING]/phenotips/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java:[91,50] unchecked cast
  required: T
  found:    java.lang.Object
[WARNING] /phenotips/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java:[164,50] unchecked cast
  required: java.util.List<T>
  found:    java.lang.Object
```

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
@@ -25,7 +25,7 @@ import org.phenotips.data.VocabularyProperty;
 import org.phenotips.data.internal.AbstractPhenoTipsVocabularyProperty;
 
 import org.xwiki.bridge.DocumentAccessBridge;
-import org.xwiki.model.reference.EntityReference;
+import org.xwiki.model.reference.ObjectPropertyReference;
 
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -75,13 +77,12 @@
             }
             Map<String, T> result = new LinkedHashMap<String, T>();
             for (String propertyName : getProperties()) {
-                BaseProperty field = (BaseProperty) data.getField(propertyName);
+                BaseProperty<EntityReference> field = (BaseProperty<EntityReference>) data.getField(propertyName);
```
sdumitriu: This should be `ObjectPropertyReference` instead of `EntityReference`

PR #1854, opened 2015-07-31
was not sure where to put these in the xml files

JKereliuk:
The homepage must use both of these extensions

PR #1860, opened 2015-08-10
Jira issue PT-62: Show incompatibility page for IE 8 users

DeanWay:
Updated from @JKereliuk's pull request #1854


components/skin/ui/src/main/resources/PhenoTips/OutdatedBrowser.xml
```diff
@@ -0,0 +1,341 @@
...
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>default</cache>
+    </property>
+    <property>
+      <code>var browser = navigator.userAgent.toLowerCase();&#xd;
```
sdumitriu: Run `mvn xar:format` once again to get rid of these `&#xd;` at the end of lines.

components/skin/ui/src/main/resources/PhenoTips/OutdatedBrowser.xml
```diff
@@ -0,0 +1,341 @@
...
+        if (browser.indexOf("msie") != -1 &amp;&amp; parseInt(browser.split("msie")[1]) &lt;= 8)&#xd;
```
sdumitriu: There already is a `browser` global variable which does better browser detection. Remove the first line and use:

    if (browser.isIE && browser.versionMajor < 9)

components/skin/ui/src/main/resources/PhenoTips/OutdatedBrowser.xml
```diff
@@ -0,0 +1,341 @@
...
+        {&#xd;
+        // remove html on page&#xd;
+        document.body.innerHTML = "";&#xd;
+        &#xd;
+        &#xd;
+        // make the banner&#xd;
+        var logo = document.createElement("img");&#xd;
+        logo.src = "https://ccm.sickkids.ca/wp-content/uploads/2015/02/phenotips-logo-300x56.png";&#xd;
```
sdumitriu: Don't depend on external resources, the user might be behind a firewall.

components/skin/ui/src/main/resources/PhenoTips/OutdatedBrowser.xml
```diff
@@ -0,0 +1,341 @@
...
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>#container&#xd;
```
sdumitriu: This CSS affects too many generic elements, and is always active. It has a high chance of having unintended consequences. Make it more specific.

components/skin/ui/src/main/resources/PhenoTips/OutdatedBrowser.xml
sdumitriu: Really weird indentation, our codestyle is to indent by two spaces.
Open curlies on the same line as the `if` statement.
Subsequent closing curlies and new rules should have 0 indentation.

sdumitriu:
Applied a simpler fix in 351f2eebba8241f758a112b0ffbc174d2230f1fe

What I don't like about this:
- too much javascript to generate a simple HTML
- most people using IE8 are doing so because it's mandated by their institution, they can't just download a new browser; providing links won't make the upgrade easier in any way
- the UI doesn't fit at all with the rest of the PhenoTips skin
- completely hiding the normal PhenoTips UI is a bit too extreme

marta-:
> most people using IE8 are doing so because it's mandated by their institution, they can't just download a new browser; providing links won't make the upgrade easier in any way

False. Many are allowed to use Chrome. Showing icons of browsers they can use is definitely helpful.

sdumitriu:
OK, but saying that they should download such a browser is not helpful.

Issue #1581, opened 2015-04-16
In "Login" page password field is not outlined with blue when on-focus


veronikaslc:
"Login" page: if username input field is on-focus, then it is outlined with blue border line, but no outlining for the password field - should be consistent

PR #1865, opened 2015-08-18
Issue #1581: UI: In "Login" page password field is not outlined with …

mjshepherd:
…blue when on-focus


components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -481,7 +481,7 @@ textarea {
   box-sizing: border-box;
   width: 100%;
 }
-input[type=text]:focus, textarea:focus {
+input[type=text]:focus, input[type=password]:focus, textarea:focus {
```
sdumitriu: See line 474 above as well.

Issue #1758, opened 2015-06-11
Display suggested phenotypes in a browsable list of pages


sdumitriu:
This allows users to see other results outside the top 20.

CourtneyHum:
Do you mean the phenotypes with the lightning bolt? Deleting them is already an option.

sdumitriu:
@CourtneyHum No, the phenotypes showing under "You may want to investigate...", for which the only options currently available are Y and N, which will lead to many negative annotations, something we don't really want.

sdumitriu:
@mjshepherd Useful: https://github.com/phenotips/clinical-text-analysis-extension/blob/master/ui/src/main/resources/PhenoTips/Clinical%20Notes%20Annotation.xml#L332

PR #1869, opened 2015-08-20
Issue #1758: Users should be allowed to discard suggested phenotypes …

mjshepherd:
…-- suggested phenotypes are now paginated

### DIFF
--- a/components/solr-access-service/ui/src/main/resources/PhenoTips/DiffDiagnosisService.xml
+++ b/components/solr-access-service/ui/src/main/resources/PhenoTips/DiffDiagnosisService.xml
@@ -44,7 +44,11 @@
     #set ($_service = $services.omim)
     #set ($limit = $mathtool.toInteger("$!{request.limit}"))
     #if (!$limit || $limit &lt; 0)
-      #set ($limit = 20)
+      #set ($limit = 6)
+    #end
+    #set($page = $mathtool.toInteger("$!{request.page}"))
+    #if (!$page || $page &lt; 0)
+      #set ($page = 1)
     #end
     #set ($data = $collectionstool.set)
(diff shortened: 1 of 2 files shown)


mjshepherd:
@sdumitriu after speaking some discussion it was decided that the suggestion should be paginated so that users can scroll through. The discard mechanism in biolark removes the suggestion from the patient permanently which is decidedly too drastic.

mjshepherd:
I nominate @itaiGershtansky  @tmarathe for review

tmarathe:
:+1:

itaiGershtansky:
It's more readable to have the changes committed in smaller units, specifically separated to styling and functionality. Other than that, +1.

tmarathe:
When you select and then deselect the term (within the 0.8 second delay) it does not get removed from the current selection. This was an issue in the previous iteration of this section as well.

components/solr-access-service/ui/src/main/resources/PhenoTips/DiffDiagnosisService.xml
```diff
@@ -55,29 +59,46 @@
...
-{{/html}}
-   #set ($hasOutput = true)
-#else
-   #set ($hasOutput = false)
-#end## results.size() &gt; 0
+      #set ($resultsToFetch = $limit * $page)
+      #set ($results = $_service.getDifferentialPhenotypes($data, $ndata, $resultsToFetch))
+      #if ($results.size() &gt; 0)
+        ##paginate results
+        #set ($from = ($page - 1) * $limit)
```
sdumitriu: It may be possible that less than the number of requested pages are returned, in which case `subList` will throw an exception.

sdumitriu:
Manually applied and rebased to match the new issue title.

PR #1871, opened 2015-08-20
Changed GeneList and RejectedGeneList Controllers to send logger mess…

DeanWay:
…ages when gene class is not found


sdumitriu:
The majority of controllers work with data inside the `PatientClass`, and that one is fundamental for a patient record. Thus, it's normal for most controllers to complain if a patient record doesn't have a `PatientClass`. On the other hand, genes are optional, it's normal for a patient record not to have any objects of type `InvestigationClass`.

However, there should be a `logger.debug` statement indicating that no such objects exist.

Issue #1280, opened 2014-12-08
One unified gene section with significance specified per gene


buske:
Instead of separate candidate gene and rejected gene sections, we should have one section for all gene-related information.

'Add a new gene' should bring up a row with the options in #1149.

buske:
Should include causal gene (and perhaps PMID), a la #1420. PMID pending discussion with Bailey/Taila; might be better elsewhere.

PR #1873, opened 2015-08-24
Issue #1280: One unified gene section with significance specified per gene.

veronikaslc:
Fixed.


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -0,0 +1,407 @@
...
+        Collection<String> selectedFieldNames)
+    {
+        int count = 0;
+        for (String property : keys) {
+            if (StringUtils.isBlank(item.get(property))
+                || (selectedFieldNames != null
+                && !selectedFieldNames.contains(enablingProperties.get(count)))) {
+                item.remove(property);
+            }
+            count++;
```
DeanWay: Rather than implementing `enablingProperties` as a list, perhaps it would be better to implement it as a map of property_name --> enabling_field_name. That way this code is not dependent on the indices of the `keys` list and the `enablingProperties` list matching up

antonkatz:
How come there's only one commit?

components/export/api/pom.xml
```diff
@@ -29,7 +29,7 @@
   <name>PhenoTips - Data export - Spreadsheet Export APIs</name>
 
   <properties>
-    <coverage.instructionRatio>0.20</coverage.instructionRatio>
+    <coverage.instructionRatio>0.18</coverage.instructionRatio>
```
antonkatz: Is there a good reason for this?

components/patient-data/impl/pom.xml
```diff
@@ -29,7 +29,7 @@
   <name>PhenoTips - Patient data - Default implementation for the Java APIs</name>
 
   <properties>
-    <coverage.instructionRatio>0.42</coverage.instructionRatio>
+    <coverage.instructionRatio>0.13</coverage.instructionRatio>
```
antonkatz: This is too much.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -70,27 +70,17 @@ public void phenotypeSetup(Set<String> enabledFields) throws Exception
     {
         String sectionName = "phenotype";
         String[] fieldIds = { "phenotype", "phenotype_code", "phenotype_combined", "phenotype_code_meta",
-            "phenotype_meta", "negative_phenotype", "negative_phenotype_code", "negative_phenotype_combined",
-            "phenotype_by_section" };
+        "phenotype_meta", "negative_phenotype", "negative_phenotype_code", "negative_phenotype_combined",
+        "phenotype_by_section" };
```
antonkatz: Formatting change, don't commit. Same right below.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1359,4 +1427,41 @@ private String getUsername(DocumentReference reference)
...
+        return present;
+    }
+
+    private String parseEvidence(String value)
+    {
+        Map<String, String> valueTranslates = new HashMap<String, String>();
+        valueTranslates.put("rare", "Rare (MAF<0.01); ");
+        valueTranslates.put("predicted", "Predicted damaging by in silico models; ");
+        valueTranslates.put("reported", "Reported in other individuals; ");
+        valueTranslates.put("segregates", "Segregates with phenotype within family; ");
```
antonkatz: These need to be translated. 
```     @Inject
    private LocalizationManager localization;
    @Inject
    private LocalizationContext localizationContext;
    
     this.localization.getTranslation(key, this.localizationContext.getCurrentLocale());
```

sdumitriu: There's something better, https://github.com/phenotips/phenotips/blob/master/components/xwiki-translation-manager/src/main/java/org/phenotips/translation/TranslationManager.java#L32

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -283,100 +287,222 @@ public DataSection genesHeader() throws Exception
...
 
-        if (present.contains("comments") || present.contains("rejected_comments")) {
-            cell = new DataCell("Comments", hX, 1, StyleOption.HEADER);
+        List<String> fields =
+            Arrays.asList("protein", "transcript", "dbsnp", "zygosity", "effect", "interpretation", "inheritance",
+                "evidence", "segregation", "sanger", "resolution");
+        Map<String, String> headerTranslates = new HashMap<String, String>();
+        headerTranslates.put("dbsnp", "db SNP");
+        headerTranslates.put("segregation", "Segregation Studies");
+        headerTranslates.put("sanger", "Sanger validation");
```
antonkatz: Translations. See below.

components/export/api/src/main/java/org/phenotips/export/internal/SheetAssembler.java
```diff
@@ -212,8 +215,8 @@ private DataSection assembleSectionsY(List<DataSection> sections, Boolean finali
     }
 
     /**
-     * @return a {@link org.phenotips.export.internal.DataSection} that contains all {@link
-     * org.phenotips.export.internal.DataCell}s
+     * @return a {@link org.phenotips.export.internal.DataSection} that contains all
+     *         {@link org.phenotips.export.internal.DataCell}s
```
antonkatz: Cosmetic change.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -0,0 +1,274 @@
...
+        } else {
+            BaseStringProperty field = (BaseStringProperty) variantObject.getField(property);
+            if (field == null) {
+                return null;
+            }
+            return field.getValue();
+        }
+    }
+
+    private void removeKeys(Map<String, String> item, List<String> keys, Map<String, String> enablingProperties,
```
antonkatz: Is this function not identical to the one in the file above? Maybe factor it out to a separate file?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/SolvedControllerTest.java
```diff
@@ -93,14 +93,13 @@ public void checkGetProperties() throws ComponentLookupException
...
-        Assert.assertEquals(4, result.size());
+        Assert.assertEquals(3, result.size());
         Assert.assertThat(result, Matchers.hasItem(STATUS_KEY));
         Assert.assertThat(result, Matchers.hasItem("solved__pubmed_id"));
-        Assert.assertThat(result, Matchers.hasItem("solved__gene_id"));
         Assert.assertThat(result, Matchers.hasItem("solved__notes"));
     }
 
-    //--------------------save() is Overridden from AbstractSimpleController--------------------
+    // --------------------save() is Overridden from AbstractSimpleController--------------------
```
antonkatz: Cosmetic change. Same right below.

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/PatientEmptyObjectsRemover.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.inject.Singleton;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.objects.BaseStringProperty;
+
+/**
+ * Removes empty objects from the document.
```
antonkatz: Feels inelaborate

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -283,100 +287,222 @@ public DataSection genesHeader() throws Exception
...
+
+        PatientData<Map<String, String>> allGenes = patient.getData("genes");
+
+        // empties should be created in the case that there are no genes to write
+        if (allGenes == null || !allGenes.isIndexed()) {
+            /* Status and gene name columns are always present */
             Integer emptyX = 0;
-            for (int i = 0; i < columns; i++) {
-                DataCell cell = new DataCell("", emptyX, 0);
+            for (int i = 0; i < present.size(); i++) {
```
meatcar: you don't use `i` at all here. Maybe you should use it instead of `emptyX`, since they are identical.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1359,4 +1427,41 @@ private String getUsername(DocumentReference reference)
...
+        String field = "";
+        for (String property : valueTranslates.keySet()) {
+            if (value.contains(property)) {
+                field += valueTranslates.get(property);
+            }
+        }
+        return field;
+    }
+
+    private String capitalizeValue(String value)
```
meatcar: don't use https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#capitalize%28java.lang.String%29 instead.

components/export/api/src/main/java/org/phenotips/export/internal/SheetAssembler.java
meatcar: i think these kinds of changes are fine if they improve code readability.

antonkatz: I say that's a question for @sdumitriu

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
```diff
@@ -120,105 +120,79 @@
...
-             var geneLinks = element.next('div.gene-links');
-             if (geneLinks) {
-               geneLinks.removeClassName('loading');
-             }
-           });
-         }
-       }
-    );
+  var updateInfo = function(el, value, isSuggestSelected) {
+    var newValue = (!value || value == "") ? el.value : value;
```
meatcar: in javascript, '""' is falsey, which will be caught by '!value'. This statement is effectively creating a default parameter, which you could do with a short-circuiting trick. It's a pretty common pattern.

```js
var value = value || el.value;
```

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+#set($rawValue = "")
+|(% class="variant-moreinfo variant-property-name-$variantIndex %)**${prop.translatedPrettyName}**| (% class="variant-property-value ${prop.getName()}-$variantIndex variant-label-$variantIndex" %) #if($isInputString) $rawValue #else $doc.display($prop.getName(), 'view', $fakeObj )#end (%%) (% class="v-collapsed variant-input-$variantIndex" %)$doc.display($prop.getName(), $options.mode, $fakeObj )#end##
+#end
+))) (% class="variant-moreinfo-editdonebutton-row variant-$variantIndex" %)\\
+#end
+##
+##
+##
+#macro (__variant_row $variantObj $count $variantIndex $geneObjNumber $options)
+#if ($options.mode == 'edit')|(% class="v-collapsed" %) $doc.display('genesymbol', $options.mode, $variantObj)#end|(% class="pseudoindent" %) |(% class="variant-row-count variant-${variantIndex} variant" %)$count#foreach($propName in $options.defaultProperties)|(% class="variant variant-default-input $propName variant-$variantIndex" %)$doc.display($propName, $options.mode, $variantObj)#end#if ($options.mode == 'edit')#__variant__deleteTool($variantObj)#end\\
```
meatcar: can we try and break up lines like this, so that it's easier to read and see what's going on? Not sure how we should go about it.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      var el = null;
+      if (event.element().className == 'variant moreinfo')
+        el = event.element();
+      else
+        el = event.element().up('td.variant.moreinfo');
+      el.select('a.variant-edit').invoke('click');
+      var className = el.down('.variant-moreinfo-editbutton-row').className;
+      var varIndex = className.substring(className.lastIndexOf('-')+1);
+      event.stopPropagation();
+      document.observe('click', function (event) {
```
meatcar: This listener hangs around forever, so if you call `areaEditData` multiple times, there would be multiple listeners attached. Not sure if this was intended.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      }
+      
+      var geneRowTemplateEl = this.newGeneRowElement;
+      var buttonRowTemplateEl = this.addVariantButtonElement;
+
+      var idx = dataTable.select('td.row-count').size() + 1;
+      var geneIndex = 0;
+      if (dataTable.select('td.row-count').size() > 0) {
+        var className = dataTable.select('td.row-count')[idx - 2].next().className;
+        geneIndex = parseInt(className.substring(className.lastIndexOf('-')+1)) + 1;
```
meatcar: when using `parseInt`, you need to specify a radix, since many JS implementations treat a string with a leading 0 as an octal, which causes weird bugs.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      if (!variantFooter) {
+        new XWiki.widgets.Notification("$services.localization.render('phenotips.tableMacros.listNotFound')" + ' ' + classname, 'error');
+      }
+      
+      var varCount = 1;
+      var geneCount = parseInt($$('td.gene.gene-'+geneIndex)[0].up('tr').down('td').innerHTML);
+      var varIndex = 0;
+      $$('.variant-moreinfo-editbutton-row').each( function(item) {
+        var indx = parseInt(item.className.substring(item.className.lastIndexOf('-')+1));
+        varIndex = (indx > varIndex) ? indx : varIndex;
```
meatcar: use `Math.max()`

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      var geneIndex = 0;
+      if (dataTable.select('td.row-count').size() > 0) {
+        var className = dataTable.select('td.row-count')[idx - 2].next().className;
+        geneIndex = parseInt(className.substring(className.lastIndexOf('-')+1)) + 1;
+      }
+
+      var newGeneRow = new Element('tr');
+      var newButtonRow = new Element('tr', {'class' : 'variant-gene-'+geneIndex+' variant-footer v-collapsed'}); 
+
+      var geneRowInner = geneRowTemplateEl.innerHTML.replace( /GENE_NUMBER_PLACEHOLDER/g, idx).replace( /GENE_INDEX_PLACEHOLDER/g, geneIndex).replace(/_\d+_/g, '_'+geneIndex+'_').replace(/=\d+&amp;/g, '='+geneIndex+'&amp;');
```
meatcar: looks like you're rolling your own little templating library. Thing is, chaining string.replace functions is inefficient. I'd recommend you look into using a proper front-end templating solution, maybe jquery templates.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+        var variantIndex = td.className.substring(td.className.lastIndexOf('-')+1);
+        var showMoreinfoWrapper = new Element('div', {'class' : 'show-moreinfo-button triRight', 'title' : "$services.localization.render('PhenoTips.GeneVariantClass.showMoreinfo.hint')", id : 'PhenoTips.GeneVariantClass_'+variantIndex+'_showMoreinfo'});        
+        showMoreinfoWrapper.observe('click', function (event) {
+          event.stop();
+          event.element().up('tr').next().toggleClassName('v-collapsed');
+          if (event.element().up('tr').next().hasClassName('v-collapsed')) {
+            event.element().toggleClassName('triRight', true);
+            event.element().toggleClassName('triDown', false);
+          }
+          else{
```
meatcar: else shouldn't be on a new line

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      var id = event.element().id;
+      var variantIndex = id.substring(id.indexOf('_')+1, id.lastIndexOf('_'));
+      $$('.gene-table tr .variant-'+variantIndex+' tr').each( function(item) {
+        item.toggleClassName('moreinfo-view', true);
+      }); 
+      event.element().toggleClassName('v-collapsed', true); 
+      $(this.geneVariantClassName+'_'+variantIndex+'_edit').toggleClassName('v-collapsed', false);
+      var inputs = $$('.variant-input-'+variantIndex);
+      var labels = $$('.variant-label-'+variantIndex);
+      for (var i=0;i&lt;inputs.length;i++) {
```
meatcar: space out the operations for readability.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+    }
+  });
+  return ExtraGeneVariantData;
+}(ExtraGeneVariantData || {}));
+
+document.observe('xwiki:dom:loaded', function() {
+  new ExtraGeneVariantData.tools.Editor();
+})
+
+/*---TODO-- warning user for unfilled cDNA and gene name fields---
```
meatcar: remove commented out code.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+      element.select('.variant.variant-default-input.cdna input').each(function(input) {
+        if (!input.__Variant_validator) {
+          input.__Variant_validator = new XWiki.widgets.VariantValidator(input);
+        }
+      });
+    });
+    return true;
+  };
+
+  (XWiki.domIsLoaded &amp;&amp; init()) || document.observe("xwiki:dom:loaded", init);
```
meatcar: since this is not an assignment, you should probably use an if statement.

marta-: Not necessarily. This is shorter and consistent with _all_ other widget initialization lines in PhenoTips.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -283,100 +287,222 @@ public DataSection genesHeader() throws Exception
...
+        valueTranslates.put("insertion_frameshift", "insertion - frameshift");
+        valueTranslates.put("indel_in_frame", "indel - in frame");
+        valueTranslates.put("indel_frameshift", "indel - frameshift");
+        valueTranslates.put("repeat_expansion", "repeat expansion");
+        List<String> upperCase =
+            Arrays.asList("pathogenic", "benign", "complete", "rejected", "unknown", "missense", "nonsense",
+                "duplication", "synonymous", "other", "maternal", "paternal");
+
+        for (Map<String, String> variant : variants) {
+            int j = 0;
```
meatcar: you use `i` and `j` and `hX` and `emptyX` as a counter, interchangeably. You should pick one and stick with it.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+    ajaxAddVariantData : function (event) {
+      event.stop();
+      var addTrigger = event.element();
+      if (addTrigger.disabled) 
+         return;
+      var classname = '';
+      try {
+        var classname = /classname=([^&amp;]+)/.exec(addTrigger.href)[1];
+      } catch (err) {
+         new XWiki.widgets.Notification('Cannot add data: type not found', 'error');
```
meatcar: if this is going to be shown to the user, it should be localized. If this is an internal error, no banner should be shown to the user.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1420 @@
...
+          this.invalid();
+        } else {
+          this.available();
+        }
+        this.responded();
+      }
+    },
+    validate : function(value) {
+      if (this.state == 'DONE') {
+        this.value == value &amp;&amp; (this.valid || Validate.fail("This variant has already been entered."));
```
meatcar: error string should be localized.

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
sdumitriu: [Prototype.js has that too](http://api.prototypejs.org/language/String/prototype/interpolate/).

components/export/api/src/main/java/org/phenotips/export/internal/SheetAssembler.java
sdumitriu: The proper syntax is the changed one. However, Anton is right, cosmetic changes shouldn't be included along with code in the same pull request, but in their own pull request. Otherwise, it's harder to review.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
veronikaslc: Should I make a separate issue to translate all places in a file? Seems nothing is translated there yet (see phenotypeHeader(), documentInfoHeader(),patientInfoHeader(), familyHistoryHeader(), etc).

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
veronikaslc: are there any guidelines how to safely break up lines here and not generate anything new?

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
sdumitriu: Yes, better leave that as a separate issue.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
sdumitriu: Well, `var value = value...` doesn't need the `var` at all, right?

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
sdumitriu: Split any line that doesn't end with a velocity directive with `##` followed by a return, as in:

```
#if ($options.mode == 'edit')
|(% class="v-collapsed" %) $doc.display('genesymbol', $options.mode, $variantObj)##
#end
|##
(% class="pseudoindent" %) |##
(% class="variant-row-count variant-${variantIndex} variant" %)$count##
#foreach($propName in $options.defaultProperties)
|(% class="variant variant-default-input $propName variant-$variantIndex" %)$doc.display($propName, $options.mode, $variantObj)##
#end
#if ($options.mode == 'edit')
#__variant__deleteTool($variantObj)##
#end\\
```

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/PatientEmptyObjectsRemover.java
```diff
@@ -0,0 +1,115 @@
...
+            List<BaseObject> xWikiObjects = doc.getXObjects(refs.get(key));
+            if (xWikiObjects == null || xWikiObjects.isEmpty()) {
+                continue;
+            }
+            for (BaseObject object : xWikiObjects) {
+                if (object == null) {
+                    continue;
+                }
+                BaseStringProperty field = (BaseStringProperty) object.getField(key);
+                if (field == null || "".equals(field.getValue())) {
```
antonkatz: Use ```StringUtils.isEmpty``` rather than ```"".equals```

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/PatientEmptyObjectsRemover.java
```diff
@@ -0,0 +1,115 @@
...
+                    doc.removeXObject(object);
+                    try {
+                        context.getWiki().saveDocument(doc, "Removed empty object", true, context);
```
antonkatz: Should this not be outside the loop?

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/PatientEmptyObjectsRemover.java
```diff
@@ -0,0 +1,115 @@
...
+    public void onEvent(Event event, Object source, Object data)
+    {
+        XWikiDocument doc = (XWikiDocument) source;
+        XWikiContext context = (XWikiContext) this.execution.getContext().getProperty("xwikicontext");
+
+        Map<String, EntityReference> refs = new LinkedHashMap<String, EntityReference>();
+        refs.put(GENE_KEY, GENE_CLASS_REFERENCE);
+        refs.put(VARIANT_KEY, VARIANT_CLASS_REFERENCE);
+
+        for (String key : getKeys()) {
```
antonkatz: Am I missing something or could you just iterate over the ```refs``` map rather than ```getKeys```?

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
```diff
@@ -0,0 +1,290 @@
...
+            R70190PhenoTips1280DataMigration.this.entityResolver.resolve(GENE_CLASS);
+        DocumentReference rejectedGenesClassReference =
+            R70190PhenoTips1280DataMigration.this.entityResolver.resolve(REJECTED_CLASS);
+
+        Query q =
+            session.createQuery("select distinct o.name from BaseObject o, StringProperty p where o.className = '"
+                + this.serializer.serialize(geneClassReference) + OR
+                + this.serializer.serialize(rejectedGenesClassReference) + OR
+                + this.serializer.serialize(patientClassReference) + "' and p.id.id = o.id and p.id.name = '"
+                + SOLVED_NAME + "' and p.value <> ''");
```
antonkatz: Why does ```p.id.name``` have to equal ```SOLVED_NAME```

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
```diff
@@ -0,0 +1,290 @@
...
+            setSolvedGenes(doc, xwiki, patientClassReference, geneClassReference, context, session, geneList);
+            setRejectedGenes(doc, xwiki, rejectedGenesClassReference, geneClassReference, context, session, geneList);
+            setCandidateGenes(doc, xwiki, investigationClassReference, geneClassReference, context, session, geneList);
+        }
+
+        return null;
+    }
+
+    private void setSolvedGenes(XWikiDocument doc, XWiki xwiki, DocumentReference patientClassReference,
+        DocumentReference geneClassReference, XWikiContext context, Session session, List<String> geneList)
```
antonkatz: I can't see where ```geneList``` is used

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
```diff
@@ -0,0 +1,290 @@
...
+        throws HibernateException, XWikiException
+    {
+        BaseObject patient = doc.getXObject(patientClassReference);
+        StringProperty oldTarget = (StringProperty) patient.get(SOLVED_NAME);
+        patient.removeField(SOLVED_NAME);
```
antonkatz: @sdumitriu Is it wise to remove a field before creating a new one? Or does it make no difference since the ```save``` call needs to happen before any changes take place?

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
antonkatz: Whoops, nevermind

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
```diff
@@ -0,0 +1,290 @@
...
+            StringProperty oldGeneName = (StringProperty) gene.get(GENE_NAME);
+            LargeStringProperty oldGeneComments = (LargeStringProperty) gene.get(COMMENTS_NAME);
+            if (!geneList.contains(oldGeneName.getValue())) {
+                BaseObject newgene = doc.newXObject(geneClassReference, context);
+                newgene.setStringValue(GENE_NAME, oldGeneName.getValue());
+                newgene.setLargeStringValue(COMMENTS_NAME, oldGeneComments.getValue());
+                newgene.setStringValue(STATUS_NAME, "candidate");
+            } else {
+                // we have a duplicate in either solved or rejected genes - updating its comment
+                String commentUpend =
```
antonkatz: Did you mean ```commentAppend```?

veronikaslc: yup

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
antonkatz: I never understood how this works under the hood. In any other language, I think of ```||``` as a function taking 2 booleans. Clearly not the case with javascript.

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
veronikaslc: name of a property for solved genes

antonkatz: But isn't "GENE_CLASS" and "REJECTED_CLASS" now also have to conform to ```p.id.name = SOLVED_NAME```?

veronikaslc: solved genes don't belong to those classes, they are only in PATIENT_CLASS

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
sdumitriu: `||` is a binary operator, just like `+`, `-`, or `*`. And just like those, it takes two operands and returns one value, and they can be used anywhere where an expression is evaluated. And in JavaScript, an expression doesn't have to be assigned to be valid, you can just write `2+3;` as a statement, it will evaluate to `5` and it doesn't care that that 5 is lost/discarded.

In JavaScript, all operators accept any type of value, and works with those values somehow behaving like the expected type of value. For example, `"5" - "3"` returns `2`, even if the inputs are strings.

So boolean operators evaluate each value as either `true` or `false`, without actually converting those values to an actual boolean; thus values in JavaScript are called "truthy" or "falsey".

Unlike the boolean logic tables that are usually taught, where `false || true = true`, JavaScript evaluates boolean expressions in terms of its inputs, using short circuiting: the output of a boolean operator is one of the two inputs:

- falsey && anything = falsey (left operand)
- truthy && something = something (right operand)
- truthy || anything = truthy (left operand)
- falsey || something = something (right operand)

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
antonkatz: I think I'm just reading the query wrong. I'm reading it (part of it) as "Get a StringProperty which belongs to an object of type GeneClass OR RejectedGeneClass OR PatientClass. Once such a StringProperty is found (from any of the classes), it's name must equal SOLVED_NAME"
That's fine if the property is in PatientClass, but not if it's in any other class.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
meatcar: Javascript implements short-circuit evaulations of logical operators. That
means that if the left side is true in an `||`, the right side is not
evaluated. See
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-Circuit_Evaluation

antonkatz: Where's the upvote button?

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R70190PhenoTips1280DataMigration.java
veronikaslc: In other words, we are collecting all records from INVESTIGATION_CLASS (candidate genes) and REJECTED_CLASS (rejected genes), but from PATIENT_CLASS we need only one property - SOLVED_NAME (solved genes)

components/export/api/pom.xml
veronikaslc: lowering coverage check ratio to pass build

components/patient-data/impl/pom.xml
veronikaslc: rebase side-accident change

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
veronikaslc: separate file for just a one function.. hmm.. seems over-optimizing may be?

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
veronikaslc: prototypes doesnt seem to work, however, the .replace() itself surprizingly can do a templating, will stick to that solution:
var res = str.replace(/blue|house|car/gi, function myFunction(x){return x.toUpperCase();});

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -2397,9 +2397,8 @@ return PhenoTips;
         onSuccess: function (response) {
           this.setContent(this.buildTree(this._getDataFromResponse(response)));
           this.__crtRoot = id;
-          if (this.container.content) {
-            Event.fire(document, 'xwiki:dom:updated', {'elements' : [this.container.content]});
-          }
+          if (this.container.contentContainer != null)
```
antonkatz: What's the reason for this change?

components/widgets/ui/src/main/resources/PhenoTips/GeneVariantMacros.xml
```diff
@@ -0,0 +1,1532 @@
...
+.gene-input-label{
+    display: inline;
+    padding-right: 5px;
+}
+.xHelpButton.gene-info{
+    padding-top: 5px;
+    margin-right: -16px;
+    padding-left: 10px;
+}
+.suggested.suggest-gene.gene-name{
```
antonkatz: very minor comment - there are a few places where there is no space between the ```{``` and the selector

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
veronikaslc: bugfix, found this  bug as a blocker for my stuff

meatcar:
:+1: looks good to merge.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
sdumitriu: But it causes a regression:
- in edit mode, open the "Clinical symptoms" section
- expand one of the categories
- click on the `i` info button next to one of the phenotypes
- click on "Browse related terms..."
- click on the `i` button next to the displayed phenotype or its parents, but not its children
- nothing happens

The standard field for most containers (instances of `PhenoTips.widgets.ModualPopup) is `content`, `contentContainer` is custom defined for one specific type of element, the branches shown when expanding a predefined phenotype.

In other words, this change is wrong.

sdumitriu: Not just this change in particular, but the fact that sometimes `content` is needed, other times `contentContainer` is used.

veronikaslc: will change it back to content then

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -161,17 +183,17 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
         json.put(getJsonPropertyName(), new JSONArray());
         JSONArray container = json.getJSONArray(getJsonPropertyName());
 
+        List<String> keys =
+            Arrays.asList(GENE_KEY, STRATEGY_KEY, COMMENTS_KEY);
```
buske: Missing STATUS_KEY

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -161,17 +183,17 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+
+        List<String> enablingProperties =
+            Arrays.asList(GENES_STATUS_ENABLING_FIELD_NAME, GENES_STRATEGY_ENABLING_FIELD_NAME,
```
buske: Missing GENES_ENABLING_FIELD_NAME

buske:
Closed in favor of #1913.

PR #1874, opened 2015-08-24
Issue #1274 abstract complex controllers

DeanWay:


DeanWay:
I nominate @antonkatz @buske for review

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerCodeFieldsTestImplementation.java
```diff
@@ -0,0 +1,87 @@
...
+    protected List<String> getBooleanFields()
+    {
+        return Collections.emptyList();
+    }
+
+    @Override
+    protected List<String> getCodeFields()
+    {
+        ParameterizedType p = (ParameterizedType) this.getClass().getGenericSuperclass();
+        System.out.println(p.getActualTypeArguments()[0]);
```
buske: Why is this printing to stdout?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.data.internal.controller;
+
+import com.xpn.xwiki.objects.BaseProperty;
```
buske: Are these imports in the codestyle order? Would have expected this `com.xpn.xwiki` import to be with the others below.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+    @Test(expected = UnsupportedOperationException.class)
+    public void saveIsUnsupported() throws ComponentLookupException
+    {
+        this.mocker.getComponentUnderTest().save(this.patient);
+    }
+
+    //-----------------------------------writeJSON() tests-----------------------------------
+
+    @Test
+    public void writeJSONReturnsWhenGetDataReturnsNull() throws ComponentLookupException
```
buske: Perhaps `writeJSONWithNullDataReturnsNull`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+        doReturn(null).when(this.patient).getData(DATA_NAME);
+        JSONObject json = new JSONObject();
+
+        this.mocker.getComponentUnderTest().writeJSON(this.patient, json);
+
+        Assert.assertNull(json.get(DATA_NAME));
+    }
+
+    @Test
+    public void writeJSONWithSelectedFieldsReturnsWhenGetDataReturnsNull() throws ComponentLookupException
```
buske: See above comment about "...Returns"

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+        JSONObject json = new JSONObject();
+        Collection<String> selectedFields = new LinkedList<>();
+
+        this.mocker.getComponentUnderTest().writeJSON(this.patient, json, selectedFields);
+
+        Assert.assertNull(json.get(DATA_NAME));
+    }
+
+    @Test
+    public void writeJSONReturnsWhenDataIsNotKeyValueBased() throws ComponentLookupException
```
buske: See above comment about "...Returns"

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+        doReturn(patientData).when(this.patient).getData(DATA_NAME);
+        JSONObject json = new JSONObject();
+
+        this.mocker.getComponentUnderTest().writeJSON(this.patient, json);
+
+        Assert.assertNull(json.get(DATA_NAME));
+    }
+
+    @Test
+    public void writeJSONWithSelectedFieldsReturnsWhenDataIsNotKeyValueBased() throws ComponentLookupException
```
buske: See above comment about "...Returns"

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,428 @@
...
+
+        Assert.assertNotNull(json.get(DATA_NAME));
+        Assert.assertTrue(json.get(DATA_NAME) instanceof JSONObject);
+        JSONObject container = json.getJSONObject(DATA_NAME);
+        Assert.assertEquals("datum1", container.get(PROPERTY_1));
+        Assert.assertEquals("datum2", container.get(PROPERTY_2));
+    }
+
+    @Test
+    public void writeJSONWithSelectedFieldsAddsContainerWithAllValues() throws ComponentLookupException
```
buske: "...With**All**SelectedFields..."

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/EthnicityControllerTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void checkGetProperties() throws ComponentLookupException
+    {
+        List<String> result =
+            ((AbstractComplexController<List<String>>) this.mocker.getComponentUnderTest()).getProperties();
+
+        Assert.assertEquals(2, result.size());
+        Assert.assertThat(result, Matchers.hasItem("maternal_ethnicity"));
```
buske: Not sure, but couldn't you just do `result.contains("maternal_ethnicity")`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/FamilyHistoryControllerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void checkGetProperties() throws ComponentLookupException
+    {
+        List<String> result =
+            ((AbstractComplexController<Integer>) this.mocker.getComponentUnderTest()).getProperties();
+
+        Assert.assertEquals(3, result.size());
+        Assert.assertThat(result, Matchers.hasItem(CONSANGUINITY));
```
buske: See above re: testing `result.contains`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/GeneListControllerTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+        DocumentReference patientDocument = new DocumentReference("wiki", "patient", "00000001");
+        doReturn(patientDocument).when(this.patient).getDocument();
+        doReturn(this.doc).when(this.documentAccessBridge).getDocument(patientDocument);
+        this.geneXWikiObjects = new LinkedList<>();
+        this.geneXWikiObjects.add(obj1);
+        this.geneXWikiObjects.add(obj2);
+        this.geneXWikiObjects.add(obj3);
+        int i = 1;
+        for(BaseObject gene : this.geneXWikiObjects){
```
buske: Does this pass style checker? I thought there should be a space before `(` and after `)`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
DeanWay: The problem is that `writeJSON()`'s return type is `void`, so it doesn't return null when `getData()` returns null; it simply returns without performing any further actions. I suppose this test could be renamed: `writeJSONChecksForDataBeforeWritingToJSON`?

buske: Good point. Perhaps `writeJSONDoesNotWriteNullData`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/GeneListControllerTest.java
DeanWay: Seems to pass checkstyle but i'll change it anyway

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
DeanWay: :+1:

buske:
Keep in mind that the `RejectedGeneListControllerTest` will be removed and the `GeneListControllerTest` will need to be updated for Veronika's gene/variant refactor.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/EthnicityControllerTest.java
DeanWay: I could, but I believe `assertThat(someList, hasItem(someItem))` is better than `assertTrue(someList.contains(someItem))`. They end up being functionally the same when they pass, but `assertThat()` shows a more informative error message when it fails. This is because `assertTrue(someList.contains(someItem))` will simply evaluate the `contains()` method at runtime and pass a true or false value to the assertion. So if `contains()` returns false, then Junit will just say it was expecting a true value and got false. On the other hand, `assertThat()` will actually pass the contents of the list to Junit.
Examples:
Failure of `assertTrue()`:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
```
Failure of `assertThat()`:
```
java.lang.AssertionError: 
Expected: a collection containing "some_random_ethnicity"
     but: was "maternal_ethnicity", was "paternal_ethnicity"
```

DeanWay:
@buske In that case, should I just delete the RejectedGeneController tests to avoid future merge conflicts?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,432 @@
...
+        list1.add("HP:00000015");
+        doReturn(list1).when(this.baseProperty1).getValue();
+        doReturn(list1).when(this.baseProperty2).getValue();
+
+        PatientData<List<VocabularyProperty>> result =
+            this.codeFieldImplMocker.getComponentUnderTest().load(this.patient);
+
+        List<VocabularyProperty> propertyOneList = result.get(PROPERTY_1);
+        List<VocabularyProperty> propertyTwoList = result.get(PROPERTY_2);
+        Assert.assertNotNull(propertyOneList);
```
antonkatz: Maybe check for something more that it just not being ```null```?

buske:
@DeanWay Unsure, sorry; chat with @veronikaslc and @tmarathe, perhaps?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,432 @@
...
+    @Test(expected = UnsupportedOperationException.class)
+    public void saveIsUnsupported() throws ComponentLookupException
+    {
+        this.mocker.getComponentUnderTest().save(this.patient);
+    }
+
+    //-----------------------------------writeJSON() tests-----------------------------------
+
+    @Test
+    public void writeJSONDoesNotWriteNullData() throws ComponentLookupException
```
antonkatz: Why is this test called "Does not write ```null```"?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
```diff
@@ -0,0 +1,432 @@
...
+        this.mocker.getComponentUnderTest().writeJSON(this.patient, json, selectedFields);
+
+        Assert.assertNull(json.get(DATA_NAME));
+    }
+
+    @Test
+    public void writeJSONAddsContainerWithAllValues() throws ComponentLookupException
+    {
+        Map<String, String> map = new LinkedHashMap<String, String>();
+        map.put(PROPERTY_1, "datum1");
```
antonkatz: ```datum1``` and ```datum2``` should be in local variables

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
antonkatz: Nevermind, I see now

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/GeneListControllerTest.java
```diff
@@ -0,0 +1,313 @@
...
+        PatientData<Map<String, String>> patientData = new IndexedPatientData<>(CONTROLLER_NAME, internalList);
+        doReturn(patientData).when(this.patient).getData(CONTROLLER_NAME);
+        JSONObject json = new JSONObject();
+        Collection<String> selectedFields = new LinkedList<>();
+        selectedFields.add("some_string");
+        selectedFields.add(GENES_COMMENTS_ENABLING_FIELD_NAME);
+
+        this.mocker.getComponentUnderTest().writeJSON(this.patient, json, selectedFields);
+
+        Assert.assertNull(json.get(CONTROLLER_NAME));
```
antonkatz: I'm not sure how this tests for "Fields contains gene enabler"

DeanWay: Basically the test case is for when selectedFields does not contain `GeneListController.GENES_ENABLING_FIELD_NAME`. The behaviour being test for in that case is that the JSON is  not affected. If the current name makes that unclear, perhaps it could be renamed to: `writeJSONWhenSelectedFieldsDoesNotContainGeneEnabler`. Neither describes the actual behaviour particularly well but at least this describes the test case a little better?

antonkatz: I see.
I guess just put your reply as a comment in the test itself, so it becomes more clear.

DeanWay: Added the comment

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/AbstractComplexControllerTest.java
DeanWay: Added a few more tests here

allasm:
Do we want to review again and merge this?

Issue #1880, opened 2015-08-26
IE9 reports some JS errors when loading the patient edit page


allasm:
In version 1.2M5

allasm:
1.2M5 has those errors. 1.2-Snapshot (~1.2-RC1) does not.

This seems to be related to "xwiki:dom:updated" event, for which in 1.2M5 in IE9 "event.memo.elements" contains an array with just one element which is "undefined".

I fixed our code which handles "xwiki:dom:updated" events in a few different places, undefined will now be simply ignored. But since the root cause has likely been fixed elsewhere in 1.2-RC1, I di dnot fix or investigate similar errors produced in the xwiki.js code.

sdumitriu:
Fixed by f2c7e604260fa021bf0a0e455479bc9ed9bf2c3c.

PR #1881, opened 2015-08-26
Issue #1880: Errors on patient edit page in IE9.

allasm:
Fixed (PhenoTips code).


sdumitriu:
This wouldn't fix the source of the error, it would just mask it. Running those callbacks is important, the real question is there an undefined object, and the answer (and the real fix) is in f2c7e604260fa021bf0a0e455479bc9ed9bf2c3c.

Issue #825, opened 2014-03-31
The patient symptoms index should preserve the ontology hierarchy


sdumitriu:
Currently the patient index only stores the actual symptoms, which makes it impossible to match patients with closely related, but not exactly the same phenotypes. Instead of storing a simple list of terms, the full hierarchy from the HPO should be stored somehow, similar to how [path hierarchies are stored](http://lucene.apache.org/core/4_6_0/analyzers-common/org/apache/lucene/analysis/path/PathHierarchyTokenizerFactory.html).

buske:
Replaced by https://phenotips.atlassian.net/browse/PT-825

PR #1886, opened 2015-09-03
Issue #825: The patient symptoms index should preserve the ontology hierarchy.

veronikaslc:
Fixed by adding 'phenotype_ancestors' to the patient indexing.

### DIFF
--- a/components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
+++ b/components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
@@ -23,6 +23,8 @@ import org.phenotips.data.PatientRepository;
 import org.phenotips.data.indexing.PatientIndexer;
 import org.phenotips.data.permissions.PermissionsManager;
 import org.phenotips.vocabulary.SolrCoreContainerHandler;
+import org.phenotips.vocabulary.Vocabulary;
+import org.phenotips.vocabulary.VocabularyTerm;
 
 import org.xwiki.component.annotation.Component;
 import org.xwiki.component.phase.Initializable;
@@ -32,9 +34,12 @@ import org.xwiki.query.QueryException;
 import org.xwiki.query.QueryManager;
 
 import java.io.IOException;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.apache.solr.client.solrj.SolrClient;
@@ -75,6 +80,10 @@ public class SolrPatientIndexer implements PatientIndexer, Initializable
     @Inject
     private PermissionsManager permissions;
 
+    @Inject
+    @Named("hpo")
+    private Vocabulary ontologyService;
+
     @Override
     public void initialize() throws InitializationException
     {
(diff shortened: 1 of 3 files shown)


veronikaslc:
@antonkatz , please review

antonkatz:
+1

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
```diff
@@ -94,9 +104,19 @@ public void index(Patient patient)
 
         for (Feature phenotype : patient.getFeatures()) {
             input.addField((phenotype.isPresent() ? "" : "negative_") + phenotype.getType(), phenotype.getId());
+            VocabularyTerm properTerm = this.ontologyService.getTerm(phenotype.getId());
+            if (properTerm != null) {
+                Set<VocabularyTerm> parents = properTerm.getAncestorsAndSelf();
+                for (VocabularyTerm term : parents) {
+                    allAncestors.add(term.getId());
+                }
+            } else {
```
buske: Empty else branch? I'd suggest either logging something on debug or remove the branch.

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
```diff
@@ -94,9 +104,19 @@ public void index(Patient patient)
 
         for (Feature phenotype : patient.getFeatures()) {
             input.addField((phenotype.isPresent() ? "" : "negative_") + phenotype.getType(), phenotype.getId());
+            VocabularyTerm properTerm = this.ontologyService.getTerm(phenotype.getId());
+            if (properTerm != null) {
+                Set<VocabularyTerm> parents = properTerm.getAncestorsAndSelf();
+                for (VocabularyTerm term : parents) {
```
buske: This not quite correct. Terms should only be added to allAncestors if they are present. We don't want the full ancestry of the negative terms dominating the seeding. It might actually be best to put negative terms under a new solr field, `negative_phenotype_ancestors`.

resources/solr-configuration/src/main/resources/patients/conf/schema.xml
```diff
@@ -61,6 +61,7 @@
     <field name="reporter" type="string" indexed="false" stored="true" required="true" />
     <field name="visibility" type="string" indexed="true" stored="true" required="false" />
     <field name="accessLevel" type="int" indexed="true" stored="true" required="false" />
+    <field name="phenotype_ancestors" type="string" indexed="true" stored="true" multiValued="true"/>
```
buske: Turns out this is basically redundant with the `extended_phenotype` and `extended_prenatal_phenotype` fields.

veronikaslc: Similarly we have extended_negative_phenotype and extended_negative_prenatal_phenotype. In that perspective do we need a new solr field, negative_phenotype_ancestors?

veronikaslc: Also, we do not index any of extended fields : extended_phenotype or extended_prenatal_phenotype or extended_negative_phenotype. So which of them is would make sence to index to run solr query in patient network? Additionally, I discovered that by some reason patient-extended-phenotype-updater is not updating negative_prenatal_phenotype to its extended version with others, but we do have this field in the PatientClass. Does it mean we shouldn't consider it?

buske:
Closing this in favor of #1912.

PR #1888, opened 2015-09-11
PT-2049: Remove yellow on-hover row highlight of a table. 

veronikaslc:
Fixed


sdumitriu:
Don't leave dead code in the sources, remove it completely.

Issue #1624, opened 2015-04-27
Pedigree should have a setting for number of pages when printing


antonkatz:

allasm:
Isn't it more a question of zoom levels not number of pages? E.g "fit on one page" or "zoom in an din" until it eventually occupies 1x2 or 1x3 or 3x1 or 2x3 or whatever pages, depending on the shape?

antonkatz:
Ask Mike as to what exactly they wanted

marta-:
@allasm the goal is to have an easy way to print large pedigrees while keeping them readable. Basically there should be a "print preview" where they can zoom until reaching the desired size, and the preview should be clearly split into pages. Please note that it's much more likely for the pedigree to be too wide than too tall.

PR #1889, opened 2015-09-18
Issue #1624: Added a pedigree print dialog/ability to print pedigrees.

allasm:
Also fixes visible editor-mode only symbols (e.g. node shading) visible in saved pedigree images.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -439,7 +439,7 @@
   display: block;
 }
 .import-block-bottom {
-  margin: 1.25em 0 0.5em 0;
+  margin: 1.2em 0 0.5em 0;
 }
 .import-section {
   font-weight: bold;
@@ -826,14 +826,23 @@
   line-height: 15pt;
 }
 .print-preview-footer {
-  margin-top: 3px;
-  margin-bottom: 10px;
+  margin-top: 1px;
+  margin-bottom: 13px;
   margin-left: 10px;
   margin-right: 25px;
+  display: block !important;
+  color: #733;
+}
+.print-preview-footer-text {
+  color: black;
+  margin-left: 7px;
 }
 .print-settings-block {
   margin: 0.25em 0 0.5em 0;
   display: block;
+}
+.print-move-span {
+  margin-left: 70px;
 }
 .print-zoom-value {
   display: inline-block;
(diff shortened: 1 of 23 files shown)


sdumitriu:
Pedigree menu looks completely broken:
1624-broken-menu

allasm:
It is not for me - I saw it happened once before, and was fixed after a refresh, seems to be caching issue. Is it still broken after a refresh?

sdumitriu:
Yes, still broken.

sdumitriu:
Menu is now fixed. Other issues:

- [ ] Sometimes when there's an arched connection between two nodes on the top row, it doesn't get printed (see attached)
- [ ] The landscape mode isn't supported on FF 38.2.1 ESR / Linux, so the pages have too much extra white space at the bottom
- [ ] "slightly overlapped" is probably too slightly, there's almost no difference with or without that option
- [ ] There should be a smarter way to split the image, right now it just cuts nodes in half, along with any text it may have. But this is a difficult problem, it can wait for a future version.

[ped.pdf](https://github.com/phenotips/phenotips/files/2855/ped.pdf)

allasm:
1) Will check

2) There is no way to force FF to use landscape mode programmatically ([known missing feature](https://bugzilla.mozilla.org/show_bug.cgi?id=851441)). I should probably add an option to print in portrait, but generally users are supposed to manually set landscape mode in FF's print dialogue before printing

3) overlap is about 0.5 centimeters when printed - the idea that should be enough to overlap/glue pages together. Need to ask Courtney again if this is needed.

4) I though about this, this is hard and in general case there may be no perfect split. My idea was that by enlarging/shrinking by 10% you can get a better fit manually. Not sure if manual left-right shift controls are useful. Overlaps may help as well to mitigate this problem.

allasm:
#1 is fixed.
#2 unfixable, made warning about manually selecting correct orientation more prominent
#3 increased overlap slightly
#4 added manual controls so that users can move the pedigree slightly for best split between the pages

sdumitriu:
Why was this closed?

sdumitriu:
The top is correctly printed, but now the bottom is cropped:
cropped-bottom

allasm:
Fixed. I will improve page splitting more significantly for 1.3, but this should work for now.

allasm:
Rebased on top of current master, updated for require.js and did a smoke test - ready to be merged

Issue #1859, opened 2015-08-10
Reword parts of cancer section in pedigree editor


allasm:
- default option: display <blank> instead of "not tested"
- "age of diagnosis" -> "as of"
- new button "none of the above", auto-record "as of" to be current age, if age is available

PR #1890, opened 2015-09-18
Issue #1859: Updated cancer section in the pedigree editor

allasm:
(some rewordings, added "none of the above" button).

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -403,7 +403,7 @@
 }
 .menu-box .field-box textarea {
   width: 99%;
-  resize: none;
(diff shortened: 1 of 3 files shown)


components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -626,6 +626,27 @@
   padding: 0.5em 0 0.5em;
 }
 
+.menu-box .patient-menu-button {
+  -webkit-border-radius: 2;
```
sdumitriu: `border-radius` is [well supported](http://caniuse.com/#feat=border-radius), so thete's no need for the prefixed proprties. Same for [`box-shadow`](http://caniuse.com/#feat=css-boxshadow).

PR #1891, opened 2015-09-18
PT-1161: Add a REST-full API for accessing vocabularies

mjshepherd:


mjshepherd:
I nominate @itaiGershtansky and @veronikaslc for review

veronikaslc:
The  patient-vocabularies-rest-1.2-SNAPSHOT.jar does not appear in the build among other jars, so has to be copied manually to test stuff, is it the intention?

veronikaslc:
The API returns xml, isn't JSON format easier to parse/work with and is a kind of universal REST data return format?

mjshepherd:
@veronikaslc Thanks for pointing those out. The first was not intentional and has been corrected. In regards to your second comment, a few of the resources can return either JSON or XML. You must specify JSON in the request to get a JSON response. (see https://phenotips.org/DevGuide/RESTfulAPI)

allasm:
> You must specify JSON in the request to get a JSON response

But if none is specified (via Accept header?) should'n JSON be the default?

mjshepherd:
I think that you are probably right. JSON is way more useful. But this is something that we will have to talk about. Currently, certain resources in patient-data return xml by default as well.

sdumitriu:
Build fails because `org.phenotips.vocabulary.VocabularyManager.getAvailableVocabularies()` doesn't exist.

sdumitriu:
Which is weird, since it does exist, it's the first change in the patch.

sdumitriu:
Oh, I think the reason is that you need to update the version in the poms to `1.3-SNAPSHOT`

veronikaslc:
same reason now preventing build of phenomecentral.org to be on top of phenotips 1.3-SNAPSHOT if you build locally

mjshepherd:
@veronikaslc @itaiGershtansky Build fails on distribution somewhere, I think its ready for review

sdumitriu:
Already working on merging it, polishing up a few things locally.

PR #1894, opened 2015-09-28
PT-2059: "Enter" should accept first phenotype suggestion

mjshepherd:


sdumitriu:
This is not enough, Enter is disabled globally.

PR #1899, opened 2015-10-08
Issue PT-1619: ExportFilter count does not reflect number of patients…

antonkatz:
… that will be exported.

### DIFF
--- a/components/export/ui/src/main/resources/PhenoTips/ExportFilter.xml
+++ b/components/export/ui/src/main/resources/PhenoTips/ExportFilter.xml
@@ -212,6 +212,38 @@
 #else
   #set ($query = '')
 #end
+
+## Rights management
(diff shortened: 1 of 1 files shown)


components/export/ui/src/main/resources/PhenoTips/ExportFilter.xml
```diff
@@ -212,6 +212,38 @@
 #else
   #set ($query = '')
 #end
+
+## Rights management
```
antonkatz: I don't like this approach, but to me it seems that using the internal security components which depend on ```Patient``` will be too slow.

antonkatz: This code is copied from the ```AllData``` page

PR #1900, opened 2015-10-13
PT-2094: Modify button label on a Pedigree save dialog from DON'T QUIT to DON'T CLOSE

veronikaslc:
Fixed.


veronikaslc:
@allasm please review

PR #1901, opened 2015-10-14
PT-2046

tmarathe:


components/xwiki-platform-users/xwiki-platform-users-api/pom.xml
```diff
@@ -63,6 +63,11 @@
       <version>${xwiki.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-component-api</artifactId>
+      <version>7.0.1</version>
```
sdumitriu: Don't hardcode versions like this, use `${xwiki.version}`.

This makes it easier to upgrade the XWiki version, since it only requires changing one parameter in one place, instead of searching for all the individual dependencies in all the pom files.

components/patient-data/impl/pom.xml
```diff
@@ -122,6 +122,12 @@
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
     </dependency>
+    <dependency>
```
sdumitriu: This isn't needed.

components/obo2solr/pom.xml
```diff
@@ -46,6 +46,11 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
```
sdumitriu: This isn't needed.

tmarathe: Yup, I just noticed unnecessary changes to pom files, removing them now.

PR #1904, opened 2015-10-15
PT-2131: BOADICEA export incorrectly lists age of dead family members

allasm:
Fixed

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
@@ -169,16 +169,25 @@ define([], function(){
             return jsDate;
         },
 
-        // Returns either a decade or the year (both as string)
+        // Returns either a decade or the year (as string, which may include non-numeric characters, e.g. "1920s")
         getBestPrecisionStringYear: function() {
             if (!this.isComplete()) return "";
             if (this.year == null) return this.decade;
             return this.year.toString();
         },
 
+        // If year is given returns the year; for decades returns the first year of the decade
+        // (as string representation of an integer)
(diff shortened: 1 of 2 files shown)


components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
```diff
@@ -200,11 +209,15 @@ define([], function(){
...
-        // Iff "failsafe" returns first year of the decade if year is not set and decade is
-        getYear: function(failsafe) {
+        // Iff "failsafe" returns a value even if only decade is availabe (and year is not):
+        //  - iff "average" the middle year of the decade is returned, otherwise the first year of the decade.
+        getYear: function(failsafe, average) {
             if (this.isComplete() && this.year == null && failsafe) {
                 // remove trailing "s" from the decade && convert to integer
                 var year = parseInt( this.decade.slice(0,-1) );
+                if (average) {
+                    year += 5;
```
sdumitriu: What happens if the date is even fuzzier, like "1700s"? Will this block never be reached, since `isComplete` returns false?

allasm: Good point, I forgot we have those extremely fuzzy dates.

It will fail (work incorrectly) by returning 1705. Not sure 1800-s and 1700-s matter for BOADICEA, but to fix this better we probably need different suffixes for decades and millenniums. Though after available date range has been extended down to 1850 this problem becomes purely academic IMHO.

Do you want me to fix this?

sdumitriu: Not really.

PR #1907, opened 2015-10-19
PT-1951: Patient measurements - UI improvements

danielpgross:
nominating @marta- and @meatcar for review


marta-:
@danielpgross I'm looking at pull requests anyway. Please nominate another reviewer.

danielpgross:
Ok. Nominating @veronikaslc for review!

danielpgross:
The build appears to be failing on ci server, but I'm able to build locally without any problems. Reviewers, please let me know if you're having trouble building or testing.

mjshepherd:
Hey @danielpgross, the build is failing for me. The rest api's dependencies are incomplete. You should add the following:
```<dependency>
      <groupId>org.phenotips</groupId>
      <artifactId>patient-measurements-api</artifactId>
    </dependency>
    <dependency>
      <groupId>org.phenotips</groupId>
      <artifactId>vocabularies-api</artifactId>
    </dependency>```

**Comment by mjshepherd** [comment]
Also I know I'm not a reviewer but would you consider moving the static methods in `AbstractMeasurementHandler` to a utility class?

**Comment by danielpgross** [comment]
Thanks @mjshepherd! Yes, I did consider creating a separate utility class for the static methods, but I though it would keep things simpler to have everything in that class. What rationale did you have in mind for moving them into a separate class?

**Comment by mjshepherd** [comment]
I find it odd because we have a `MeasurementHandler` interface and abstract class. I tend to view abstract classes as providing shared code for classes implementing an interface. I don't think of calling or using the abstract class itself. It would be better to define the static methods in the the interface but thats a java 8 feature.

**Comment by sdumitriu** [comment]
@mjshepherd is right.

**Comment by danielpgross** [comment]
@mjshepherd @sdumitriu my thinking was based on the [discussions listed here](http://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html#utility-classes-are-evil) about why utility classes should be avoided because they can be an imposition of a procedural paradigm within an object-oriented language. This case is tricky since the methods don't belong well in the abstract class either. Maybe the ideal solution would be to make them instance methods called on the subclassed handlers instead of being static.

I do see however that we have [a few precedents](https://github.com/phenotips/phenotips/search?utf8=%E2%9C%93&q=utility) in the project for using utility classes, and if @sdumitriu prefers this style then that's fine with me. @sdumitriu, I am also curious to hear your rationale in this case and what you have to say about the discussion that I've linked to above.

**Comment by veronikaslc** [comment]
I can't spot anything for now, so I vote +1.

File: components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationClass.xml [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+
+<xwikidoc version="1.1">
+  <web>PhenoTips</web>
+  <name>MeasurementsConfigurationClass</name>
+  <language/>
+  <defaultLanguage/>
+  <translation>0</translation>
+  <creator>xwiki:XWiki.Admin</creator>
+  <creationDate>1438355127000</creationDate>
+  <parent>XWiki.XWikiClasses</parent>
```
antonkatz: I don't think this is the right parent

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
```diff
@@ -0,0 +1,386 @@
...
+#set ($prefix = "${configObj.xWikiClass.name}_${configObj.number}_")
+#set ($formAction = $configDoc.getURL('save'))
+#set ($formSubmit = "action_save")
+#set ($formParams = {
+  "xredirect" : "$doc.getURL($xcontext.action, $!request.queryString)"
+})
+##
+#macro(measurementTool $name $icon $hint $url $attrs)&lt;a href="$url" title="$!{hint}" class="extension-tool ${name}" $!{attrs}&gt;&lt;i class="fa fa-${icon}"&gt;&lt;/i&gt;&lt;/a&gt;#end
+##
+#macro(measurementTools $meas)
```
antonkatz: ```$meas``` does not seem to be used

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationClass.xml
sdumitriu: Why not, that's the standard parent for classes in XWiki.

We do tend to use `PhenoTips.PatientClass` for classes related to the patient sheet, but in this case `MeasurementsConfigurationClass` isn't related to the patient sheet.

antonkatz: It's part of PhenoTips, shouldn't it have a PhenoTips parent?

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
```diff
@@ -0,0 +1,386 @@
...
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>var PhenoTips = (function (PhenoTips) {
+  var widgets = PhenoTips.widgets = PhenoTips.widgets || {};
+  widgets.MeasurementsConfigurationEditor = Class.create({
+
+    _ENHANCED_TOOL_CSS_CLASS : 'initialized',
```
antonkatz: I *feel* like the name is misleading

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
```diff
@@ -0,0 +1,386 @@
...
+    _silenceMoveButtons : function () {
+      this._element.select('a.move' + this._NEW_TOOL_SELECTOR).invoke('observe', 'click', function(event) {
+        event.stop();
+      });
+      this._element.select('a.move' + this._NEW_TOOL_SELECTOR).invoke('addClassName', this._ENHANCED_TOOL_CSS_CLASS);
+    },
+
+    _syncDefaultValues: function() {
+      this._element.select('.extension-maquette').each((function(el) {
+        el.down('input[name$=default]').disabled = !!!el.up('#default');
```
antonkatz: ```!!!```? I've never seen that before

allasm: !!! is just a triple negation, and is equivalent to just one !. The only reason to use !!! is to make it stand out as far as I know - http://stackoverflow.com/questions/21154510/the-use-of-the-triple-exclamation-mark

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -41,8 +41,299 @@
...
+#charts-new {
+  padding: 15px; padding-left: 190px;
+  height: 100%;
+  box-sizing: border-box;
+  float: right; width: 48%;
+  position: relative;
+}
+#charts-new .sets {
+  width: 120px;
+  position: absolute; top: 0; left: 0;
```
meatcar: Each property should probably be on a newline.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -41,8 +41,299 @@
...
+  float: right;
+  margin-left: 15px;
+}
+.measurement-row {
+  clear: both;
+}
+.measurement-row &gt; label,
+.measurement-row &gt; span,
+.measurement-row &gt; div,
+.measurement-row span.val
```
meatcar: newline here is unnescessary.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -41,8 +41,299 @@
...
+  display: none;
+}
+.measurement-set {
+  border: 1px dotted #aaa;
+  padding: 10px;
+  float: left;
+  clear: both;
+  overflow: auto;
+  margin: 5px;
+  position: relative;
```
meatcar: Don't mean to be picking on you, since this is probably something we do everywhere, but the ordering of the CSS properties is kind of important. We should consider following something like this: https://github.com/necolas/idiomatic-css#declaration-order

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -108,51 +399,268 @@
...
+    #end
+    &lt;span class="unit"&gt;#if ("$!unit" != "")$services.localization.render("phenotips.UIXField.measurements.units.${unit}")#end&lt;/span&gt;
+  &lt;/div&gt;
+  &lt;span class="pctl#if ("$!val" != "") $services.measurements.getFuzzyValue($sd)#end"&gt;#if ("$!val" != "")$eval#end&lt;/span&gt;
+  {{/html}}
+)))
+#end##
+##
+## Source: http://extensions.xwiki.org/xwiki/bin/view/Extension/Sort+a+list+of+XWiki+objects
+#macro (sortXWikiObjectsList $inputObjects $propertyName $outputObjects)
```
meatcar: can we pull this out into some generic "util" macro, that's not specific to this particular sheet?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
mjshepherd: I believe @sdumitriu prefers alphabetical order.

meatcar: As of 2012, he is in the minority :P
https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

On Fri, Oct 30, 2015 at 3:36 PM, Matthew Shepherd <notifications@github.com>
wrote:

> In
> components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
> <https://github.com/phenotips/phenotips/pull/1907#discussion_r43543625>:
>
> > +}
> > +.extreme-below-normal,
> > +.extreme-above-normal {
> > +  color: red;
> > +}
> > +.measurement-set.proto {
> > +  display: none;
> > +}
> > +.measurement-set {
> > +  border: 1px dotted #aaa;
> > +  padding: 10px;
> > +  float: left;
> > +  clear: both;
> > +  overflow: auto;
> > +  margin: 5px;
> > +  position: relative;
>
> I believe @sdumitriu <https://github.com/sdumitriu> prefers alphabetical
> order.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phenotips/phenotips/pull/1907/files#r43543625>.
>



-- 
Denys Pavlov

allasm: Alphabetical does not make any sense for CSS, as ordering is only important for human readability, and readability means placing related properties near each other (e.g. margin/position/float/left/right)

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
sdumitriu: But it does help readability.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
sdumitriu: It's hard to come up with a good semantical order, so in the absence of clear rules, one simple rule wins: everyone knows the alphabet.

components/patient-measurements/api/src/main/java/org/phenotips/measurements/internal/AbstractMeasurementHandler.java
```diff
@@ -166,6 +199,51 @@ public void initialize() throws InitializationException
...
+    @Override
+    public Collection<VocabularyTerm> getAssociatedTerms(Double standardDeviation)
+    {
+        ResourceBundle configuration = ResourceBundle.getBundle("measurementsAssociatedTerms");
+        List<VocabularyTerm> terms = new ArrayList<>();
+
+        if (null == standardDeviation || standardDeviation <= -3) {
+            addResolvedTermsToList(configuration, this.getName(), "extremeBelowNormal", terms);
+        }
+        if (null == standardDeviation || standardDeviation <= -2) {
```
meatcar: you don't need repeating null checks. It would have been caught by the first one. If you want, you can even initially do 
```java
if (standardDeviation == null) {
    standardDeviation = Double.NEGATIVE_INFINITY
}
```

meatcar: You can also remove duplicating calls to `addResolvedTermsToList()` by refactoring this to do one call at the end.

meatcar: Why aren't you using the constants in `MeasurementUtils`?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
allasm: well, it is not that important to have any ordering for CSS, as there are 3-4-5 maybe 7-8 properties in one block to start with, and it is not going to be expanded much in future releases (unlike regular code). Within those few, no order is about as good as alphabetical, as none of them really helps. IMHO the only thing which somewhat helps to understand what some CSS does is to group by function. Enforcing alphabetical makes developers waste time on alphabetic arrangement (maybe code review + a commit to fix mix-arranged), for no apparent benefit. For CSS I'm ine with no required order, but if we do require one I would vote by function (up to every developer's understanding of it, any reasonable order will be better than alphabetical, and have no future drawbacks like maybe random order of Java imports may have)

components/patient-measurements/ui/src/main/resources/PhenoTips/MeasurementSetDeletionService.xml
```diff
@@ -0,0 +1,66 @@
...
+    #break
+    #end
+    #set ($objs = $thatDoc.getObjects("PhenoTips.MeasurementClass", "age", $age))
+    #foreach($obj in $objs)
+    #set ($discard = $thatDoc.removeObject($obj))
+    #end
+    #set ($discard = $thatDoc.save("Removed measurement set"))
+    $response.setStatus(200)
+    #end##
+    {{/html}}
```
meatcar: shouldn't the `#if` statements be indented? It's kind of painful to read this.

danielpgross: You're right, this file lost its indentation at some point before I committed.

marta-: I think @danielpgross was trying to avoid the generation of wiki markup. The other option is to use the parameter `wiki="false"` (works for both `{{velocity}}` and `{{html}}` macros. Also `{{html}}` is unnecessary here.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4866,6 +4869,193 @@ document.observe("xwiki:dom:loaded", function() {
...
+      return "(" + this.pieces.s + this.pieces.n + this.pieces.s + this._regexpifyWord(this.pieces[unit] || "") + this.pieces.s + ")" + (makeOptional ? '?' : '');
+    },
+    _regexpifyWord : function(word) {
+      return word.replace(/(.)/g, "$1?").replace("?", "");
+    },
+    match : function(text) {
+      return (text || "").match(this.regexp);
+    },
+    getValue : function(text) {
+      if ((text || "").match(this.regexp)) {
```
meatcar: can't you just use `match` defined above? There are other occurrences of this in the file.

components/patient-measurements/api/src/main/java/org/phenotips/measurements/internal/AbstractMeasurementHandler.java
danielpgross: The repeating null checks are there on purpose -- if null is passed, *all* terms are supposed to be returned. Setting to negative or positive infinity will not accomplish that because the positive SD terms will be missed.

`addResolvedTermsToList()` can't be factored out because, again, there might be multiple terms to add to the list in one method call. This is also why I used `if` instead of `else if`.

I'm not using the constants in `MeasurementUtils` because those ones are `delimiter-separated` as opposed to `camelCase`.

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
danielpgross: Yes, I think this just a typo. Thanks for pointing it out.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
marta-: right... somehow the `match` method was implemented and then forgotten, thanks for the catch.

danielpgross: This is @marta-'s code :-) But yes I can factor that down.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
danielpgross: I agree with Andrei. If we're trying to reduce cognitive load, does it really make sense to ask developers to think about alphabetically ordering their properties...?
I have used declaration order before and I also much prefer that to alphabetical.

meatcar: I think we should optimize for reading, not for writing. If we do move
towards a system where we bundle the respective css, velocity, and js files
outside the xwiki docs, We can just run a formatter over them as part of
the build process.

On Fri, Oct 30, 2015 at 4:20 PM, Daniel Gross <notifications@github.com>
wrote:

> In
> components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
> <https://github.com/phenotips/phenotips/pull/1907#discussion_r43547869>:
>
> > +}
> > +.extreme-below-normal,
> > +.extreme-above-normal {
> > +  color: red;
> > +}
> > +.measurement-set.proto {
> > +  display: none;
> > +}
> > +.measurement-set {
> > +  border: 1px dotted #aaa;
> > +  padding: 10px;
> > +  float: left;
> > +  clear: both;
> > +  overflow: auto;
> > +  margin: 5px;
> > +  position: relative;
>
> I agree with Andrei. If we're trying to reduce cognitive load, does it
> really make sense to ask developers to think about alphabetically ordering
> their properties...?
> I have used declaration order before and I also much prefer that to
> alphabetical.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phenotips/phenotips/pull/1907/files#r43547869>.
>



-- 
Denys Pavlov

buske:
I would strongly suggest pulling any substantial css, js, and velocity code out into separate resources. Here's an example of how we did it with the unified genes and variants: https://github.com/phenotips/phenotips/commit/ce6d1d2ec95f5593955d3e0e021ff1f34531ff6e

components/patient-measurements/api/src/main/java/org/phenotips/measurements/internal/AbstractMeasurementHandler.java
meatcar: After some thinking on this with the team, there are a couple of issues here.

3. I feel like the logic in `MeasurementUtils.getFuzzyValue(double)` can be somehow combined with this.

1. I don't think the way this function is written makes it clear that `addResolvedTermsToList` will be called a certain number of times in certain situations. Initially, the pattern that this was written in caused me to interpret it as a typical "switch-style" function. 

    Something like this might be clearer:
    ```java
Set<String> fuzzys = new HashSet(Arrays.asList(
            "extremeBelowNormal",
            "belowNormal",
            "aboveNormal",
            "extremeAboveNormal"
            ));

if (standardDeviation != null) {
     if (standardDeviation > -3) {
         set.remove("extremeBelowNormal");
     } else if (standarDeviation > -2) {
         set.remove("belowNormal");
     } /* .... */
}

for (String fuzzy : fuzzys) {
     addResolvedTermsToList(/*...*/ fuzzy /*...*/);h ...
}
    ```

2. @matt doesn't like that `addResolvedTermsToList` is very C-like, as in it takes a pointer to a list it's supposed to modify, instead of returning a list that gets appended to the existing list like a proper "Object-Oriented" function.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
antonkatz: Though I have to say that this is one of my biggest pet-peeves about auto-formatters for css. I like related properties to be on the same line. It saves space and improves readability.

components/patient-measurements/api/src/main/java/org/phenotips/measurements/internal/AbstractMeasurementHandler.java
antonkatz: I'm with @mjshepherd on this one. Immutability above all (with a caveat).

danielpgross: Thanks for the feedback, guys. Admittedly, I wrote this part of the code in a bit of a rush and not with too much thought towards style. I will go ahead with a refactor, doing my best to take all of this into account.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
danielpgross: I will go ahead and order the CSS alphabetically, although I'm still not sure about having this policy for the team going forward.

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
```diff
@@ -0,0 +1,386 @@
...
+           }
+         });
+       });
+    },
+  });
+  return PhenoTips;
+}(PhenoTips || {}));
+
+document.observe("xwiki:dom:loaded", function(){
+  window.wiz = new PhenoTips.widgets.MeasurementsConfigurationEditor($('measurements-configuration-editor'));
```
antonkatz: Why ```wiz```?

components/navigation/ui/src/main/resources/data/WebHome.xml
```diff
@@ -29,8 +29,8 @@
   <parent>Main.WebHome</parent>
   <author>xwiki:XWiki.Admin</author>
   <contentAuthor>xwiki:XWiki.Admin</contentAuthor>
-  <date>1401825901000</date>
-  <contentUpdateDate>1401822185000</contentUpdateDate>
+  <date>1444427833000</date>
+  <contentUpdateDate>1444407758000</contentUpdateDate>
```
antonkatz: We usually don't commit date changes

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -108,51 +399,268 @@
...
-  $xwiki.ssx.use('XWiki.DateTimePicker', {'colorTheme' : "$xwiki.getSpacePreference('colorTheme')"})##
-  $xwiki.jsx.use('XWiki.DateTimePicker')##
-  #if ($config &amp;&amp; "$!{config.getProperty('enableImperialUnitSystem').value}" == '1')
-    #set ($measurementUnitInfo = "enable-unit-switch $!{config.getProperty('measurementUnitSystem').value}")
+$xwiki.jsx.use('XWiki.DateTimePicker')##
+$xwiki.ssx.use('PhenoTips.UIX_Field__measurements')##
+#set ($measurementClassId = $doc.getObjectNumbers('PhenoTips.MeasurementClass'))##
+#set ($objs = $services.query.xwql('from doc.object(PhenoTips.MeasurementsConfigurationClass) as obj').setLimit(1).execute())##
+#set ($itemDoc = $xwiki.getDocument($objs.get(0)))##
+#set ($configObj = $itemDoc.getObject('PhenoTips.MeasurementsConfigurationClass'))##
```
antonkatz: I don't think this is very important, as the chances of it happening are tiny, but the user will get an ugly error if the configuration document is not found

allasm: What is not very important?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
```diff
@@ -108,51 +399,268 @@
...
+      #else
+        #set($vals = {})
+      #end
+      #measurement($handler $vals $proto)
+    #end
+  #end
+#end##
+##
+#macro (measurementSet $set)
+#if ("$!set" == "")
```
antonkatz: Sometimes there is a indentation right after the ```macro``` declaration (like above) and sometimes not

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
antonkatz: The user getting an ugly error.

allasm: Hmm, but where is that error? You are commenting on a line that just gets a config object? Do you mean that there should be a check that that object exists?

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
danielpgross: This is left over from the [form designer code](https://github.com/phenotips/phenotips/blob/master/components/vocabularies/hpo/ui/src/main/resources/PhenoTips/FormDesigner.xml#L875) upon which this was based. I assume it was for testing/debug purposes, but I don't see any harm in leaving it in.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__measurements.xml
antonkatz: Ideally there should be a check that the document exists, and that the object exists. But like I said, this is not important.

components/application-configuration/ui/src/main/resources/PhenoTips/MeasurementsConfigurationEditor.xml
antonkatz: I meant that ```wiz``` is an odd name. If I see ```wiz``` somewhere in a different piece of code, I would not think ```MeasurementsConfigurationEditor```

components/patient-measurements/rest/src/main/java/org/phenotips/data/rest/internal/DefaultMeasurementChartResourcesResourceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+@Singleton
+public class DefaultMeasurementChartResourcesResourceImpl extends AbstractMeasurementRestResource implements
+        MeasurementChartResourcesResource
+{
+    /** Provides access to the XWiki data. */
+    @Inject
+    private DocumentAccessBridge bridge;
+
+    /** The set of chart resource objects corresponding to this request. */
+    private List<ChartResource> charts;
```
antonkatz: Hold on... you can't use state like this in a singleton

antonkatz: Why not just have ```generateChartResources``` return ```List<ChartResource>``` and ```getResponse``` consume it?

mjshepherd:
You are probably super sick of hearing all these comments but I was working on another issue and noticed a couple bugs.
1.
screen shot 2015-11-06 at 4 33 24 pm
There is still a growth charts section under measurements, I don't think its necessary any more.

2.
screen shot 2015-11-06 at 4 40 40 pm
measurement dates before birth have some odd behaviour (don't know if this is a big deal or not)

3. 
If the user inputs an invalid string in the age field (say "five") and views the patient summary it will trigger a velocity macro exception. Not terrible but the mistake persists in editing and breaks measurements so you can not go back and fix it once it has been saved.

components/patient-measurements/rest/src/main/java/org/phenotips/data/rest/internal/DefaultMeasurementChartResourcesResourceImpl.java
danielpgross: I had originally used a class property so that the list generated from the current request could be used in different methods, if necessary. Since the list is reset with every request, this usage is safe as far as I can tell, but I do agree that this isn't the best practice given that it's a singleton with state persisting across requests. I'll go with your recommended change.

danielpgross:
Thanks @mjshepherd for reporting these. I've removed the now unused charts section. I've left the other 2 issues as they're validation issues and my understanding is that most of our validation is loose right now.

mjshepherd:
@danielpgross The last issue I reported is more than validation. It was causing a velocity exception to be thrown and more or less broke the patient form. However, I was using the measurements section yesterday and the the issue seems to be gone.

danielpgross:
Merged into https://github.com/phenotips/phenotips/tree/feature-measurements for further development, will be merged from there into master.

PR #1908, opened 2015-10-21
PT-2013: Date-of-death-unknown checkbox in a very weird place

antonkatz:


sdumitriu:
In view mode, the "date of death is unknown" text appears indented without any label. Looks weird.

buske:
Not only that, clicking on the checkbox hides the entire "Date of death" section, which doesn't make sense.

Would it make more sense to have a `Patient is deceased` checkbox that reveals the "Date of death" field? Or should we have a life status selector like in the pedigree with "alive", "deceased" (, "unborn"?)?

antonkatz:
@buske 
1. Agreed. But if we want to use the ```__displayFormatted``` macro ( @sdumitriu do we?), then that's the way it flows.

2. Again agreed. But since that will introduce changes in the controllers, it should be it's own separate issue.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__date_of_death.xml
```diff
@@ -115,20 +115,38 @@
...
   #if ($showDod)
     (% class="controlled" %) (((
     #__displayFormatted('2-col' 'date_of_death' '' 'date' '')
     )))
   #end
+  #if ($showDodUnknown)
+    #if ($xcontext.action == 'edit')
+      ## The style must be applied iff the unknown option is displayed
+      {{html wiki="false" clean="false"}}
+        &lt;style&gt;
```
sdumitriu: Never do something like this.

sdumitriu:
Here's an idea: first column (div) has the label, second column (div) has the two fields, the date and the checkbox.

buske:
Or, to beat a dead horse, should we simply add a field, `vitality status`: `"alive"`, `"deceased"`, where selecting `"deceased"` causes the date of death field to appear? The field should be shared with the pedigree's life status field, which should probably be renamed.

buske:
Looking at the patient form, `Term birth` is an identical example of this. The two fields should use the same macros.

sdumitriu:
There seems to be too much code in this pull request, I'm almost certain there is a much simpler solution.

veronikaslc:
@buske @sdumitriu Was looking at the issue. I kind of agree with Anton. `Term birth` differs from `Date of Death is Unknown` in a way that `Term birth` checkbox and input under checkbox are one field `gestation` and thus are generated from the PatientTemplate as an html. `controller` and `controlled` CSS class names are used to mark elements that will control the hide/show and those that will be hidden, correspondingly. `Date of Death is Unknown` and `date of death` are two different fields, so we use `UIX_Field__date_of_death.xml` for displaying them. To avoid clicking on the checkbox hides the entire "Date of death" section, we need to specify the `date of death` datepicker input as `controlled` and so we can not use `#__displayFormatted` macro that auto generates both label and input as one row. Probably I just do not see a sillier way than explicitly generate label and inputa separately to label them as contoller and controlled, but I made corrections to Anton's code, it should work as expected.

PR #1910, opened 2015-10-23
pt-2108: PatientData should be lazy-loaded

mjshepherd:


mjshepherd:
@meatcar @itaiGershtansky  please review

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -325,7 +333,8 @@ public JSONObject toJSON(Collection<String> onlyFieldNames)
             result.element(JSON_KEY_DISORDERS, diseasesToJSON());
         }
 
-        for (PatientDataController<?> serializer : this.serializers) {
+
+        for (PatientDataController<?> serializer: this.serializers.values()) {
```
buske: As long as nothing depends on the ordering in which the serializer JSONs are written being consistent between calls, this is okay. If there's any possibility that consistent ordering might be a good thing (e.g., if this data is ever diff'd), then I'd suggest replacing the `HashMap` with a `TreeMap`.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -177,8 +176,12 @@ private void loadDisorders(XWikiDocument doc, BaseObject data) throws XWikiExcep
     private void loadSerializers()
     {
         try {
-            this.serializers =
-                ComponentManagerRegistry.getContextComponentManager().getInstanceList(PatientDataController.class);
+            List<PatientDataController<?>> availableSerializers = ComponentManagerRegistry
+                    .getContextComponentManager()
+                    .getInstanceList(PatientDataController.class);
+            for (PatientDataController<?> serializer : availableSerializers) {
+                this.serializers.put(serializer.getName(), serializer);
```
buske: I'd suggest logging a warning if two serializers have the same name instead of silently overwriting the first one.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -139,7 +139,6 @@ public PhenoTipsPatient(XWikiDocument doc)
         this.disorders = Collections.unmodifiableSet(this.disorders);
 
         loadSerializers();
```
buske: Can we take this one step further and only load the serializers on the first call to `getData`? :)

buske:
Went ahead and reviewed it since I've been working with this a lot lately. I hope that's okay.

sdumitriu:
> I hope that's okay

More than OK. Ideally everybody should review all commits, to check for bugs, to keep up to date with what's happening, and to learn how some things are done.

itaiGershtansky:
+1

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -187,9 +190,10 @@ private void loadSerializers()
     /**
      * Loops through all the available serializers and passes each a document reference.
```
allasm: comment is out of date

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
allasm: +1

buske:
@sdumitriu, ready for you.

meatcar:
:+1:

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
sdumitriu: A `TreeMap` requires that the values are comparable, which isn't the case for `PatientData`.

sdumitriu: No, my bad, it requires that the keys are comparable.

PR #1913, opened 2015-10-26
PT-1280: New gene and variant section

buske:
This PR updates the genotype section of the patient form by:
- Merging candidate, rejected, and solved genes within the same table
- Adding support for manually-entered candidate variants (PT-2023)
- Storing evidence and other information for candidate genes (PT-1381, PT-1149)

I made a few adjustments on top of @veronikaslc's base commit:
- a few bug fixes (mostly found after merging with remote-matching and patient-network functionality)
- removed formatting/style changes if spotted

Replaces #1873.

PR #1897, opened 2015-09-30
PN-90: Genotypic seeding results need to be combined with phenotypic seeding.

veronikaslc:
Added genes to indexing of the Patient.

PR #1914, opened 2015-10-26
PT-2188: Added genes to patient index

buske:
- Based on @veronikaslc's commit 480d1a90
- Refactored gene indexing to use constants and avoid switch statement
- Index genes with blank status as candidates
- Should work even before merging in unified gene section (#1913)

Replaces #1897.

### DIFF
--- a/components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
+++ b/components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
@@ -60,24 +60,6 @@ import org.slf4j.Logger;
 @Singleton
 public class SolrPatientIndexer implements PatientIndexer, Initializable
 {
-    private static final String GENES_KEY = "genes";
-
-    private static final String GENE_NAME_FIELD = "gene";
-
-    private static final String GENE_STATUS_FIELD = "status";
-
-    private static final String GENE_STATUS_SOLVED = "solved";
-
-    private static final String GENE_STATUS_CANDIDATE = "candidate";
-
-    private static final String GENE_STATUS_REJECTED = "rejected";
-
-    private static final String SOLR_FIELD_SOLVED_GENES = "solved_genes";
-
-    private static final String SOLR_FIELD_CANDIDATE_GENES = "candidate_genes";
-
-    private static final String SOLR_FIELD_REJECTED_GENES = "rejected_genes";
-
     /** Logging helper object. */
     @Inject
     private Logger logger;
@@ -143,13 +125,7 @@ public class SolrPatientIndexer implements PatientIndexer, Initializable
         input.setField("visibility", this.permissions.getPatientAccess(patient).getVisibility().getName());
         input.setField("accessLevel", this.permissions.getPatientAccess(patient).getVisibility().getPermissiveness());
 
-        // Add genes
-        PatientData<Map<String, String>> allGenes = patient.getData(GENES_KEY);
-        addGenesWithStatusToField(input, allGenes, GENE_STATUS_SOLVED, SOLR_FIELD_SOLVED_GENES);
-        addGenesWithStatusToField(input, allGenes, GENE_STATUS_CANDIDATE, SOLR_FIELD_CANDIDATE_GENES);
-        addGenesWithStatusToField(input, allGenes, GENE_STATUS_REJECTED, SOLR_FIELD_REJECTED_GENES);
-        // Index genes without a status as candidates
-        addGenesWithStatusToField(input, allGenes, null, SOLR_FIELD_CANDIDATE_GENES);
+        addGenes(input, patient);
 
         try {
             this.server.add(input);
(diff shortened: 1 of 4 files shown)


buske:
@veronikaslc and @allasm, please review.

buske:
Closing and re-opening to restart Jenkins build. #kludge

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
```diff
@@ -95,8 +116,18 @@ public void index(Patient patient)
...
         input.setField("visibility", this.permissions.getPatientAccess(patient).getVisibility().getName());
         input.setField("accessLevel", this.permissions.getPatientAccess(patient).getVisibility().getPermissiveness());
+
+        // Add genes
+        PatientData<Map<String, String>> allGenes = patient.getData(GENES_KEY);
+        addGenesWithStatusToField(input, allGenes, GENE_STATUS_SOLVED, SOLR_FIELD_SOLVED_GENES);
+        addGenesWithStatusToField(input, allGenes, GENE_STATUS_CANDIDATE, SOLR_FIELD_CANDIDATE_GENES);
+        addGenesWithStatusToField(input, allGenes, GENE_STATUS_REJECTED, SOLR_FIELD_REJECTED_GENES);
+        // Index genes without a status as candidates
+        addGenesWithStatusToField(input, allGenes, null, SOLR_FIELD_CANDIDATE_GENES);
```
allasm: As discussed with Orion, it may be an empty string not a null (probably if UI is not implemented well or a check  is missed on save?) But in any case from the perspective of this component it is better to have a false positive than to miss a match

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
```diff
@@ -138,4 +169,18 @@ public void reindex()
             this.logger.warn("Failed to search patients for reindexing: {}", ex.getMessage());
         }
     }
+
+    private void addGenesWithStatusToField(SolrInputDocument input, PatientData<Map<String, String>> allGenes,
+        String status, String field)
+    {
+        if (allGenes != null && allGenes.isIndexed()) {
```
allasm: isIndexed() is probably unnecessary as we already require PatientData to be of the Map<String,String> kind (Sergiu should kno better if this is a good idea to keep it?)

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
veronikaslc: so should I add 'addGenesWithStatusToField(input, allGenes, null, SOLR_FIELD_CANDIDATE_GENES);' to it?

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
veronikaslc: @sdumitriu , when should we check isIndexed() value?

buske: Looking at this in more detail, my understanding was only partial and I was wrong to say it wa unnecessary. That test ensures that the data is indeed a list of `Map` (rather than a single `Map` or a `Map` of `Map`s) before iterating over it.

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
```diff
@@ -138,4 +165,31 @@ public void reindex()
...
+                if (StringUtils.isBlank(geneStatus) || GENE_STATUS_CANDIDATE.equals(geneStatus)) {
+                    solrField = SOLR_FIELD_CANDIDATE_GENES;
+                } else if (GENE_STATUS_SOLVED.equals(geneStatus)) {
+                    solrField = SOLR_FIELD_SOLVED_GENES;
+                } else if (GENE_STATUS_REJECTED.equals(geneStatus)) {
+                    solrField = SOLR_FIELD_REJECTED_GENES;
+                }
+
+                if (solrField != null) {
+                    input.addField(solrField, name);
```
veronikaslc: might be overthinking, but as alternative `geneStatus = StringUtils.isBlank(geneStatus) ? GENE_STATUS_CANDIDATE : geneStatus` and `input.addField(geneStatus+"_genes", name)` instead of `if-else` chain and many static vars. Came from that `solrField` seems to be never `null`

veronikaslc: other than that, +1

components/patient-data/indexing/src/main/java/org/phenotips/data/indexing/internal/SolrPatientIndexer.java
antonkatz: Patient data comes in 3 flavours (no ice cream flavour :( ). The underlying value can be stored as a single variable value, as a map or as a list. The checks ```isIndexed``` and ```isDictionary``` are there to determine which flavour it is.

buske:
@veronikaslc How about this? I removed the static variables since they were a bit extraneous, but kept the if/else. My concern with implicit assignments like `+ "_genes"` is that unexpected inputs will result in unexpected outputs. This imposes implicit constraints between this file and the solr schema that can be especially difficult to debug and track down later, as was the case with the `actual_*`, `extended_*` and `*henotype` indexing.

veronikaslc:
looks good, +1.

buske:
need another reviewer. @meatcar?

meatcar:
:+1: LGTM

PR #1916, opened 2015-10-29
PT-2195: Some JSON is exported even if nothing is marked for export 

antonkatz:
@allasm @itaiGershtansky  please review

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java
@@ -49,6 +49,8 @@ import net.sf.json.JSONObject;
 @Singleton
 public class ClinicalStatusController implements PatientDataController<String>
 {
+    private static final String UNAFFECTED = "unaffected";
+
     /** Logging helper object. */
     @Inject
     private Logger logger;
@@ -66,7 +68,6 @@ public class ClinicalStatusController implements PatientDataController<String>
     @Override
     public PatientData<String> load(Patient patient)
     {
-        String unaffected = "unaffected";
         String affected = "affected";
         try {
             XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
@@ -74,11 +75,11 @@ public class ClinicalStatusController implements PatientDataController<String>
             if (data == null) {
                 return null;
             }
-            int isNormal = data.getIntValue(unaffected);
+            int isNormal = data.getIntValue(UNAFFECTED);
             if (isNormal == 0) {
                 return new SimpleValuePatientData<String>(getName(), affected);
             } else if (isNormal == 1) {
-                return new SimpleValuePatientData<String>(getName(), unaffected);
+                return new SimpleValuePatientData<String>(getName(), UNAFFECTED);
             }
         } catch (Exception e) {
             this.logger.error("Could not find requested document or some unforeseen"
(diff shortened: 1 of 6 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java
```diff
@@ -90,6 +91,10 @@ public String getName()
     @Override
     public void writeJSON(Patient patient, JSONObject json, Collection<String> selectedFieldNames)
     {
+        if (selectedFieldNames != null && !selectedFieldNames.contains(UNAFFECTED)) {
```
antonkatz: Ideally, this should check if ```selectedFieldNames``` contains ```getName()```, but the export UI doesn't work that way.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java
```diff
@@ -66,7 +68,6 @@ public String getName()
     @Override
     public PatientData<String> load(Patient patient)
     {
-        String unaffected = "unaffected";
         String affected = "affected";
```
buske: I would move affected to a static variable as well for consistency.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java
buske: Isn't the field called `clinicalStatus`?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
```diff
@@ -168,8 +169,7 @@ public void writeJSONWithSelectedFieldsAddsAffectedValueToJSON() throws Componen
             new SimpleValuePatientData<>(DATA_NAME, AFFECTED);
         doReturn(data).when(this.patient).getData(DATA_NAME);
         JSONObject json = new JSONObject();
-        Collection<String> selectedFields = new LinkedList<>();
-        selectedFields.add(DATA_NAME);
+        Collection<String> selectedFields = Arrays.asList(UNAFFECTED);
```
buske: Why is this UNAFFECTED and not DATA_NAME?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ClinicalStatusControllerTest.java
```diff
@@ -196,8 +210,7 @@ public void writeJSONWithSelectedFieldsAddsUnaffectedValueToJSON() throws Compon
         PatientData<String> data = new SimpleValuePatientData<>(DATA_NAME, UNAFFECTED);
         doReturn(data).when(this.patient).getData(DATA_NAME);
         JSONObject json = new JSONObject();
-        Collection<String> selectedFields = new LinkedList<>();
-        selectedFields.add(DATA_NAME);
+        Collection<String> selectedFields = Arrays.asList(UNAFFECTED);
```
buske: Again, why UNAFFECTED?

buske:
I think this confounds the distinction between `unaffected` the status and `unaffected` the field name. I'd suggest having separate variables for each (even if the have the same value) to make this distinction very clear. Both here, and in the tests.

allasm:
Looks good other than +1 to @buske comment above

antonkatz:
@buske's comment was addressed in the code. Not sure why the last commit is showing up before his comment.

allasm:
I meant "I would move "affected" to a static variable as well for consistency", it doesn't seem to have been done?

antonkatz:
Done

allasm:
I think this pull request has been merged and can be deleted?

PR #1917, opened 2015-11-04
PT-2035: ParentalAgeController unit tests

tmarathe:


components/patient-data/impl/pom.xml
```diff
@@ -121,6 +121,18 @@
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
     </dependency>
+    <dependency>
```
sdumitriu: This isn't needed, the `patient-data-api` module is already declared as a dependency above.

allasm:
@sdumitriu, the request was mostly fine but it did not have a "+1" yet at the moment you merged it, I think if there is a reviewer we should always wait for  a+1, unless it is very crucial and urgent

PR #1915, opened 2015-10-27
PT-2192: Tests for patient indexing

buske:
PR originally by @veronikaslc, patched by @buske

PR #1912, opened 2015-10-26
PT-825: include phenotype ancestors in patient index

buske:
The actual phenotypes are copied into Solr fields according to their type:
- present features -> `phenotype`
- absent features -> `negative_phenotype`
- present prenatal features -> `prenatal_phenotype`
- absent prenatal features -> `negative_prenatal_phenotype`

The ancestor hierarchy is merged across all types and stored in two fields:
- `extended_phenotype` - all present phenotypes, regardless of type
- `extended_negative_phenotype` - all absent phenotypes, regardless of type

This PR replaces #1886 and #1909.

PR #1918, opened 2015-11-06
Patient index tests: genes and extended phenotypes

veronikaslc:
Adds tests for #1912 (PT-825) and #1914 (PT-2188).
Extends basic tests from #1915.


buske:
Closing in favor of integrating tests into PR: PT-2188

PR #1924, opened 2015-11-13
Integrating changes from Anton's commit 5bd437602a3239b0e11806132c9c7…

mjshepherd:
…5b4b0ea567b to use new gadgets


antonkatz:
But it won't work without the rest of family studies

allasm:
This PR will go into PT-1548 branch which IS the new 1.3-based family studies branch

PR #1925, opened 2015-11-17
Escape UIX page titles when needed

alexhenrie:
This pull request fixes two related bugs:

- When creating a new user interface extension, the title of the extension is cut off and displayed as garbage after the "Title" field.

- If no title parameter is specified, the patient form designer falls back to displaying the extension title, but incorrectly converts the double underscores to an underline.

See the screenshots below.

inline-form-editor

patient-form-designer


sdumitriu:
Thank you, see https://phenotips.atlassian.net/browse/PT-2247

alexhenrie:
Thank you!

PR #1927, opened 2015-11-19
Fixes for PT-2229 and PT-2250

allasm:


components/pedigree/resources/src/main/resources/pedigree/model/helpers.js
```diff
@@ -9,12 +9,9 @@
...
-    // Fixes bug introduced after utilizing requirejs where the global console variable becomes undefined.
-    if (!console) { var console = window.console; }
-    
-    // To allow debug code to run in IE7 && IE8
-    if (!window.console) { var console = {log: function() {}}; }
+
+    // To allow debug code to run in IE7 && IE8 && IE9
+    if (!window.console) { window.console = {log: function() {}}; }
 
     // For IE7 && IE8 again
```
veronikaslc: dont we aiming IE only from 9?

components/pedigree/resources/src/main/resources/pedigree/view.js
```diff
@@ -717,22 +717,24 @@ define([
 
             //timer.printSinceLast("=== highlight: ");
 
-            // re-evaluate which buttons & handles are appropriate for the nodes (e.g. twin button appears/disappears)
-            for (var nodeID in this._nodeMap) {
-                if (this._nodeMap.hasOwnProperty(nodeID)) {
-                    if (editor.getGraph().isPerson(nodeID) && !this.getNode(nodeID).getGraphics().getHoverBox().isMenuToggled()) {
-                        this.getNode(nodeID).getGraphics().getHoverBox().removeButtons();
-                        this.getNode(nodeID).getGraphics().getHoverBox().removeHandles();
+            if (!editor.isReadOnlyMode()) {
```
veronikaslc: As far as I understand, I am fixing [PT-2096](https://phenotips.atlassian.net/browse/PT-2096) by totaly disallowing users with read-only rights to ever enter pedigree-editor. Why we still taking care of this here?

allasm: This is "different" read-only mode. E.g. the user may have edit rights, but we disallow editing in old browsers, e.g. IE8 (since edit mode is harder to make compatible than view-only mode). Also, view-only mode can be used for pedigree preview on patient/family page (instead of a static SVG image), or to render template previews (since those are not editable, even in an editable pedigree)

components/pedigree/resources/src/main/resources/pedigree/model/helpers.js
allasm: we now do, but since compatibility code is already there, why remove it. The issue at hand that this commit fixes - problems with console - affects IE9 as well. If at some point even view does not work in IE8 it will be disallowed on PT level (we do a warning right now but let user continue), but for now there is no hard reason to disallow viewing pedigrees in IE8

components/pedigree/resources/src/main/resources/pedigree/view/workspace.js
```diff
@@ -139,17 +140,17 @@ define([
             // set display:block
             svgText = svgText.replace(/(<svg[^<>]+style=")/g, "$1display:block; ");
```
veronikaslc: @meatcar just out of curiosity, is it looks like the case of chaining .replace() which is inefficient (like it was in my code for variants), or we can not optimize here (like template evaluating)?

allasm: It is definitely inefficient. But it is done once on save, and once on print, so maintainability/simplicity/speed of development is more important than the speed of regexp evaluation which is 0 for all practical purposes here.

meatcar: The issue with this code is definitely maintainability. If you ask me, all regex is evil, unmaintainable (even by the person who wrote it), and spooky. It looks like what you're doing is modifying the DOM structure, by first doing a `el.innerHTML()`, and then nuking parts of the string with regex. Why not just pick the elements you want to remove with selectors?

allasm: Good point. Unfortunately, there is a reason - the idea is to get a copy of the SVG, and leave the original as is. It is possible to modify DOM for the original, but then all of the things removed should be added back, all even handlers for removed parts restored, etc. This can be done, but it is very messy to re-run all the code which generated all of those things which need to be removed in the copy (@marta also think we should investigate this path, and I agree it is a good idea in general but it is not easy). This may also be visibly slow for a large pedigree (though this needs to be tested)

P.S. There are some things which are also hard to do in DOM due to browser bugs. E.g. when a gradient is used in an SVG, Chrome (and only Chrome) creates a <linearGradient> element for each gradient. And when an element is removed, that <linearGradient> is not removed (which is kind of ok in itself, but makes SVG messy). But even worse, if a copy of an SVG is used with the same <linearGradient>, and then the copy is removed, the gradients will break in the original SVG as well, even though original still has the <linearGradient> - though not sure what happens when they are removed and then redefined anew

meatcar: Maybe instead of removing elements from the pedigree to display a slimmed
down one, we should instead re-do how a pedigree is built, and the
slimmed-down pedigree would be generated by building a pedigree, and
stopping before you add all the fancy features that are removed here. Also,
RE: cloning node, take a look at [Node.cloneNode()](
https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode). The
performance is ~20% of innerHTML.

On Thu, Nov 19, 2015 at 5:51 PM, allasm <notifications@github.com> wrote:

> In
> components/pedigree/resources/src/main/resources/pedigree/view/workspace.js
> <https://github.com/phenotips/phenotips/pull/1927#discussion_r45413310>:
>
> > @@ -139,17 +140,17 @@ define([
> >              // set display:block
> >              svgText = svgText.replace(/(<svg[^<>]+style=")/g, "$1display:block; ");
>
> Good point. Unfortunately, there is a reason - the idea is to get a copy
> of the SVG, and leave the original as is. It is possible to modify DOM for
> the original, but then all of the things removed should be added back, all
> even handlers for removed parts restored, etc. This can be done, but it is
> very messy to re-run all the code which generated all of those things which
> need to be removed in the copy (@marta <https://github.com/marta> also
> think we should investigate this path, and I agree it is a good idea in
> general but it is not easy). This may also be visibly slow for a large
> pedigree (though this needs to be tested)
>
> P.S. There are some things which are also hard to do in DOM due to browser
> bugs. E.g. when a gradient is used in an SVG, Chrome (and only Chrome)
> creates a element for each gradient. And when an element is removed, that
> is not removed (which is kind of ok in itself, but makes SVG messy). But
> even worse, if a copy of an SVG is used with the same , and then the copy
> is removed, the gradients will break in the original SVG as well, even
> though original still has the - though not sure what happens when they are
> removed and then redefined anew
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phenotips/phenotips/pull/1927/files#r45413310>.
>



-- 
Denys Pavlov

veronikaslc:
+1

components/pedigree/resources/src/main/resources/pedigree/view/workspace.js
sdumitriu: I agree with @meatcar, you should make DOM changes, but not on the actual DOM that's displayed, on an in-memory clone.

sdumitriu: But that's something for later.

allasm: I'm afraid with a large SVG DOM may be visibly slow - but that needs to be tested (in general from my experience DOM manipulation is MUCH slower than text manipulation, while being safer/less prone to mistakes, of course - the slow part is not to clone, but to do actual modifications) - especially as for say print preview we need effectively N copies of SVG, with different viewports (then again, can do an in-memory copy, do DOM manipulations, and only change vieports as text)

meatcar: DOM manipulation of a rendered DOM is expensive, since each change causes
the page to re-render. An in-memory DOM subtree that has not been added to
the page is a bit less expensive. What you could also do is use a "virtual
DOM" library like mithril.js to generate a virtual DOM, which then you
render to actual dom elements when nescessary. (http://mithril.js.org/)

I'm just talking about all the hypothetical ways to "skin a cat." You did
say that in this particular usecase, the performance isn't a huge issue
since it's only done once, so I doubt we need to involve a whole new
library in this :P

On Thu, Nov 19, 2015 at 10:14 PM, allasm <notifications@github.com> wrote:

> In
> components/pedigree/resources/src/main/resources/pedigree/view/workspace.js
> <https://github.com/phenotips/phenotips/pull/1927#discussion_r45432199>:
>
> > @@ -139,17 +140,17 @@ define([
> >              // set display:block
> >              svgText = svgText.replace(/(<svg[^<>]+style=")/g, "$1display:block; ");
>
> I'm afraid with a large SVG DOM may be visibly slow - but that needs to be
> tested (in general from my experience DOM manipulation is MUCH slower than
> text manipulation, while being safer/less prone to mistakes, of course -
> the slow part is not to clone, but to do actual modifications) - especially
> as for say print preview we need effectively N copies of SVG, with
> different viewports (then again, can do an in-memory copy, do DOM
> manipulations, and only change vieports as text)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phenotips/phenotips/pull/1927/files#r45432199>.
>



-- 
Denys Pavlov

antonkatz: Why not do all of this in Java? Send the current DOM down to the server, and have the server parse it?

allasm: This seems to be blown out of proportion :)

#1 reason for not sending to server is 10 lines of js code vs 100+ lines of js+velocity+java code, since 10 lines of regexp in JS would have to be replaced with 1) at least as many lines to call Ajax, and asynchronously wait for a reply 2) new script service or new (unrelated to anything else) entry point for an existing one 3) new java classes to process SVG + time figuring out how Java libs handle SVG structure (I presume regular html/xml parsers would do, but never tried). Each js regexp line would also likely have to be replaced with more than one java line on average as well + boilerplate

allasm: @meatcar - those are all good ideas, but a bit more on performance: performance of 10 stacked regexps is not as good as say one regexp or something smarter, but slowdown is negligible (even if it is 10x, the base time is ~0). But (potentially) DOM manipulation for a large SVG is 100x or 1000x the time, and may become noticeable (e.g. just adding all possible years to a datepicker via DOM manipulation offscreen is noticeably slow, while composing HTML via javascript and doing one assignment is fast.). Virtual DOM should resolve this, but a straightforward implementation likely will become slow enough to worry about it, even if it is done once

antonkatz: @allasm Do you not save the pedigree by calling velocity/java?

allasm: 1) no velocity/java in 1.2 (we start using it for family-studies only)

2) it is used not only for saving, but e.g. for print preview, which is completely inside JS/pedigree and will stay this way for 1.3/family-studies

PR #1930, opened 2015-11-20
PT-2234: Replace the full pedigree with its thumbnail in the Family history and pedigree section.

veronikaslc:
Fixed.
[cherry-piked from PT-2234 branch to resolve conflict]


allasm:
Cherry-picked into PT-1548

PR #1932, opened 2015-11-20
More PMD optimizations

alexhenrie:
[UseCollectionIsEmpty](http://pmd.sourceforge.net/pmd-5.2.0/pmd-java/rules/java/design.html#UseCollectionIsEmpty): Collection.size() == 0 is O(N) if the collection is a linked list, whereas Collection.isEmpty() is always O(1) and it's also easier to read.

[SimplifyConditional](http://pmd.sourceforge.net/pmd-5.2.0/pmd-java/rules/java/design.html#SimplifyConditional): "null instanceof ClassName" always returns false (the null check is built in).

[AddEmptyString](http://pmd.sourceforge.net/pmd-5.2.0/pmd-java/rules/java/optimizations.html#AddEmptyString): Using "" with the + operator creates an extra string object unnecessarily.

### DIFF
--- a/components/xwiki-platform-users/xwiki-platform-users-wiki/src/main/java/org/xwiki/users/internal/WikiUser.java
+++ b/components/xwiki-platform-users/xwiki-platform-users-wiki/src/main/java/org/xwiki/users/internal/WikiUser.java
@@ -94,10 +94,11 @@ public class WikiUser extends AbstractUser
         if (this.profileReference == null) {
             return "";
         }
-        String result = this.bridge.getProperty(this.profileReference, this.classReference, "first_name") + "";
-        result = StringUtils.trim(result) + " ";
-        result +=
-            StringUtils.trim("" + this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        String firstName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "first_name"));
+        String lastName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        String result = StringUtils.trim(firstName) + " " + StringUtils.trim(lastName);
         if (StringUtils.isBlank(result)) {
             result = getUsername();
         }
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -207,7 +207,7 @@ public class DataToCellConverter
             if (present.contains("meta") || present.contains("meta_code")) {
                 int mX = x;
                 Collection<? extends FeatureMetadatum> featureMetadatum = feature.getMetadata().values();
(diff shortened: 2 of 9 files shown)


components/xwiki-platform-users/xwiki-platform-users-wiki/src/main/java/org/xwiki/users/internal/WikiUser.java
```diff
@@ -94,10 +94,10 @@ public String getName()
         if (this.profileReference == null) {
             return "";
         }
-        String result = this.bridge.getProperty(this.profileReference, this.classReference, "first_name") + "";
+        String result = this.bridge.getProperty(this.profileReference, this.classReference, "first_name");
```
sdumitriu: This causes compilation failure, the `+""` was used to convert it to a string. `String.valueOf` would be better.

alexhenrie: You're right. I forgot to compile the project again after this last change. I will force-push a fix momentarily.

components/xwiki-platform-users/xwiki-platform-users-wiki/src/main/java/org/xwiki/users/internal/WikiUser.java
```diff
@@ -94,14 +94,20 @@ public String getName()
...
         }
-        String result = this.bridge.getProperty(this.profileReference, this.classReference, "first_name") + "";
-        result = StringUtils.trim(result) + " ";
-        result +=
-            StringUtils.trim("" + this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        String firstName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "first_name"));
+        String lastName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        StringBuilder sb;
```
sdumitriu: Not initialized, will trigger NPE two lines down.

components/xwiki-platform-users/xwiki-platform-users-wiki/src/main/java/org/xwiki/users/internal/WikiUser.java
```diff
@@ -94,14 +94,15 @@ public String getName()
...
-            StringUtils.trim("" + this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        String firstName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "first_name"));
+        String lastName =
+            String.valueOf(this.bridge.getProperty(this.profileReference, this.classReference, "last_name"));
+        String result = StringUtils.trim(firstName) + " " + StringUtils.trim(lastName);
         if (StringUtils.isBlank(result)) {
             result = getUsername();
         }
-        return result.trim();
```
sdumitriu: This might still be needed for example when firstname is set, but lastname is empty. This will show up as "firstname ".

PR #1939, opened 2015-11-25
Prefer "true" and "false" to calling Boolean.toString with a constant

alexhenrie:
This is mainly just to simplify the code base, although it may also offer a slight performance improvement.

### DIFF
--- a/components/solr-access-service/api/src/main/java/org/phenotips/solr/AbstractSolrScriptService.java
+++ b/components/solr-access-service/api/src/main/java/org/phenotips/solr/AbstractSolrScriptService.java
@@ -429,8 +429,8 @@ public abstract class AbstractSolrScriptService implements ScriptService, Initia
         if (StringUtils.isNotBlank(sort)) {
             result.put(CommonParams.SORT, sort);
         }
-        result.put("spellcheck", Boolean.toString(true));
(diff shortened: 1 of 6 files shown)


sdumitriu:
This will trigger checkstyle exceptions, duplicate strings are not allowed. We could change the checkstyle rules to allow `true` and `false` as special strings.

marta-:
The checkstyle rules are like that for a reason. This particular rule prevents errors caused by typos in the strings. It happens surprisingly often.

Bottom line: the checkstyle rules will not be changed; however, we are open to changes with the same purpose that do not imply repeating constant strings.

Thanks!

sdumitriu:
And I'm not sure that this really has a performance benefit. If not the compiler, the JVM JIT optimizer can easily detect that this method call always has the same effect and just replace it with the constant.

sdumitriu:
Ran a quick test, using string constants is about 25% faster than `Boolean.toString` calls, but the time gained is insignificant. 10M calls take less than 100ms, and the code optimized is not that critical for performance.

sdumitriu:
Actually my test was flawed, the second time also included the time needed to allocate the buffer. After correcting it, there is no difference between the performance of either method.

Given that the current code is easier to understand, it should stay as it is.

PR #1940, opened 2015-12-02
Return the patient ID after creating a new patient via the REST API

alexhenrie:
The Location header in the HTTP response is always <site>/rest/patients. [The PhenoTips REST documentation](https://phenotips.org/DevGuide/RESTfulAPI#add) says that it's actually supposed to contain a reference to the newly created patient, e.g. /rest/patients/P000123. The REST client does not have a good way to get the new patient ID otherwise.

[UriBuilder.build](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriBuilder.html#build-java.lang.Object...-) is for substituting template parameters, which we don't have. [UriBuilder.path](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriBuilder.html#path-java.lang.String-) is what we want to simply append a slash and a string.

### DIFF
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImpl.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsResourceImpl.java
@@ -104,8 +104,8 @@ public class DefaultPatientsResourceImpl extends XWikiResource implements Patien
             Patient patient = this.repository.createNewPatient();
             patient.updateFromJSON(jsonInput);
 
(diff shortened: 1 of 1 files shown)


sdumitriu:
Thanks for reporting this. The issue isn't exactly that `UriBuilder.build` doesn't append a parameter, but that it's actually been given the wrong type of resource. See 852c1673de8d93e7dc10a7a2761b5ccd70497117

By the way, the proper issue tracker is at https://phenotips.atlassian.net/

alexhenrie:
Great, thanks for the fix.

PR #1942, opened 2015-12-03
PT-2272: Can't open two history windows at the same time

alexhenrie:
Changed target="__blank" to target="_blank"

This looks like a typo. I fixed it with `find -type f | xargs sed -i "s/__blank/_blank/"`


buske:
My understanding is that `rel="__blank"` (two underscores) is an XWiki convention to be XHTML-compliant, with javascript to provide functionality, and should not be changes. If `target="__blank"`, that is probably a typo, and should either be changed to one underscore or instead use the `rel` attribute.

alexhenrie:
Okay, I've amended the commit to only change the target attributes and leave rel alone.

PR #1944, opened 2015-12-04
PT-1447: Better usability for the pedigree Date Picker

mjshepherd:


mjshepherd:
@allasm @meatcar please review

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
+      this.element.observe("keydown", this.onKeyDown.bind(this));
+    },
+    getValue : function() {
+      return this.element.value;
+    },
+    isEmpty : function() {
+      return this.getValue() === "";
+    },
+    setValue : function (value) {
+      value && this.enable();
```
meatcar: you're using `&&` as an `if` here. Just use an if, it's more readable.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
+    getElement : function() {
+      return this.element;
+    },
+    onDateChange : function() {
+      this.refreshAvailableValues();
+      this.validate();
+      if (!this.isValid && (this.getValue() !== "")) {
+        this.element.addClassName("invalid");
+      } else {
+        this.element.removeClassName("invalid");
```
meatcar: you can use [`element.toggleClassName`](http://prototypejs.org/doc/latest/dom/Element/toggleClassName/) instead.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: I think in JS this is a very common method of using a variable only if defined

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: we still need an IF because we check for validity there (i.e. can't just do a "toggle"), and once inside the branch, it is safer to explicitly add or remove, to arrive to the desired state regardless of the previous state

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
-    enable : function () {
-      if (!this.options.alwaysEnabled) {
-          this.dropdown.enable();
-          if (this.dropdown.selectedIndex <= 0 && this._tmpSelectedIndex < this.dropdown.options.length) {
-            this.dropdown.selectedIndex = this._tmpSelectedIndex;
-            return (this._tmpSelectedIndex > 0);
-          }
-      }
-      return false;
+      return this.availableValues.map(function(a){return a.indexOf(currentVal) == 0}).reduce(function(a,b){return a || b;});
```
meatcar: I may be wrong, but couldn't you use [`Array.every`](http://prototypejs.org/doc/latest/language/Array/prototype/every/)?

meatcar: Wait. I mean [`Array.some`](http://prototypejs.org/doc/latest/language/Array/prototype/some/)

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
sdumitriu: +1 for explicit `add` and `removeClassName`. `toggle` may give the wrong result if the previous state wasn't the expected one.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
sdumitriu: The pedigree code is trying to be free of Prototype.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
meatcar: `toggle` takes a boolean..

```js
this.element.toggleClassName("invalid", !this.isValid && this.getValue() !== "");
```

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: Just a note that we still need to make IE9 compatible code and test in IE9 if using some more modern JS extensions (I think all of {Prototype.every, JS.map , JS.reduce } work in IE9, but it sometimes fails in some very basic and surprising ways)

meatcar: [It's present in all browsers since IE9, like `Array.reduce`](http://devdocs.io/javascript/global_objects/array/reduce)

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: well, if we advocate for readabilty, the line above may be a bit less obvious for someone not involved in JS development :) (and toggle may still not be safe is the previous state is not what you expect)

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: I did not mean to say it is not (I checked myself and it was), just mentioned that we need to test all other place sin the code.

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -967,26 +967,14 @@
...
-  max-width: 48px;
-}
-.fuzzy-date-picker .day {
-  font-size: 14px;
-  max-width: 48px;
-}
-.fuzzy-date-picker .empty {
-  font-size: .7em;
-  opacity: .5;
+.fuzzy-date-picker input[type=text].text {
```
meatcar: you don't have any non-`[type=text]` inputs, should we still be getting super specific with the `.text` class selector?

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: I was wrong, toggleClassName with a boolean parameter works just as we need it to here. So functionality-wise it is a good solution, though some (e.g. me :) may not immediately know how this boolean works (toggle only if boolean is true, or toggle to the state given by the boolean)

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
+      //The isFuzzy boolean specifies whether the value contained in the iput is specific or fuzzy(a range, decade etc.). Blank and invalid values are not fuzzy.
+      this.isFuzzy = false;
+      this.isValid = true;
+
+      if (this.alwaysEnabled) {
+        this.enable();
+      } else {
+        this.disable();
+      }
+      this.element.onkeypress = this.onKeyPress.bind(this);
```
meatcar: how come you didn't use `observe("keypress", ...`?

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
+            this._input.fire("xwiki:date:changed");
+          }
         }
+    },
+    initStrings: function() {
+      var strings = {
+        dayPlaceholder: 'DD',
+        monthPlaceholder: 'MM'
+      };
+      strings.yearPlaceholder = this.decadesEnabled ? 'YYYYs' : 'YYYY';
```
meatcar: you can have this ternary inside the `{}` above. Then, you don't even need this function, you can just inline the object creation right into the constructor.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
mjshepherd: I think I will switch to array.some, seems more efficient

components/pedigree/resources/src/main/resources/pedigree/editor.css
mjshepherd: uh yeah... Had to add that class to add specificity...

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,437 @@
...
-        var y = this.yearSelector.getSelectedValue();
+        var y = this.inputs.y.getValue();
         if (y.match(/\d\d\d\ds$/)) {
-            dateObject["decade"] = y;
+          dateObject["decade"] = y;
         } else {
-            if (y != "") {
-                dateObject["year"] = y;
-            }
+          if (y != "") {
```
meatcar: make this an `else if`

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
mjshepherd: I tried that first and it didn't work

allasm: should work as long as you initialize strings after you initialize this.decadesEnabled

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
marta-: Fwiw, I find both versions almost equally readable, and I prefer Matt's because it's shorter (which is also why it's so commonly used in JS).

sdumitriu:
Any +1s? What are we waiting for?

mjshepherd:
Yes this still needs review. I'll fix the conflict today

marta-:
@allasm @meatcar please re-review.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,445 @@
...
+      }
+    },
+    onKeyUp: function(e) {
+      //if the input is valid on keyUp the input will consider the date selected before the user tabs away
+      this.validate() && this.onValueSelected();
+    },
+<<<<<<< HEAD
+
+=======
+>>>>>>> PT-2274: Date is now saved before blur
```
meatcar: This is an artefact of a merge conflict, and shouldn't be commited.

meatcar: since its not valid javascript, and breaks the build.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,445 @@
...
-      });
+    validate: function() {
+      this.isValid = this.availableValues.indexOf(this.getValue()) > -1 || this.getValue() === "";
+      return this.isValid;
     },
+    formatValue : function(value) {return value;},
+<<<<<<< HEAD
+
+=======
+>>>>>>> PT-2274: Date is now saved before blur
```
meatcar: same here.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,276 +1,445 @@
...
+    onKeyPress : function(e, regexp, maxLength, padZeros) {
+      var charCode = e.charCode || e.keyCode;
+      var character = String.fromCharCode(charCode);
+      //Call select handler on tab press.
+      if(charCode == 9){
+        this.onValueSelected();
+        return true;
+      }
+      //Allow non-character keys always
+      if (character === "") {return true;}
```
veronikaslc: '===' is probably an overkill, since  String.fromCharCode() doesn't return an object

mjshepherd: What do you mean by overkill?

sdumitriu: It's not overkill. Many advocate for always using `===`, unless type coercion is explicitly wanted, since it helps avoid certain types of bugs, and has marginally better performance.

veronikaslc: by overkill I mean in that particular case we don't care if they are the same type, just want to consider non-character keys

allasm:
TODO: update to use new JSON date format

sashaandjic:
anything here that I should pay attention to?

veronikaslc:
I have tested it and it worked for me. This is the question to @allasm may be

sashaandjic:
@allasm: I noticed that drop down are no longer available. Wasn't sure if it was intentional or not.

PR #1945, opened 2015-12-04
PT-2275: Add ability to import gene data

antonkatz:


antonkatz:
@meatcar @mjshepherd please review

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,78 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+                Map<String, String> gene = iterator.next();
+                BaseObject xwikiObject = doc.newXObject(GENE_CLASS_REFERENCE, context);
+
+                for (String property : this.getProperties()) {
+                    String value = gene.get(property);
+                    if (value != null) {
+                        xwikiObject.set(property, value, context);
+                    }
+                }
+                // todo. this is deprecated in the new version, but without this setting the gene section breaks
```
allasm: will this code go in before the new version?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,78 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+            // todo. does not check for existing genes
+            PatientData<Map<String, String>> genes = patient.getData(getName());
+            if (!genes.isIndexed()) {
+                return;
+            }
+
+            XWikiContext context = (XWikiContext) this.execution.getContext().getProperty("xwikicontext");
+            Iterator<Map<String, String>> iterator = genes.iterator();
+            while (iterator.hasNext()) {
+                Map<String, String> gene = iterator.next();
```
allasm: As our previous struggles with items in a list show, maybe we should add a try-catch within the while(), so that we catch and omit any individual mis-formatted/broken-in-any-way genes individually, without failing the entire gene import? Genes will be a complicated entity, so we are somewhat likely to fail import of amn individual gene even if all other genes are ok?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,78 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+        }
+
+        try {
+            JSONArray genesJson = json.getJSONArray(this.getJsonPropertyName());
+            List<Map<String, String>> allGenes = new LinkedList<Map<String, String>>();
+            for (Object geneJsonUncast : genesJson) {
+                JSONObject geneJson = JSONObject.fromObject(geneJsonUncast);
+                Map<String, String> singleGene = new LinkedHashMap<String, String>();
+                for (String property : getProperties()) {
+                    // todo. it makes no sense to have no gene name, but to have comments, which can occur
```
allasm: I don't get the comment :) If it makes no sense, then when does it occur, and should we support it, and if yes why?

mjshepherd: ```this.getProperties()```

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,78 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+                    if (geneJson.has(property)) {
+                        String field = geneJson.getString(property);
+                        if (field != null) {
+                            singleGene.put(property, field);
+                        }
+                    }
+                }
+                allGenes.add(singleGene);
```
mjshepherd: Is there a possibility that `` singleGene `` is just an empty map at this point?

meatcar:
As per our discussion, since the `todo` comments are just instructions for things that need to be changed in the new gene section that is currently being developed, they should instead be JIRA tasks that document what needs to be done. No one reads `todo` comments once the code is merged :(

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,78 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+    public void save(Patient patient)
+    {
+        try {
+            XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
+            if (doc == null) {
+                throw new NullPointerException(ERROR_MESSAGE_NO_PATIENT_CLASS);
+            }
+
+            // todo. does not check for existing genes
+            PatientData<Map<String, String>> genes = patient.getData(getName());
```
mjshepherd: this.getName()

meatcar: is this a style that we prefer? In java we can refer to methods of our object without `this`, and my IDE highlights the method appropriately.

allasm: I think @sdumitriu advocates for the use of .this. I guess given different people work on the code and we sometimes have not the best code/a lot of methods&variables => a habit to place explicit .this may save a few second here and there when reading the code, at the expense of a few seconds to place and discuss this :)

mjshepherd: Yeah, we prefer this. I believe i've tried to omit ``this `` before but Sergiu caught me.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
antonkatz: good catch

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
sdumitriu: We use `this.` only for field access, not for methods. It's clear that a method without an explicit object to call on belongs to the current object, but a variable can either be a class field, a method argument, or a local variable. Explicitly prefixing fields makes it clear that we're accessing the object state. It can also help avoid bugs if we have two things with the same name, a field and an argument.

Anyway, I never write it manually, I set up Eclipse to do it for me: Window -> Preferences -> Java -> Editor -> Save Actions -> Add 'this' qualifier to unqualified field accesses

save-actions

mjshepherd:
+1

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -80,6 +82,10 @@
     @Inject
     private Logger logger;
 
+    /** Provides access to the current execution context. */
+    @Inject
+    private Execution execution;
```
sdumitriu: Use `Provider<XWikiContext>` instead.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,81 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
             }
         }
     }
+
+    @Override
+    public PatientData<Map<String, String>> readJSON(JSONObject json)
+    {
+        Object testGeneJson = json.get(this.getJsonPropertyName());
```
sdumitriu: if (!json.has(getJsonPropertyName())

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,81 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+            if (!genes.isIndexed()) {
+                return;
+            }
+
+            XWikiContext context = (XWikiContext) this.execution.getContext().getProperty("xwikicontext");
+            Iterator<Map<String, String>> iterator = genes.iterator();
+            while (iterator.hasNext()) {
+                try {
+                    Map<String, String> gene = iterator.next();
+                    BaseObject xwikiObject = doc.newXObject(GENE_CLASS_REFERENCE, context);
```
sdumitriu: You're adding new objects, but what happens if the patient already had some objects? You'll get duplicates, or have deprecated objects.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -183,4 +189,81 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
...
+    public void save(Patient patient)
+    {
+        try {
+            XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
+            if (doc == null) {
+                throw new NullPointerException(ERROR_MESSAGE_NO_PATIENT_CLASS);
+            }
+
+            PatientData<Map<String, String>> genes = patient.getData(this.getName());
+            if (!genes.isIndexed()) {
```
sdumitriu: You may get a NPE here.

PR #1949, opened 2015-12-11
PT-2706: Automatic phenotypes do not appear in differential

mjshepherd:


mjshepherd:
@antonkatz @veronikaslc please review

antonkatz:
+1. Though that code has finicky flow, so I expect bugs.

marta-:
@antonkatz if nominated reviewers also tested the change whenever possible in addition to just reading it, we migh end up with better quality code and fewer bugs...

veronikaslc:
+1

PR #1955, opened 2016-01-22
PT-1044: Automatically expand the list of subcategories (if it exists…

mjshepherd:
…) when selecting YES for a phenotype in the predefined phenotype list


mjshepherd:
@itaiGershtansky @danielpgross

sdumitriu:
This only works for the predefined terms, it doesn't apply for new terms that have been added by expanding a term.

PR #1958, opened 2016-01-28
PT-1044: Deep subterms will automatically expand when selecting YES

mjshepherd:


mjshepherd:
@itaiGershtansky  @danielpgross

sdumitriu:
I already implemented this in d52e5adead3e700ad499273dc70e6665997814fa

marta-:
Sorry @mjshepherd , didn't realize this was already done. @sdumitriu should have mentioned it.

sdumitriu:
Well, I did fix it but only for phenotypes displayed in the main patient record. The ontology browser ("Browse related terms") doesn't expand with my fix. @mjshepherd your fix works in the obrowser as well, right?

sdumitriu:
One thing I don't like is that it adds a light dependency between the obrowser and the ynpickers widget. The way the code is organized now, the base obrowser doesn't display ynpickers, just a plain checkbox. And this is how it still work, for example, on the "Browse all" page, when selecting phenotypes to filter on. So, things to check:

- does it autoexpand in the "Browse related terms" dialog?
- does it still work, without throwing any JS errors, in the "Browse related terms" dialog when used in the "Browse all data" page?

mjshepherd:
@sdumitriu TBH I completely forgot about the browse related terms. It works now. No errors are formed. I agree with you that the dependency is not great.

marta-:
@itaiGershtansky @veronikaslc please review. @danielpgross too, if you have time.

veronikaslc:
+1, works.

PR #1960, opened 2016-02-01
PT-1698: Vocabularies should auto-generate their Solr configuration at startup

veronikaslc:
Fixed


marta-:
@AlexLapena please test & review

AlexLapena:
+1 @marta-

sdumitriu:
I'd like to see two things:

- [ ] Configure all cores not to load automatically: put `loadOnStartup=false` in every `core.properties` file, and load them from java
- [ ] Also add a `phenotips.version` property in `core.properties`, configure the build to interpolate those files, and instead of always extracting the files from the jar, first check if the core exists already, and if yes, if the version in it is different from the current PhenoTips version.

PR #1962, opened 2016-02-02
Ping back number of users and cases from all installed PT instances

meatcar:
This depends on https://github.com/phenotips/pingback and a running instance of ElasticSearch, the configuration for which is provided in pingback repo.

### DIFF
--- a/components/base-war/pom.xml
+++ b/components/base-war/pom.xml
@@ -244,7 +244,7 @@ extension.repositories=maven-central:maven:https://nexus.phenotips.org/nexus/con
 extension.repositories=maven-phenotips:maven:https://nexus.phenotips.org/nexus/content/repositories/releases/
 extension.repositories=phenotips.org:xwiki:https://phenotips.org/rest/
 rendering.transformations=macro
-activeinstalls.pingURL=http://localhost:9200
+activeinstalls.pingURL=http://stats.phenotips.org
 activeinstalls.ipFetchURL=https://phenotips.org/get/Stats/Id
                 </xwikiPropertiesAdditionalProperties>
               </properties>
--- a/components/base-war/pom.xml
+++ b/components/base-war/pom.xml
@@ -245,6 +245,7 @@ extension.repositories=maven-phenotips:maven:https://nexus.phenotips.org/nexus/c
 extension.repositories=phenotips.org:xwiki:https://phenotips.org/rest/
 rendering.transformations=macro
 activeinstalls.pingURL=http://localhost:9200
+activeinstalls.ipFetchURL=https://phenotips.org/get/Stats/Id
                 </xwikiPropertiesAdditionalProperties>
               </properties>
             </configuration>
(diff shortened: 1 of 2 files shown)


meatcar:
Nominating @veronikaslc and @itaiGershtansky for review.

distribution/war/pom.xml
```diff
@@ -743,6 +743,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pingback-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pingback-client-api</artifactId>
+      <version>${projcet.version}</version>
```
itaiGershtansky: projcet instead of project

veronikaslc:
Docker Compose is currently not supported on the Windows operating system, please, re-assign a reviewer.

veronikaslc:
Codereview comments for phenotips/pingback:
-  logger seems to be unused in DistributionPingDataProvider
- net.sf.json is unmaintained, phenotips switched to org.json
- execution looks unused in PatientsPingDataProvider and UsersPingDataProvider
- comparing PatientsPingDataProvider and UsersPingDataProvider, they look identical code-wise and differ only by some constants such as a query string or logger error string, just curious if we can fit what they share in base class or other kind of optimization
- for other classes for ex. `Map<String, Object> map = new HashMap<>(); map.put("type", "string"); map.put("index", "not_analyzed");` happens in 6 of them, may be a room for optimization too

distribution/war/pom.xml
```diff
@@ -743,6 +743,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pingback-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>pingback-client-api</artifactId>
+      <version>${project.version}</version>
```
sdumitriu: If the pingback modules are in a different repository, this will fail when releasing.

components/base-war/pom.xml
```diff
@@ -243,6 +244,7 @@ extension.repositories=maven-central:maven:https://nexus.phenotips.org/nexus/con
 extension.repositories=maven-phenotips:maven:https://nexus.phenotips.org/nexus/content/repositories/releases/
 extension.repositories=phenotips.org:xwiki:https://phenotips.org/rest/
 rendering.transformations=macro
+activeinstalls.pingURL=http://localhost:9200
```
sdumitriu: Shouldn't this be pointing to phenotips.org?

meatcar: I'll change it once it passes first review, and I get access to phenotips.org to set up and secure elastic search.

distribution/war/pom.xml
meatcar: I guess we should add http://github.com/phenotips/pingback to jenkins then :)

meatcar:
@sdumitriu pending final review.

PR #1965, opened 2016-02-09
Use bold and color, not all caps, for headers in the patient form editor

alexhenrie:
Currently, the patient form editor displays headings in all caps, whereas the patient form itself does not. This makes it hard to see from the form editor how the form will actually be displayed.

I've put together a simple fix for this problem that displays the headings in bold and color instead of all caps:

**- Before -**
before

**- After -**
after


alexhenrie:
Thank you!

PR #1968, opened 2016-02-15
Fix XHTML syntax errors

alexhenrie:
There are numerous XHTML syntax errors in PhenoTips that prevent scraping pages as XML. The multiplicity of syntax errors also makes it hard to find other, more serious problems.

### DIFF
--- a/components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
+++ b/components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
@@ -701,7 +701,7 @@ $xwiki.ssx.use($className)##
     #end
     #set($displayedValue = "(% class='buttonwrapper' %)[[{{icon name='copy' /}} ${configTitle}&gt;&gt;${value}||rel='__blank' class='button secondary']]")
   #end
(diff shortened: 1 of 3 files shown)


components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -1715,7 +1715,7 @@ $xwiki.ssx.use('PhenoTips.PushPatient')##
 #foreach ($mapEntry in $configuredServers.entrySet())
   ## make sure PhenomeCentral is enabled/configured, and if it is list it as the first push target
   #if ($mapEntry.key.getServerID() == "PhenomeCentral")
-    (% id="push-patient-pc" name="PhenomeCentral" %)((([[$services.localization.render('phenotips.PushPatient.pushPhenomeCentral')&gt;&gt;path:#]])))
```
sdumitriu: This is used in the javascript code. HTML5 allows `data-*` attributes which could be used instead, but I'm not sure what kind of parser you're using and whether it would accept such an attribute or not. The other standard attribute that could be used is `title`.

Also note that this code is repeated a few lines below, for all other targets except PhenomeCentral.

alexhenrie: For now, I will just drop this commit from the pull request.

allasm:
For the record, PushPatient.xml has been refactored in a different branch

PR #1971, opened 2016-02-20
PT-1844: Make "create sibling" handle go to the right instead of to the left of the parent link

veronikaslc:


veronikaslc:
@allasm please review

allasm:
+1

PR #1973, opened 2016-03-08
Fix for and id attributes to link labels to inputs

alexhenrie:
ConfigurableClassMacros.xml expects attributes to be single-quoted just
like the ones that XWiki generates are. If double quotes are used
instead, it will change the for attributes on the labels without
changing the id attributes on the inputs to match.

If the input is inside a label tag, or there is no input tag at all, the
for attribute is not necessary.

### DIFF
--- a/components/application-configuration/ui/src/main/resources/PhenoTips/DBConfigurationClass.xml
+++ b/components/application-configuration/ui/src/main/resources/PhenoTips/DBConfigurationClass.xml
@@ -51,7 +51,7 @@
       <customDisplay>{{velocity}}
 #if ($xcontext.action == 'edit' || $xcontext.action == 'admin')
 {{html wiki=false clean=false}}
-&lt;select name="$prefix$name" id="$prefix$iname"&gt;
+&lt;select name="$prefix$name" id="$prefix$name"&gt;
 #foreach ($val in ['MM/dd/yyyy', 'dd/MM/yyyy', 'yyyy-MM-dd', 'MM/yyyy', 'MMMM yyyy'])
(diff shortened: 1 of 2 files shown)


alexhenrie:
@sdumitriu Ping?

sdumitriu:
I'm not sure I understand what you mean by this:

> If double quotes are used
instead, it will change the for attributes on the labels without
changing the id attributes on the inputs to match.

alexhenrie:
This code modifies the `for` and `id` attributes, but only if they use single quotes:

https://github.com/phenotips/phenotips/blob/38f91e6351b22258a4bb3265f306afbed8915a9f/resources/ui/src/main/resources/XWiki/ConfigurableClassMacros.xml#L262

        ## Step 3: Prepend app name to all ID and FOR attributes to prevent id collision with multiple apps on one page.
        #set ($out = $out.replaceAll(
            " id='$objClass.getName()_$obj.getNumber()_$propName", " id='${escapedAppName}_$objClass.getName()_$obj.getNumber()_$propName").replaceAll(
            " for='$objClass.getName()_$obj.getNumber()_$propName", " for='${escapedAppName}_$objClass.getName()_$obj.getNumber()_$propName"))
        ## App Name is prepended to for= to make label work with id which is modified to prevent collisions.

If `for` uses single quotes and `id` uses double quotes, only `for` will be modified, and `for` and `id` will no longer match.

sdumitriu:
Ah, I see now. Wouldn't it be simpler to update the code that changes things to allow both single and double quotes? Singe the code uses `replaceAll`, which accepts a regular expression, it would be enough to use something like:

    replaceAll("( id=['""])($objClass.getName()_$obj.getNumber()_$propName)", "$1${escapedAppName}_$2")

alexhenrie:
I would be OK with that.

alexhenrie:
The commit that I just force-pushed has the changes you requested.

PR #1974, opened 2016-03-09
PT-1941, PT-1936: Synchronize consent across instances and configurable set of fields for push

allasm:
Fixes/implements PT-1936, PT-1941, PT-1976, PT-2420, PT-2396, PT-2413, PT-2426


allasm:
@meatcar , @veronikaslc - please review

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -93,27 +157,21 @@
...
-  #set($activeFields = ['external_id', 'gender', 'phenotype', 'negative_phenotype', 'omim_id'])
-  #set($ignored = {})
-  #set($customValues = {})
+#if ($request.push)
+  #set($customValues = {
+       'allergies' : "$services.localization.render('PhenoTips.AllergiesDataClass_allergies')",
+       'genes': "$services.localization.render('phenotips.exportPreferences.field.genes')",
+       'genes_comments': "$services.localization.render('phenotips.exportPreferences.field.genesComments')",
+       'rejectedGenes': "$services.localization.render('phenotips.exportPreferences.field.rejectedGenes')",
+       'rejectedGenes_comments': "$services.localization.render('phenotips.exportPreferences.field.rejectGenesComments')"
```
veronikaslc: We'll have to change this for new genes-variants?

allasm: 1) yes, if this code is used
2) this code is actually no longer used after Marta's changes (https://github.com/phenotips/phenotips/commit/f7581aa1f7c8cd299382204759fb009cc804b327), so should be removed

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -218,40 +207,40 @@
...
+        "$services.localization.render('phenotips.patientSharing.push.dialog.consents.label')",
+        "$services.localization.render('phenotips.patientSharing.push.dialog.consents.defaulthint')"));
+
+      // TODO: a sectionto preview patient data being pushd. It used to be easy, but
+      //       now that we push a lot of different fields the preview is disabled until
+      //       we come up with an easy way to display all of the data
+      // if (!this._pushManyPatients) {
+      //   this._container.insert(generateSection('fields-selection',"$services.localization.render('phenotips.patientSharing.push.dialog.preview.label')","$services.localization.render('phenotips.patientSharing.push.dialog.preview.hint')"));
+      //   this._fieldsManager = this._container.down('#fields-selection .section-contents');
+      // } else {
```
meatcar: might have commented on this before, but commented/out code shouldn't be commited. We have the git history to go look for it :)

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -332,7 +328,10 @@
       this._patientData = undefined;
 
       this._groupManager.update(this._generateServerAndUserFirstMessage());
-      !this._pushManyPatients &amp;&amp; this._fieldsManager.update(this._generateServerAndUserFirstMessage());
+      this._consentsSection.hide();
+
+      // TODO: fields sectionis disabled for now
+      // !this._pushManyPatients &amp;&amp; this._fieldsManager.update(this._generateServerAndUserFirstMessage());
```
meatcar: shouldn't comment out code.

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -713,11 +716,14 @@
...
-        this._generateDataPreview(fieldList);
-      }
+
+      this._showApproveCheckbox();
+      // TODO: patient preview is disabled for now
+      // if (this._pushManyPatients) {
+      //   this._showApproveCheckbox();
+      // } else {
+      //  this._generateDataPreview(fieldList);
+      // }
```
meatcar: shouldn't comment out code.

components/application-configuration/api/src/main/java/org/phenotips/configuration/internal/consent/DefaultConsentAuthorizer.java
```diff
@@ -76,77 +81,81 @@
...
+        if (missingFields.size() == 0 || missingFields.contains(element.getExtension().getId())) {
+            return false;
         }
         return true;
     }
 
-    private static List<String> getGrantedIds(List<Consent> consents)
+    /**
+     * Returns the union of all fields which the provided set of missing consents prevents form
+     * being used. Returns an empty list of all fields are affected. Returns null if no fields are affected.
```
meatcar: form -> from.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -48,27 +49,27 @@ phenotips.PushPatient.male=Male
...
 phenotips.PushPatient.oneServerMessage=There is only one configured remote server: 
-phenotips.PushPatient.patientUploadedMessage=This patient has been uploaded to this server
+phenotips.PushPatient.patientUploadedMessage=This patient was uploaded to
 phenotips.PushPatient.patientUploadedLessThanHour=less than an hour ago
 phenotips.PushPatient.patientUploadedDays=days ago
 phenotips.PushPatient.patientUploadedHours=hours ago
-phenotips.PushPatient.patientUploadedAsID=as
-phenotips.PushPatient.patientUploadedUpdate=Further pushes will attempt to update the patient on the remote server.
+phenotips.PushPatient.patientUploadedAsID=and was assigned patient ID
+phenotips.PushPatient.patientUploadedUpdate=If data in the record has changed, push to __serverID__ again to update the record there.
```
meatcar: Seems that we're using `__identifier__` in some translation strings, and `{0}` in others. Not your fault, but just pointing it out.

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
```diff
@@ -267,84 +299,61 @@
...
         },
 
         /**
          * @param callback function fired when the consent are successfully loaded and displayed
+         * @param errorBlock (optional) an HTML block which may have elements with the following IDs, which
+         *        (if present) are displayed and hidden depending on the state of consent data:
+         *        "#patient-consent-update" - supposedly an "update" button, hidden when any errors in consent data are present
+         *        ".view-title-required-not-granted" - a block shown when a required consent is not granted (in view mode)
+         *        ".view-title-none-granted" - a block shown when no consents have been granted (in view mode)
+         *        ".error_no_consent_data" - a block (supposedly with an error message) displayed when no consent data was passed to the script
```
meatcar: Why do we so carefully annotate some JS functions, and not others?

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
```diff
@@ -384,33 +406,22 @@
...
           {
             var checkbox = $(event.target);
-            var rest = _this.interactWithManagingRest(checkbox.attr("data-bind"), baseUrl);
+            var name = checkbox.attr("data-bind");
             if (checkbox.is(":checked")) {
-              rest("grant");
+              callback(name, true);
             } else {
-              rest("revoke");
+              callback(name, false);
```
meatcar: you can rewrite this as 

```java
callback(checkbox.attr("data-bind", checkbox.is(":checked:");
```

meatcar:
:+1:

components/application-configuration/api/src/main/java/org/phenotips/configuration/internal/consent/ConsentAuthorizer.java
```diff
@@ -63,13 +62,12 @@
     boolean authorizeInteraction(Patient patient);
 
     /**
-     * Determines if any interactions with a patient record given the consents that have been granted should be
-     * allowed.
+     * Determines if the set of granted consents incldes all the required consents configured in the system.
```
sdumitriu: includes

components/application-configuration/api/src/main/java/org/phenotips/configuration/internal/consent/DefaultConsentAuthorizer.java
sdumitriu: of all fields -> if all fields

components/default-settings/src/main/resources/XWiki/XWikiPreferences.xml
```diff
@@ -1413,9 +1413,21 @@
...
+        <name>AllowNoConsentsFromOldClients</name>
+        <number>1</number>
+        <prettyName>AllowNoConsentsFromOldClients</prettyName>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.BooleanClass</classType>
+      </AllowNoConsentsFromOldClients>
     </class>
     <property>
-      <AllowPushesFromNonListedServers>1</AllowPushesFromNonListedServers>
+      <AllowPushesFromNonListedServers>0</AllowPushesFromNonListedServers>
```
sdumitriu: Why this change?

components/patient-data-sharing/receiver-ui/src/main/resources/data/ReceivePatientConfiguration.xml
```diff
@@ -77,6 +77,13 @@ $xwiki.ssx.use($displayDocumentName)##
     #end
     $doc.display("AllowPushesFromNonListedServers", "edit", $targetObj)
     ##
+  #if ($services.consentauthorizer.consentsGloballyEnabled())
+    {{html clean="false"}}
+    Accept data from old clients which do not ask the user to set consents when pushing: ##
```
sdumitriu: Why surround this plain text with `{{html}}`?

components/default-settings/src/main/resources/XWiki/XWikiPreferences.xml
allasm: If we include push functionality by default (i.e. it will always be listening for pushes), we don't want people to receive pushes they don't expect - of course, one needs a username and password, but even then maybe they don't want that. So this is a way to disable push while still including the component

components/patient-data-sharing/receiver-ui/src/main/resources/data/ReceivePatientConfiguration.xml
allasm: No good reason, I just wanted to exactly replicate the previous option 1-to-1 (and there checkbox title was within {{html}}), did not check if velocity will generate any extra HTML tags (e.g. an extra <p> or <div> with some style) when {{html}} is included or not.

components/default-settings/src/main/resources/XWiki/XWikiPreferences.xml
sdumitriu: Indeed, this is good.

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
allasm: I wanted to separate UI implementation form the logic, e.g. make it clear for un-initiated (like myself) that checkbox.attr("data-bind") is actually the name of the consent. Would be easier if we had named functional parameters in JS, e.g. callback( name:checkbox.attr("data-bind"), status: checkbox.is(":checked:") )

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
allasm: Because we are too lazy to annotate all? :) In this case, the function depends on external HTML block having certain elements with certain IDs/class names, which is not obvious for the user. The right way is to completely rewrite this, but in the rush I thought it is better to document and move on to refactoring more problematic areas

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
meatcar: You can add a comment like "pass name to callback". `if (x) true else false` is a common anti-pattern.

Another point I just noticed, you can eliminate jQuery from this function entirely, since all the features are implemented across all of our supported browsers.

```js
function (e) { callback(e.target.dataset.bind, !!e.target.checked) }
```

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
marta-: +1, should use {n} everywhere

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
allasm: Good catch re: jquery. It was used before, I removed that piece of code and did not realize jquery is no longer needed

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
allasm: How do we replace {0} when the parameter that comes from javascript and is not available during velocity pre-processing? Should we just emulate what velocity does, or we have a JS library that does this - I don't think I saw one, but I did not look hard

sdumitriu: Just like you replace `__serverID__`, you would replace `{0}` instead.

meatcar: Or you can bundle a small spritf JavaScript implementation, so we don't
have regex all over the place.

components/application-configuration/api/src/main/java/org/phenotips/configuration/internal/consent/ConsentAuthorizer.java
```diff
@@ -54,8 +54,7 @@
     List<RecordElement> filterForm(List<RecordElement> elements, Patient patient);
 
     /**
-     * Determines if any interactions with a patient record given the consents that have been granted should be
-     * allowed.
+     * Determines if all required (by the system) consents have been granted for the given patient record.
      *
      * @param patient record in question
      * @return {@link true} if the patient record can be modified or viewed, {@link false} otherwise
```
sdumitriu: `{@link true}` is invalid, it breaks the javadoc generation. This should be `{@code true}`. Same for `false`, and for all the other places where this is used.

allasm: Hmm, I guess we should add this to codestyle check? I just kept using what was in the code and never checked javadoc explicitly, which is I guess the workflow of most of us

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
allasm: BTW, I think __description_of_expected_variable__ (in some way or form) is more useful than {0} (including for human translators, might not immediately know what is supposed to go in place of {0} in all sentences, and maybe should use a different article or what not in whatever language, depending on the exact expected meaning of the inserted part)

sdumitriu: I agree with @allasm that if `{n}` isn't required, as is the case for translations used from Java/Velocity, a more descriptive variable name is better.

marta-: Our prospective translators are already confused enough about what words to translate and what words not to translate. Adding yet another rule won't help. We already have a convention for parameters, we should stick to it.

PR #1977, opened 2016-03-17
PT-2154: Migrate existing pedigrees

veronikaslc:
Fixed.

### DIFF
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyPermissions.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyPermissions.java
@@ -121,7 +121,15 @@ public class PhenotipsFamilyPermissions
         permissions.set(ALLOW, 1, context);
     }
 
-    private String[] getEntitiesWithEditAccessAsString(XWikiDocument patientDoc)
+    /**
+     * Returns all the users and groups that can edit the patient as array of two strings. First string in returned
+     * array contains all the users that can edit the patient, second string contains all the groups that can edit the
+     * patient.
+     *
+     * @param patientDoc patient to read permissions from
+     * @return array of rights
+     */
+    public String[] getEntitiesWithEditAccessAsString(XWikiDocument patientDoc)
     {
         String[] fullRights = new String[2];
         int i = 0;
--- a/components/family-studies/migrations/pom.xml
+++ b/components/family-studies/migrations/pom.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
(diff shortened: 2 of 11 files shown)


veronikaslc:
@allasm @meatcar please review

components/family-studies/migrations/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <!-- This would speed up build time, but we're playing with the singleton Utils.componentManager and this causes concurrency issues. -->
```
sdumitriu: This isn't true, you should remove this section.

components/family-studies/migrations/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+      <artifactId>patient-access-rules-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <!-- Test dependencies -->
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
```
sdumitriu: Before the final commit, you should check which dependencies are actually needed by running:

    mvn dependency:analyze

This will report which declared dependencies should be removed, and which transitive dependencies need to be declared.

sdumitriu:
You must also add this new module as a dependency in distribution/war and distribution/hsqldb-database.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyPermissions.java
```diff
@@ -121,7 +121,13 @@ public void setFamilyPermissionsToCurrentUser(XWikiDocument familyDoc)
         permissions.set(ALLOW, 1, context);
     }
 
-    private String[] getEntitiesWithEditAccessAsString(XWikiDocument patientDoc)
+    /**
+     * Gets a patient document owner permissions.
```
allasm: I would elaborate a bit more once this is public, since String[] does not give much information about what exactly is returned

components/family-studies/migrations/pom.xml
veronikaslc: Very useful, I was looking for that recently. Tried to run it on all phenotips. It looks like we have the same issue in all components. Probably will have to clean-up one day.

veronikaslc:
Updated.

veronikaslc:
Updated.

allasm:
Cherry-picked into PT-1548 branch

PR #1989, opened 2016-04-09
PT-2441: Candidate genes should not list an information button on patient view.

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
@@ -134,13 +134,13 @@
           element.up().down().remove();
       if (geneNames[i] != "") {
         element._name = geneNames[i];
-        var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
-        var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
-        var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
         if (isSuggestSelected) {
+          var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
+          var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
           element.insert({after: infoTool});
         }
         else {
+          var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
           element.up().insert({top: triangle});
         }
         ++i;
--- a/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
@@ -137,10 +137,12 @@
         var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
-        if (isSuggestSelected)
+        if (isSuggestSelected) {
           element.insert({after: infoTool});
-        else
+        }
+        else {
           element.up().insert({top: triangle});
+        }
         ++i;
       }
     });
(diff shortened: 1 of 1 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
```diff
@@ -137,10 +137,14 @@
         var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
-        if (isSuggestSelected)
-          element.insert({after: infoTool});
-        else
+        if (isSuggestSelected) {
+          if (!$('inline')) {
```
marta-: Just FYI, there's `XWiki.contextaction`  (which can be `view`, `edit`, `inline`, `admin`) available in the JS API. No need the modify the code though, this is also fine.

marta-: But shouldn't the condition be the other way around? If we're in edit mode, only then insert the info button?

veronikaslc: Sure, stupit typo :)

veronikaslc:
@itaiGershtansky @mjshepherd please, review

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
```diff
@@ -137,10 +137,14 @@
         var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
-        if (isSuggestSelected)
-          element.insert({after: infoTool});
-        else
+        if (isSuggestSelected) {
+          if ($('inline')) {
```
mjshepherd: this could be one if statement.

mjshepherd: Also the infoTool and triangle elements are not conditionally generated even though they are conditionally placed on the screen.

mjshepherd: Also it isn't clear from reading the code what ```if ($('inline'))``` is actually checking for. I guess its currently checking for the existence of an element which is only present in edit mode? But I can only take that guess by looking at the issue title.

veronikaslc: Indeed, that helped me to find much better fix.

veronikaslc:
@mjshepherd updated and rebased

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
```diff
@@ -137,10 +137,12 @@
         var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
```
mjshepherd: I'm still not a fan that triangle and infotool are generated regardless of whether they are used.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__genes.xml
```diff
@@ -137,10 +137,12 @@
...
-        if (isSuggestSelected)
+        if (isSuggestSelected) {
```
mjshepherd: I'm still not a fan that triangle and infotool are generated regardless of whether they are used.

veronikaslc: Agreed. This is exactly why `fetchGeneDetails()` function that generates them is now called only `if (XWiki.contextaction == "edit")`, [lines 171-173].

mjshepherd: yes thats good but its not what I mean. Also is the variable ```isSuggestSelected``` still appropriately named. Would something like ```isViewMode``` be better?

veronikaslc: The name pretty much reflects the context. If suggested term is selected then the info icon is inserted. If not - the triangle icon. It has nothing to do with view mode. I put brackets there just for code improvement, this is not the actual bugfix. The actual bugfix happens in lines 171-173 where I restrict the whole function that generates those icons to be implemented only in view mode.

veronikaslc:
@itaiGershtansky @mjshepherd : updared PR with codereview comments.

mjshepherd:
+1

marta-:
It was decided to not make this change, for usability reasons.

PR #1993, opened 2016-04-14
PT-2465: Anonymised pedigree should not show external ID, comments, names and DoB.

veronikaslc:
Fixed.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
@@ -811,12 +811,20 @@ define([
                     this.getAgeLabel().show();
                     labels.push(this.getAgeLabel());
                 }
+                if (this.getExternalIDLabel()) {
+                    this.getExternalIDLabel().show();
+                    labels.push(this.getExternalIDLabel());
+                }
+                if (this.getCommentsLabel()) {
+                    this.getCommentsLabel().show();
+                    labels.push(this.getCommentsLabel());
+                }
             } else {
                 this.getNameLabel() && this.getNameLabel().hide();
                 this.getAgeLabel() && this.getAgeLabel().hide();
+                this.getExternalIDLabel() && this.getExternalIDLabel().hide();
+                this.getCommentsLabel() && this.getCommentsLabel().hide();
             }
-            this.getExternalIDLabel() && labels.push(this.getExternalIDLabel());
-            this.getCommentsLabel() && labels.push(this.getCommentsLabel());
             var cancerLabels = this.getCancerAgeOfOnsetLabels();
             if (!Helpers.isObjectEmpty(cancerLabels)) {
                 for (var cancerName in cancerLabels) {


veronikaslc:
@allasm @mjshepherd please, review

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
```diff
@@ -811,12 +811,17 @@ define([
                     this.getAgeLabel().show();
                     labels.push(this.getAgeLabel());
                 }
+                this.getExternalIDLabel() && labels.push(this.getExternalIDLabel());
```
allasm: Don't we need to .show() and .hide() getExternalIDLabel() as well?

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
```diff
@@ -811,12 +811,17 @@ define([
...
+                if (this.getCommentsLabel()) {
+                    this.getCommentsLabel().show();
+                    labels.push(this.getCommentsLabel());
+                }
             } else {
                 this.getNameLabel() && this.getNameLabel().hide();
                 this.getAgeLabel() && this.getAgeLabel().hide();
+                this.getExternalIDLabel() && labels.push(this.getExternalIDLabel());
```
allasm: we don't want to do labels.push in this case

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
veronikaslc: Hm.. In JIRA issue external ID isn't mentioned. Can it be considered as sensitive info?

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
veronikaslc: But if we need to show external ID?
This line originaly was implemented regardless anonimization, thats why it is both if and else blocks to preserve an original order of lines in pdf.

allasm: AH, I see, I did not realize what happened. You are right, order is important in this case. But I would probably stop the if-else, do one line for externalID, do another if-else specifically for comments. Code duplication is IMHO worse and less intuitive and less maintainable than an extra if

veronikaslc: Did that already in previous PR, Sergiu did not like this way.

sdumitriu: There's always something to duplicate, either the `if`, or the `push`. I find it clearer if there's only one `if` statement.

An alternative is to move the external ID as the first label, before the `if`.

veronikaslc: Then questions to discuss: 
 1. Do we need to hide external id in anonimized pdf?
 2. Can we change the order of labels to make external id the first?

veronikaslc: In today's stand-up it was decided to merge this fix as it is, and create an issue for next milestone to refactor the pedigree anonimization to be fetched from configuration.

veronikaslc:
Updated PR commit.

marta-:
The whole anonymization implementation should be improved to have the PII fields all listed in one place (ideally configurable) and have them all treated in a generic way, instead of putting little pieces of code related to specific fields here and there. This is probably not a show stopper for this PR, but [a refactoring](https://phenotips.atlassian.net/browse/PT-2476) would drastically improve the maintainability of this code.

PR #1994, opened 2016-04-17
PT-2197: Unknown should be (default) option for Sex field.

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gender.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gender.xml
@@ -111,7 +111,7 @@
       <content>{{include reference="PhenoTips.PatientSheetMacros" /}}
 
 {{velocity}}
-#if ($xcontext.action == 'edit' &amp;&amp; $obj.getProperty('gender').value == '')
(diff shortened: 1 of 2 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gender.xml
```diff
@@ -111,6 +111,9 @@
       <content>{{include reference="PhenoTips.PatientSheetMacros" /}}
 
 {{velocity}}
+#if ($xcontext.action == 'edit' &amp;&amp; $obj.getProperty('gender').value == '')
```
sdumitriu: The second part of the condition only works after the patient record has been saved at least once. To make it more generic, use this:

    #if ($xcontext.action == 'edit' &amp;&amp; "$!obj.getProperty('gender').value" == '')

veronikaslc: Right, I see. Now it worked just because we first save an empty patient when clicking on "Create New", but since we are going to fix it soon (assigned to me :) ) this fix needs to be more generic.

PR #1997, opened 2016-04-19
Pt 2477 add get all patients

itaiGershtansky:


itaiGershtansky:
@meatcar, @veronikaslc, please review.

meatcar:
:+1:

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/PermissionsManager.java
```diff
@@ -41,4 +41,14 @@
...
     PatientAccess getPatientAccess(Patient targetPatient);
+
+    /**
+     * Receives a collection of patients and returns a new collection containing only those with
+     * visibility >= {@code requiredVisibility}.
+     *
+     * @param patients a collection of patients
+     * @param requiredVisibility minimum level of visibility required for patients
+     * @return a collection containing only those with visibility >= {@code requiredVisibility}
+     */
```
sdumitriu: Since these are new APIs, you should add a `@since 1.3M2` tag as well.

components/patient-data/api/src/main/java/org/phenotips/data/PatientRepository.java
```diff
@@ -74,4 +76,9 @@
      * @return the created patient record
      */
     Patient createNewPatient(DocumentReference creator);
+
+    /**
+     * @return a collection of all the patients.
```
sdumitriu: On a large instance, loading all the patients is likely to take a very long time, and consume plenty of memory. There are two optimizations that can be done:

1. Instead of returning a collection, return an iterator that will only load one patient at a time. Internally, this can  just keep a list of IDs.
2. The implementation should return a lazy collection that doesn't load all the needed data at once, only when it is actually requested.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatientRepository.java
```diff
@@ -204,4 +206,31 @@ private long getLastUsedId() throws QueryException
...
+    @Override
+    public Collection<Patient> getAllPatients()
+    {
+        List<String> patientIds = null;
+        try {
+            Query q = this.qm.createQuery(
+                "select doc.name "
+                + "from Document doc, "
+                + "doc.object(PhenoTips.PatientClass) as patient "
+                + "where patient.identifier is not null ", Query.XWQL);
```
sdumitriu: A sort would be nice.

sdumitriu: In PhenomeCentral we have a convention, if `patient.identifier < 0` that's not a real patient, but an OMIM template. Although that convention is temporary, at the moment you should also put that condition.

    where patient.identifier is not null

becomes

    where patient.identifier >= 0

components/patient-data/impl/src/main/java/org/phenotips/data/internal/SecurePatientRepository.java
```diff
@@ -107,8 +110,24 @@ public synchronized Patient createNewPatient()
...
+        Collection<Patient> patientsWithAccess = new LinkedList<Patient>();
+        for (Patient patient : patients) {
+            if (this.access.hasAccess(Right.VIEW, user, patient.getDocument())) {
+                patientsWithAccess.add(patient);
+            }
+        }
+        return patients;
+    }
+
+    private Patient checkAccess(Patient patient, DocumentReference user) {
```
sdumitriu: Our codestyle is to put the curly on the following line.

PR #1998, opened 2016-04-20
PT-1969: Backend: Migrator from the old measurement data structure. 

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-measurements/migrations/src/main/java/org/phenotips/measurements/internal/R71291PhenoTips1969DataMigration.java
+++ b/components/patient-measurements/migrations/src/main/java/org/phenotips/measurements/internal/R71291PhenoTips1969DataMigration.java
@@ -1,0 +1,270 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.measurements.internal;
+
+import org.phenotips.Constants;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.model.EntityType;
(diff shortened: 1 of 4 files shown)


veronikaslc:
@allasm @danielpgross please, review

danielpgross:
+1

danielpgross:
One thing to note is lack of javadoc, but I don't see this as a problem for a migrator.

components/patient-measurements/migrations/src/main/java/org/phenotips/measurements/internal/R71291PhenoTips1969DataMigration.java
```diff
@@ -0,0 +1,270 @@
...
+                migrateObject(date, age, fieldList, doc, context, newClassReference);
+                doc.removeXObject(object);
+            }
+            doc.setComment("Migrated MeasurementsClass data into MeasurementClass instances");
+            doc.setMinorEdit(true);
+            try {
+                // There's a bug in XWiki which prevents saving an object in the same session that it was loaded,
+                // so we must clear the session cache first.
+                session.clear();
+                ((XWikiHibernateStore) getStore()).saveXWikiDoc(doc, context, false);
```
allasm: Just curious do we always have to cast? Seems like saveXWikiDoc() is the one called every time, why dont we return appropriate store from getStore() in the first place? (not commenting on this commit, just a general question)?

components/patient-measurements/migrations/src/main/java/org/phenotips/measurements/internal/R71291PhenoTips1969DataMigration.java
```diff
@@ -0,0 +1,270 @@
...
+            Date birth = null;
+            if (patientObject != null) {
+                birth = patientObject.getDateValue("date_of_birth");
+            }
+
+            for (BaseObject object : oldObjects) {
+                Date date = object.getDateValue(DATE_FIELD);
+                String age = calculateAge(birth, date);
+                Collection fieldList = object.getFieldList();
+                migrateObject(date, age, fieldList, doc, context, newClassReference);
```
allasm: This is fine for this migration, but curious is there a better way to do migrations, without a function with side-effects? E.g. here a BaseObjects/XObject is created within the function, and is attached to the doc on creation. Seems like a better way is to return created detached XObject and attach it to the doc here, so that it is clear from reading only the high-level code what gets attached

allasm:
+1 (but I did not build/test it - should I?)

components/patient-measurements/migrations/src/main/java/org/phenotips/measurements/internal/R71291PhenoTips1969DataMigration.java
sdumitriu: That's the XWiki API, we can't change it in our codebase.

veronikaslc:
It may be a good idea to build and test migrator, I think. Could reveal some problems.

PR #1999, opened 2016-04-21
PT-2452: Apgar scores and twin number doesn't get pushed.

veronikaslc:
Push twinNumber.

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
@@ -127,7 +127,9 @@ public abstract class AbstractComplexController<T> implements PatientDataControl
             Map.Entry<String, T> item = iterator.next();
             String itemKey = item.getKey();
             Object formattedValue = format(itemKey, item.getValue());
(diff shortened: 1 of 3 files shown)


veronikaslc:
@allasm @itaiGershtansky please, review

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -127,7 +127,9 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
             Map.Entry<String, T> item = iterator.next();
             String itemKey = item.getKey();
             Object formattedValue = format(itemKey, item.getValue());
-            if (formattedValue != null && (selectedFieldNames == null || selectedFieldNames.contains(item.getKey()))) {
+            if (formattedValue != null
+                && (selectedFieldNames == null
+                || selectedFieldNames.contains(getControllingFieldName(item.getKey())))) {
```
allasm: I would put this on the previous line since || is not on the same level as &&, but is not worth fixing if that is the only change, just a comment

allasm:
+1

itaiGershtansky:
+1

veronikaslc:
Code is merged to push-1.3m2 branch.

PR #2001, opened 2016-04-25
PT-2438 Provide more descriptive warning to users wishing to export ...

josecv:
...to BOADICEA

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/model/export.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/model/export.js
@@ -337,6 +337,7 @@ define([
          // TODO: Contralateral breast cancer export/field?
          var cancerSequence = [ "Breast", "", "Ovarian", "Prostate", "Pancreatic" ];
 
+         var is_affected = false;
          for (var c = 0; c < cancerSequence.length; c++) {
              cancer = cancerSequence[c];
              if (cancer == "" || !pedigree.GG.properties[i].hasOwnProperty("cancers")) {
@@ -348,20 +349,19 @@ define([
                  var cancerData = pedigree.GG.properties[i].cancers[cancer];
                  if (!cancerData.affected) {
                      output += "0\t";
-                     if (yob == "0") {
-                        warnMissingDOBUnaff = true;
-                     }
                  } else {
                      var ageAtDetection = cancerData.hasOwnProperty("numericAgeAtDiagnosis") && (cancerData.numericAgeAtDiagnosis > 0)
                                           ? cancerData.numericAgeAtDiagnosis : "AU";
                      output += ageAtDetection.toString() + "\t";
-                     if (yob == "0" || ageAtDetection == "AU") {
-                         warnAboutMissingDOB = true;
-                     }
+                     is_affected = true;
                  }
              } else {
                  output += "0\t";
              }
+         }
+         if (yob == "0") {
+            warnAboutMissingDOB = warnAboutMissingDOB || is_affected;
+            warnMissingDOBUnaff = warnMissingDOBUnaff || !is_affected;
          }
 
          output += "0\t"; // TODO: Genetic test status
@@ -403,7 +403,7 @@ define([
          output += "\n";
      }
 
-     if (alertUnknownGenderFound || warnAboutMissingDOB) {
+     if (alertUnknownGenderFound || warnAboutMissingDOB || warnMissingDOBUnaff) {
          var warningText = "Pedigree can be exported, but there are warnings:\n\n\n";
          var warnings = [];
          if (alertUnknownGenderFound) {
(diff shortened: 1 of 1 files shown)


components/pedigree/resources/src/main/resources/pedigree/model/export.js
```diff
@@ -347,11 +348,14 @@ define([
...
                      output += "0\t";
+                     if (yob == "0") {
+                        warnMissingDOBUnaff = true;
+                     }
                  } else {
                      var ageAtDetection = cancerData.hasOwnProperty("numericAgeAtDiagnosis") && (cancerData.numericAgeAtDiagnosis > 0)
                                           ? cancerData.numericAgeAtDiagnosis : "AU";
                      output += ageAtDetection.toString() + "\t";
-                     if (yob == "0") {
+                     if (yob == "0" || ageAtDetection == "AU") {
```
allasm: I think ageAtDetection == "AU" is fine, and is the way to specify cancer for those whose yob is unknown. Moreover, probably it should be ("yob == "0" && ageAtDetection != "AU"), e.g. I think BOADICEA only needs yob when age is given. But need to just test it with real BOADICEA to know for sure

sdumitriu:
@josecv You're supposed to nominate two other developers to review your pull request. I'll do that for you: @allasm and @itaiGershtansky, please review this.

josecv:
@sdumitriu ok I see, thanks!

components/pedigree/resources/src/main/resources/pedigree/model/export.js
josecv: Hmm I see, is there a way I can do a test import to see what BOADICEA will be happy with?

components/pedigree/resources/src/main/resources/pedigree/model/export.js
```diff
@@ -401,22 +405,26 @@ define([
...
+                            "birth and age at cancer diagnosis specified or estimated\n\n" +
+                            "A person with cancer is missing age data, data will be exported but may not be accepted by BOADICEA");
+         }
+         if(warnMissingDOBUnaff) {
+             warnings.push("BOADICEA recommends that all unaffected individuals have their year of birth and" +
+                           " year of death, if applicable, specified or estimated." +
+                           " Not doing so may lead to an overestimation of risk.");
+         }
+         if(warnings.length > 1) {
+             warnings = warnings.map(function(v, i, a) { return (i + 1) + ") " + v; });
```
danielpgross: This is a nice pattern, but keep in mind that [`Array.prototype.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) is only available in IE9+. I think @allasm was using a helper for `Array.prototype.forEach()`, so that might be a consideration here as well.

danielpgross: Nevermind, [`map()`](http://prototypejs.org/doc/latest/language/Enumerable/prototype/map/) is provided by PrototypeJS. Given that, I'm not sure why a separate helper is defined for it in the [pedigree helpers](https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/model/helpers.js#L26).

sdumitriu: Andriy is defining his own helper because the pedigree shouldn't depend on Prototype.js.

As for IE<9, we don't really support those. At least the pedigree is completely read-only, so there's no need to make it work in IE8.

allasm: Yeah, one of the goals was to be XWiki/prototype independent, and to support all kinds of browsers. We also wanted to support IE8 a few years ago when it started, but by now it is no longer a goal.

(I think in IE8 mode the export button is still available - the idea is, if you happen to be in an institution which only has IE8, you should still be able to get a local copy for local processing - but I guess if you want to do real work (e.g. export to BOADICEA), IE8 wont work. Then again, in practice, prototype is there when pedigree is used in phenotips)

components/pedigree/resources/src/main/resources/pedigree/model/export.js
josecv: Ah, right you are. After some testing, BOADICEA will accept "AU" (even if a yob is given) and reject whenever a yob isn't given for affected individuals. Will also warn if yob is not given for unaffected individuals. Will change the warnings to mirror that.

allasm:
Is last commit (https://github.com/phenotips/phenotips/pull/2001/commits/bcf6c0b0ddc3df0a1fd653dd80063747077b5273) related to PT-2438?

josecv:
No it isn't.

Looks like i pushed to master by accident. Lemme just fix that

allasm:
+1 other than the last comment

josecv:
Done, that commit's gone now

allasm:
then this is ready to be merged

itaiGershtansky:
Don't know how I missed that. +1 and sorry for the late reply

PR #2006, opened 2016-05-04
PT-2459: Selection of Clinical Symptoms and physical findings triggers pushing of Prenatal and perinatal phenotypes.

veronikaslc:
Fixed.


allasm:
+1

P.S. I think we should add all possible phenotype types to Feature class, so that PRENATAL_PHENOTYPE_TYPE is defined in Feature, and not in a random file outside of Feature. But that is a different change and this commit fixed the issue at hanbd

veronikaslc:
Merged to push-1.3m2 branch.

PR #2007, opened 2016-05-09
PT-2509: Push dialog improvements.

veronikaslc:
Fixed.

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
@@ -188,9 +188,46 @@ document.observe('xwiki:dom:loading', function() {
     }
     
     //============================================================================
-    // Column selection
-    var columnList = content.down('.section.columns ul');
-    if (columnList) {
+    // Push dialog check boxes and expand tools
+    var titles = content.select('.push-fields.section.columns > h5');
+      if (titles) {
+        titles.each(function(item, index){
+        var sectionList = item.next('ul');
+        var showIcon = '<span class="fa fa-plus-square-o fa-lg"></span>';
+        var chapterShow = new Element('button', {'class' : 'tool button secondary', 'type' : 'button'}).update(showIcon+" "+"$services.localization.render('phenotips.patientSheet.expandSection')");
+        var hideIcon = '<span class="fa fa-minus-square-o fa-lg"></span>';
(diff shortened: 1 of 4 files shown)


components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
```diff
@@ -188,35 +188,73 @@ document.observe('xwiki:dom:loading', function() {
...
+        var chapterShow = new Element('button', {'class' : 'tool button secondary', 'type' : 'button'}).update(showIcon+" "+"$services.localization.render('phenotips.patientSheet.expandSection')");
+        var hideIcon = '<span class="fa fa-minus-square-o fa-lg"></span>';
+        var chapterHide = new Element('button', {'class' : 'tool button secondary', 'type' : 'button'}).update(hideIcon+" "+"$services.localization.render('phenotips.patientSheet.collapseSection')");
+        var chapterShowWrapper = new Element('span', {'class' : 'buttonwrapper show'}).insert(chapterShow);
+        var chapterHideWrapper = new Element('span', {'class' : 'buttonwrapper hide'}).insert(chapterHide);
+        var chapterExpandTools = new Element('span', {'class' : 'expand-tools'}).insert(chapterShowWrapper).insert(chapterHideWrapper);
+        item.insert({after: chapterExpandTools});
+        // section expand tools behaviour
+        [chapterShow, chapterHide, item].invoke('observe', 'click', function (event) {
+          if (event.path[0].tagName != "INPUT") {
```
marta-: event.path is not implemented in firefox. To find to original element, you could use prototype's event.findElement() instead of event.path[0].

marta-: @veronikaslc ^

allasm: we found this way while hacking around, but I think a better way would be to place the text in the label within another element (e.g. span) and on;l process event for the span, not for the label tag

E.g. right now we have
  `<label>  <input></input> text </label>`, and label handles clicks, and has to distinguish input vs text

while we can do
 `<label>  <input></input> <span>text</span> </label>` and make span handle expansion clicks unconditionally

sdumitriu: Fixed in bdf979fba331b061673c286585f33c56ede36b79

veronikaslc: was going to commit a fix myself, but actually took time to test in all browsers, and its not working fully properly in IE11, as well as what @allasm proposed.

PR #2008, opened 2016-05-10
PT-2544: Labels for checkbox fields in the genes and variants section target the first gene/variant added.

veronikaslc:
Fixed.

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
@@ -13,8 +13,8 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
         if ($('inline')) {
           // getting rid of 'for' attributes in labels and 'id' in inputs
           $$('.gene-table label.xwiki-form-listclass').each (function (label) {
-            label.setAttribute("for", "");
-            label.down('input').id = "";
+        	label.removeAttribute("for");
+        	label.down('input').removeAttribute("id");
           });
             
           this.warnSaving = false;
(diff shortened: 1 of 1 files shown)


components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -149,12 +149,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       var buttonRowInner = this.addVariantButtonTemplate.evaluate(data);
       var geneRowInner = this.geneRowTemplate.evaluate(data);
 
-      geneRowInner = geneRowInner.replace(/_\d+_|=\d+&/g, function (placeholder) {
+      geneRowInner = geneRowInner.replace(/_\d+_|=\d+&|xwiki-form-strategy-\d+/g, function (placeholder) {
```
marta-: This works, however I have 2 comments:
- while I love regular expressions, I'm not happy with using them for manipulating html fragments because they can so easily produce invalid results;
- the property names are hard coded; if we add more fields displayed as checkboxes (or radio buttons), we need to add even more javascript code to fix those.

I wonder if this can be done in a safer way (using dom manipulation instead of regex, e.g. _after_ the fragment is inserted in the document) and using a more generic approach that would capture all cases (for instance, they're all inputs in labels with the css class "xwiki-form-listclass"). It would suffice to remove the label's for attribute and the input's id attribute, no need to actually replace the values. 
@veronikaslc Thoughts?

veronikaslc:
@allasm @itaiGershtansky please, review

veronikaslc:
@allasm @itaiGershtansky please, review

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -9,6 +9,12 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
 
       //if edit mode
       if ($('inline')) {
+        // getting rid of 'for' attributes in labels and 'id' in inputs
+        $$('.gene-table label.xwiki-form-listclass').each (function (label) {
+          label.setAttribute("for", "");
```
itaiGershtansky: Would removeAttribute() be better?

marta-: Right, that's better.

veronikaslc:
Updated.

PR #2011, opened 2016-05-10
PT-2470: Hide empty list of genes from export pdf and patient form when in view mode.

veronikaslc:
Fixed.


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetMacros.xml
```diff
@@ -257,6 +257,5 @@
     )))
   #end
 #end
-{{/velocity}}
-</content>
+{{/velocity}}</content>
```
marta-: I would have put white space changes in a different commit, since they're not  directly related to the issue.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
```diff
@@ -112,14 +112,19 @@
 
 {{include reference="PhenoTips.TabelarDataMacros" /}}
 
-{{include reference="PhenoTips.GeneVariantMacros" /}}
```
marta-: Same for file reorganization

marta-:
All looks good otherwise.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -3,35 +3,37 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
   tools.Editor = Class.create({
 
     initialize : function () {
-      var geneTableListId = $$('.gene-table.extradata-list')[0].id;
-      this.geneClassName = geneTableListId.substring(geneTableListId.lastIndexOf('-') + 1);
-      this.geneVariantClassName = 'PhenoTips.GeneVariantClass';
```
sdumitriu: This line was removed, and now `geneVariantClassName` is no longer defined anywhere, although it is still used in several places.

PR #2018, opened 2016-05-18
Fixes for PT-1933, PT-2445, PT-2452, PT-2457, PT-2484, PT-2530, PT-2561, PT-2562, PT-2568

allasm:
And a few other small push dialogue UI fixes


components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -210,13 +212,19 @@
   #set($selectedValues = [])
 #end
 
-{{html clean="false" filter="indent"}}&lt;form class="xform" action="${escapetool.xml($request.get('export_endpoint'))}" method="post"&gt;&lt;div&gt;{{/html}}
+#set ($form_action = $!{escapetool.xml($!{request.get('export_endpoint')})})
+#if (!$form_action)
+  #set ($form_action = "none")
```
sdumitriu: This will actually send the form to `http://localhost:8080/none`, probably not what we want.

PR #2019, opened 2016-05-18
PT-2100: Warn user when he tries to save gene/variant with empty key field.

veronikaslc:
Fixed.

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
@@ -2,7 +2,7 @@ var XWiki = (function(XWiki) {
   // Start XWiki augmentation
   var widgets = XWiki.widgets = XWiki.widgets || {};
 
-  widgets.GeneValidator = Class.create({
+  widgets.GeneVariantValidator = Class.create({
     initialize : function(input) {
       this.input = input;
       this.valid = true;
@@ -10,7 +10,7 @@ var XWiki = (function(XWiki) {
       this.value = input.value;
       if (!this.input.__validation) {
         try {
-          this.input.__validation = new LiveValidation(this.input, {validMessage: '', wait : 500});
+          this.input.__validation = new LiveValidation(this.input, {validMessage: '', wait : 500, displayMessageWhenEmpty: true});
         }
         catch(err) {
           //console.log(err);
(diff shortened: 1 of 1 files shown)


marta-:
@AlexLapena please review.

allasm:
Should we include the empty-gene-remover listener in this as well?

veronikaslc:
We may, however, it's kind of different issue..

AlexLapena:
+1 @marta- @veronikaslc

sdumitriu:
Why not use the LiveValidation widget?

veronikaslc:
Good point, will re-implement to utilize LiveValidation.

veronikaslc:
Commit amended with re-implementation. @AlexLapena please, re-review.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
```diff
@@ -61,9 +66,14 @@ var XWiki = (function(XWiki) {
...
   var init = function(event) {
     ((event && event.memo.elements) || [$('body')]).each(function(element) {
-      element.select('input.gene-name').each(function(input) {
+      element.select('[name^="PhenoTips.GeneClass_"][name$="_gene"]').each(function(input) {
         if (!input.__Gene_validator) {
-          input.__Gene_validator = new XWiki.widgets.GeneValidator(input);
+          input.__Gene_validator = new XWiki.widgets.GeneVariantValidator(input);
+        }
+      });
+	  element.select('[name^="PhenoTips.GeneVariantClass_"][name$="_cdna"]').each(function(input) {
```
sdumitriu: Wrong indentation.

sdumitriu: Since the code to execute is almost identical in the two selectors, why don't you combine them? Like this: `element.select('selector 1', 'selector 2').each...`

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
```diff
@@ -72,21 +82,6 @@ var XWiki = (function(XWiki) {
 
   (XWiki.domIsLoaded && init()) || document.observe("xwiki:dom:loaded", init);
   document.observe('xwiki:dom:updated', init);
-  document.observe('ms:suggest:selected', function(event) {
```
sdumitriu: This one isn't needed anymore?

AlexLapena:
+1 once indentation issue at line 74 is fixed.

AlexLapena:
emptykeyfield

However this looks a little sloppy.

AlexLapena:
Once a variant is deleted and the warning message shows that "Input is blank", when a user tries adding in a new variant the message "This gene has already been entered" displays. There is no way to save the variant unless it is deleted and recreated. 

varientwarning

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
```diff
@@ -20,28 +20,33 @@ var XWiki = (function(XWiki) {
...
       }
     },
     validate : function(value) {
-      if (this.state == 'DONE' &&
-          this.value == value &&
-          !this.valid) {
-        Validate.fail("This gene has already been entered.");
+      if ((this.state == 'DONE' && this.value == value && !this.valid) ||
+          (this.state == 'NEW' && value.blank())) {
+        var message = (value.blank()) ? "Input is blank." : "This gene has already been entered.";
```
sdumitriu: `Input is blank` sounds unfriendly. `Please enter a value` is better, but `Please enter a [gene name|variant]` would be best.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/geneValidation.js
sdumitriu: Wouldn't a simple `Validate.Presence` be enough for variants? It seems that most of the code from `GeneVariantValidator` isn't needed, and mixing variants in there adds unneeded complexity.

PR #2025, opened 2016-05-26
PT-1618: Quick Phenotype search section overlaps the topmost item

Ruxuan:
@marta- @danielpgross @allasm Please review. Thanks!

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__phenotype.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__phenotype.xml
@@ -138,7 +138,7 @@
   var phBrowserCollapseAllWrapper = new Element('span', {'class' : 'buttonwrapper collapse-all'}).insert(phBrowserCollapseAll);
   var phBrowserExpandTools = new Element('span', {'class' : 'inner-expand-tools'}).insert(phBrowserExpandAllWrapper).insert(phBrowserCollapseAllWrapper);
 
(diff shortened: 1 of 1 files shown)


danielpgross:
Code is approved.
As for UI testing, I have tested in OSX10.11/Chrome50 with different window sizes and it seems to be working perfectly. This should be tested on IE9 before merging; please test (you can use the VM that Teyden used) and follow up here with the results. Thanks.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__phenotype.xml
```diff
@@ -482,6 +484,10 @@
   margin: .5em .75% 0em !important;
   padding-right: .75%;
 }
+.non-moving {
+  margin-top: 75px;
```
sdumitriu: Most of the UI is defined in `em` units, not in pixels. The actual height of the QS box is exactly 4em, plus 1.2 for the margin of the "Browse categories" header, set this margin to `5.2em` instead of `75px`. This way it will actually fit even if the brower uses larger fonts or a different DPI.

danielpgross:
I reviewed the IE9 rendering with Ru and it is bad -- the "Quick phenotype search" box becomes overlapped by the left-side list of phenotypes and therefore hardly usable. However, the behaviour seems to be the same in master currently, i.e. regardless of the changes in this PR. Therefore, I think the IE9 issues are separate from this issue, and if necessary, addressed in a separate thread.
With that covered, this PR is approved from my side 👍

PR #2027, opened 2016-05-27
PT-2161: Birth/death date "hidden" input fields are visible for a short time after opening patient edit page

allasm:
Fixed


sdumitriu:
That behavior was so that the patient record can still be edited without scripts, but I guess that's not a goal worth pursuing.

Now instead of seeing an input replaced by three dropdowns, the dropdowns pop out of nowhere, but I guess that's OK.

PR #2028, opened 2016-05-27
PT-2585: Make patient page more visually static and less annoying on load

allasm:
Fixed. Added a "expanded_by_default" parameter to section configuration to make
configuring which sections are expanded by default easier. Also made the
expand-collapse JS extension load first, which visually speeds things up.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -40,6 +40,130 @@
   <content/>
   <object>
     <name>PhenoTips.PatientSheetCode</name>
+    <number>1</number>
+    <className>XWiki.JavaScriptExtension</className>
+    <guid>fee0fd2a-16d1-45ac-8a44-16af9afec0f7</guid>
+    <class>
+      <name>XWiki.JavaScriptExtension</name>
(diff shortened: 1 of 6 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -2222,7 +2222,7 @@ document.observe('xwiki:dom:loaded', function()
   </object>
   <object>
     <name>PhenoTips.PatientSheetCode</name>
-    <number>8</number>
+    <number>1</number>
```
sdumitriu: Is this change right?

sdumitriu: I guess you're trying to make this one execute sooner. The whole object should be moved up in the source then.

allasm: I do and it seems to work. However moving the whole object up should be done as well, I somehow did not think of that (since the slowest code is in this object anyhow). For the record, all stuff related to UI elements visible 90% of the time on the first page should be moved up, like the date-picker widget (so that date input fields don't pop up out of nowhere(or at least do it faster),like you commented for another commit)

PR #2030, opened 2016-05-30
PT-2565, PT-2566 fixed.

williehwc:
(PT-2583, the name of this branch, is closely associated with PT-2566 and was fixed by Marta. The PT-2566 fix extends the PT-2583 fix to paragraphs and divs)

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatientConsentManager.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatientConsentManager.java
@@ -154,13 +154,13 @@ public class PhenoTipsPatientConsentManager implements ConsentManager, Initializ
         try {
             String id = xwikiConsent.getStringValue("id");
             String label = cleanDescription(
-                    configDoc.display("label", RENDERING_MODE, xwikiConsent, contextProvider.get()));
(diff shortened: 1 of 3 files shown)


williehwc:
@itaiGershtansky please review

itaiGershtansky:
Finer commit granularity (for next time). Other than that, +1.

marta-:
Please make separate PRs for issues with independent fixes.

itaiGershtansky:
The consent texts appear on a line below the checkbox, because of the <div>-s that are not removed. They cannot all be removed, but those who cause that line break should.

williehwc:
@itaiGershtansky The consent identifier (label) are one-liners that should not have divs, but they are rendered with divs using XWikiDocument.display. The divs cause the consent identifiers to appear below the checkbox, so they should always be removed. The consent descriptions have divs, but they are kept because they should appear below the checkbox as expected.

williehwc:
To-do: Remove div tags from consent labels, since PhenomeCentral (not PhenoTips, apparently) inserts div tags in consent labels.

allasm:
Maybe the difference between PhenomeCentral and PhenoTips is that in PT you manually entered the label, while in PC we use translations, which are rendered differently?

williehwc:
This pull request will be closed; two new pull requests will be made that separate the fixes.

PR #2031, opened 2016-05-31
PT-924: The "quick fix" for displaying hand length percentile when taken < 2 years old.

williehwc:
From JIRA: "The quick fix is to display nothing if we get -1 on evaluation. The better fix is to get the data for 0-2y."


allasm:
+1

PR #2032, opened 2016-05-31
PT-2565: Save and refresh patient data when consent checkbox is ticked/unticked

williehwc:
Fixed

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
@@ -184,6 +184,13 @@
 
       // attach button action
       var onUpdateButtonPress = function(event) {
+        this.screen = new Element('div', {'class': 'msdialog-screen consent-grayout'});
+        consentsBlock[0].insert(this.screen);
+
+        var removeConsentGrayout = function(){
+          delete this.screen;
+        }
+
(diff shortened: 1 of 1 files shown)


allasm:
For the record: as tested together with @williehwc, there is still a problem: if a user clicks consent check-boxes quickly various kinds of errors may pop up since while one of the rest and/or save requests is being processed further rest requests and save operations may fails. This may result in the lost check-box state or an error on save and a failure to refresh the page.

One possible solution is to disable all inputs as soon as one check-box is clicked.

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
```diff
@@ -184,6 +184,13 @@
 
       // attach button action
       var onUpdateButtonPress = function(event) {
+        this.screen = new Element('div', {'class': 'msdialog-screen consent-grayout'});
+        consentsBlock[0].insert(this.screen);
+
+        var removeConsentGrayout = function(){
+          delete this.screen;
```
sdumitriu: This won't actually remove the element from the HTML, will it?

marta-:
This solution doesn't account for form fields that are absent until specific consents are checked and the page is reloaded.

An idea that would probably work better is to modify the patient sheet to render all enabled fields and hide (client side) those dependent upon checking a specific consent, then show them if the consent is checked.

PR #2034, opened 2016-06-03
PT-2366: Improve diagnosis and "other phenotypes" suggestions by taking into account global mode of inheritance and global age of onset

williehwc:
Fixed


marta-:
Code adapted and committed in 16d651e and e251225.

PR #2038, opened 2016-06-09
PT-2602: 1.3 code should also support the 1.2 JSON structure, where the rejected and candidate genes are in a different place.

veronikaslc:
Fixed.


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -246,45 +258,96 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
 
         try {
             JSONArray genesJson = json.getJSONArray(this.getJsonPropertyName());
```
sdumitriu: This may be missing from the 1.2 JSON, so it should also be retrieved with `optJSONArray`, and `parseGenesJson` should start with an `if not null`.

PR #2040, opened 2016-06-13
PT-2563: Pushing a patient with empty data.

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
@@ -355,7 +355,7 @@ public class PhenoTipsPatient implements Patient
             result.put(JSON_KEY_NON_STANDARD_FEATURES, nonStandardFeaturesToJSON(selectedFields));
         }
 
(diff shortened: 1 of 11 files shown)


veronikaslc:
@allasm @AlexLapena please, revew/test

AlexLapena:
+1

allasm:
One thing still needs to be done - in writeJSON() in most of the controllers there is a piece of code like this:

```
PatientData<T> data = patient.getData(getName());
if (data == null) {
   return;
}
```

which results in no data place in the produced JSON if the patient has no data of that type. As f ar as I understand it should be the other way around, if `selectedFieldNames` explicitly requested that data to be there, we should put a field in the JSON, but have the data blank.

This will result in large JSON with lots of blank fields (by default - all the supported fields), but will guarantee data will be reset back to empty.

@marta- , do you confirm this?

One spin on this may be that if we just do a regular export (not push with explicitly selects fields), `selectedFieldNames` will be `null`, and we only add fields that have some data. If on the other hand a field was explicitly requested via a non-null `selectedFieldNames` it should necessarily be added, with possibly blank value. I like this option more to keep basic JSONs smaller - after all, empty data is only necessary to overwrite existing data (via push or explicit rest api overwrite call), but not necessary when e.g. exporting into a new instance

PR #2043, opened 2016-06-16
Add DurationField widget

danielpgross:
This widget was already added to [`feature-measurements`](/phenotips/phenotips/tree/feature-measurements), as `DurationValidator`. I needed to make a few important changes for [`phenotype-meta-exact-onset`](/phenotips/phenotips/tree/phenotype-meta-exact-onset), so I created a separate branch for it. I have already merged it into [`feature-measurements`](/phenotips/phenotips/tree/feature-measurements) and will fix that branch as necessary.


danielpgross:
Actually, never mind... this PR shouldn't be necessary, since the changes should be merged in with `phenotype-meta-exact-onset` when it is merged into `master`.

PR #2044, opened 2016-06-17
Phenotype metadata -- exact age and date of onset

danielpgross:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4584,14 +4588,96 @@ document.observe("xwiki:dom:loaded", function() {
             this.__input.fire("xwiki:date:changed");
           }
         }
+    },
+
+    getFormattedValue: function() {
+      if (!this.dateParts) {
+        return undefined;
+      }
+      var format = "yyyy-MM-dd";
```
danielpgross: TODO: use the configured format.

danielpgross:
Nominating @veronikaslc and @itaiGershtansky for review

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/ageOfOnset.js
```diff
@@ -0,0 +1,100 @@
+var PhenoTips = (function(PhenoTips) {
+  // Start PhenoTips augmentation
+  var widgets = PhenoTips.widgets = PhenoTips.widgets || {};
+
+  widgets.AgeOfOnset = Class.create({
+    // The age until which the specified HPO term applies
+    ageMonthsToHpoTerm: [
+      [0.92008, "HP:0003623"],
```
itaiGershtansky: Why 0.92008?

danielpgross: 28 days == 0.92008 months. Maybe this should be expressed more clearly by including the math inline, or using an explanation in a comment.

itaiGershtansky: What's the calculation? Shouldn't it be 28/(365/12) = 0.92054?

allasm: BTW, this math gets even more ugly if you consider 366 days in leap year...

danielpgross: DurationField uses 0.03286 months/day ([source](https://github.com/phenotips/phenotips/blob/d4d624b3d280fada8e06a2e58a04a7d408f35cc0/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/durationField.js#L68)), so I just multiplied this by 28. The calculation would actually use 365.25 (average days/year considering leap years), which would give 365.25/12=30.4375, 28/30.4375 = 0.91992.

There is some precision lost in the DurationField's months/day and as a result some imprecision in "28 days == 0.92008 months" but it makes sense to match the DurationField's value here rather than the more precise value.

danielpgross:
ping @veronikaslc

veronikaslc:
+1, other than a super minor note that I'd disable month and day dropdowns if user selects year range like '2010s', we don't save them but allow to select.

danielpgross:
Thanks @veronikaslc, I agree about that. This would be a separate issue to file for FuzzyDatePicker.
Ping @itaiGershtansky

itaiGershtansky:
+1

danielpgross:
Thanks @itaiGershtansky.
@sdumitriu, this is ready for final review.

danielpgross:
@sdumitriu is there anything blocking the merge?

danielpgross:
I just rebased on top of 1.3-milestone-3.

marta-:
@danielpgross why m3 and not master?

danielpgross:
@marta- because I'm extracting it out into a module for NIAID, and their instance is currently based on m3. There is no reason that it can't be rebased on top of master though.

danielpgross:
Thanks @marta-, I will investigate.

danielpgross:
@marta-, the 3 issues you list above have been fixed in https://github.com/Gene42/phenotips-niaid.
(I just fixed the second bug with https://github.com/Gene42/phenotips-niaid/commit/7197701029d9403cd78b73d8851b4113c246e215).

marta-:
@danielpgross Great, can you update the PR?

sashaandjic:
- I think having Date + Age + Predefined categories to be a bit too much for a user to comprehend
- it is not obvious that once either date or age fields are edited, the predefined age categories turn read only as the colour of the font is grey 
- i don't like the formatting in the age field (it is logical, but users can be turned off once they get "Invalid timespan" warning)
- wouldn't just one simple "As of age" (as in the pedigree editor) alone work? once populated it can be determined if it is juvenile, late without displaying the category selections at all

sdumitriu:
Remaining issues:
- data is not exported/imported through JSON
- when validation fails, the error is not reported, so for a user it looks like the Save button simply disables itself when being pressed
- if an invalid date/age is entered, then the phenotype or its details are deleted, the validation will still be triggered on the now removed fields, making the form unsaveable; the validation should be removed when the element itself is removed

PR #2046, opened 2016-06-22
Reworded heading name to reduce admin confusion

katpavlova:
Discussed with Courtney. Measurements info box states that "It is recommended to record weight, height and head circumference for all patients." Since this is a static description that does not change when items are set as "Recommended" by the admin, we decided that that the administration field name should be changed from "Recommended" to "Fixed".


allasm:
Can it confuse someone since "fixed" may mean both "fixed as in `unmodifiable` e.g. `listg of items that can't be modified`" and "fixed as in `not broken`, e.g. `the list was wrong but below is the fixed modified list`"

katpavlova:
Considering the context, both Courtney and I thought that it was the clearest wording. The two options are "Fixed" and "Expanded"; it's pretty straight-forward, especially if you actually look at the Measurements field in the patient case.

sdumitriu:
The message "It is recommended to record weight, height and head circumference for all patients" should actually be removed, since the recommended measurements will be the ones displayed by default.

So:
- don't do this change
- remove that message instead

PR #2047, opened 2016-06-28
PT-2563: Pushing a patient with empty data.

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
@@ -355,7 +355,7 @@ public class PhenoTipsPatient implements Patient
             result.put(JSON_KEY_NON_STANDARD_FEATURES, nonStandardFeaturesToJSON(selectedFields));
         }
 
-        if (!this.disorders.isEmpty() && isFieldIncluded(selectedFields, DISORDER_PROPERTIES)) {
+        if (isFieldIncluded(selectedFields, DISORDER_PROPERTIES)) {
             result.put(JSON_KEY_DISORDERS, diseasesToJSON());
         }
 
@@ -507,6 +507,7 @@ public class PhenoTipsPatient implements Patient
                 // as in constructor: make unmofidiable
                 this.disorders = Collections.unmodifiableSet(this.disorders);
 
+                data.set(DISORDER_PROPERTIES_OMIMID, null, context);
                 // update the values in the document (overwriting the old list, if any)
                 data.set(DISORDER_PROPERTIES_OMIMID, disorderValues, context);
                 context.getWiki().saveDocument(doc, "Updated disorders from JSON", true, context);
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
@@ -377,6 +377,9 @@ public class PhenoTipsPatient implements Patient
         JSONObject json)
     {
         try {
+            if (!json.has(JSON_KEY_FEATURES) && !json.has(JSON_KEY_NON_STANDARD_FEATURES)) {
+                return;
+            }
             JSONArray jsonFeatures =
                 joinArrays(json.optJSONArray(JSON_KEY_FEATURES), json.optJSONArray(JSON_KEY_NON_STANDARD_FEATURES));
 
@@ -491,6 +494,9 @@ public class PhenoTipsPatient implements Patient
     private void updateDisordersFromJSON(XWikiDocument doc, BaseObject data, XWikiContext context, JSONObject json)
     {
         try {
+            if (!json.has(JSON_KEY_DISORDERS)) {
+                return;
+            }
             JSONArray inputDisorders = json.optJSONArray(JSON_KEY_DISORDERS);
             if (inputDisorders != null) {
                 // keep this instance of PhenotipsPatient in sync with the document: reset disorders
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
@@ -84,10 +84,17 @@ public class PhenoTipsPatient implements Patient
 
     private static final String NEGATIVE_PHENOTYPE_PREFIX = "negative_";
 
+    private static final String PRENATAL_PHENOTYPE_PREFIX = "prenatal_";
+
+    private static final String PRENATAL_PHENOTYPE_PROPERTY = PRENATAL_PHENOTYPE_PREFIX + PHENOTYPE_POSITIVE_PROPERTY;
+
     private static final String PHENOTYPE_NEGATIVE_PROPERTY = NEGATIVE_PHENOTYPE_PREFIX + PHENOTYPE_POSITIVE_PROPERTY;
 
-    private static final String[] PHENOTYPE_PROPERTIES =
-        new String[] { PHENOTYPE_POSITIVE_PROPERTY, PHENOTYPE_NEGATIVE_PROPERTY };
+    private static final String NEGATIVE_PRENATAL_PHENOTYPE_PROPERTY = NEGATIVE_PHENOTYPE_PREFIX
+        + PRENATAL_PHENOTYPE_PROPERTY;
+
+    private static final String[] PHENOTYPE_PROPERTIES = new String[] { PHENOTYPE_POSITIVE_PROPERTY,
+        PHENOTYPE_NEGATIVE_PROPERTY, PRENATAL_PHENOTYPE_PROPERTY, NEGATIVE_PRENATAL_PHENOTYPE_PROPERTY };
 
     private static final String DISORDER_PROPERTIES_OMIMID = "omim_id";
 
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
@@ -133,6 +133,9 @@ public class APGARController implements PatientDataController<Integer>
         }
         PatientData<Integer> data = patient.getData(getName());
         if (data == null || !data.isNamed()) {
+            if (selectedFieldNames != null && hasAnySelected(selectedFieldNames)) {
+                json.put(DATA_NAME, new JSONObject());
+            }
             return;
         }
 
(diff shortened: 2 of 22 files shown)


veronikaslc:
@AlexLapena @allasm please, test/review

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
```diff
@@ -108,8 +108,10 @@ public void save(Patient patient)
                 Integer value = data.get(propertyName);
                 BaseProperty<ObjectPropertyReference> field =
                     (BaseProperty<ObjectPropertyReference>) dataHolder.getField(propertyName);
-                if (field != null && value != null) {
+                if (value != null) {
```
allasm: what about the "field != null" check?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
```diff
@@ -180,6 +182,9 @@ private boolean hasAnySelected(Collection<String> selectedFieldNames)
                     if (NumberUtils.isDigits(value)) {
                         parsed.put(propertyName, Integer.valueOf(value));
                     }
+                    if (StringUtils.isEmpty(value)) {
+                        parsed.put(propertyName, null);
```
allasm: do we want a "null" or an empty string here?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -507,6 +507,7 @@ private void updateDisordersFromJSON(XWikiDocument doc, BaseObject data, XWikiCo
                 // as in constructor: make unmofidiable
                 this.disorders = Collections.unmodifiableSet(this.disorders);
 
+                data.set(DISORDER_PROPERTIES_OMIMID, null, context);
```
allasm: why do we have two data.set(DISORDER_PROPERTIES_OMIMID, ...) on two neighbouring lines?

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/ParentalAgeControllerTest.java
```diff
@@ -157,7 +157,7 @@ public void loadMaternalAgeNonZero()
 
         Assert.assertEquals("parentalAge", testData.getName());
         Assert.assertTrue(testData.get(MATERNAL_AGE) == AGE_NON_ZERO);
-        Assert.assertNull(testData.get(PATERNAL_AGE));
+        Assert.assertEquals(AGE_ZERO, testData.get(PATERNAL_AGE));
```
allasm: I believe this test works, but I don't understand why it used to be a null and now it is a 0. Also, what does this case tests, and what does age 0 means - don't we actually want to return null when age is 0?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ParentalAgeController.java
```diff
@@ -181,7 +179,11 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
         for (String property : getProperties()) {
             if (data.has(property)) {
                 int age = data.getInt(property);
-                result.put(property, age);
+                if (age == 0) {
+                    result.put(property, null);
```
allasm: Is there a difference between age 0 and age == null? Don't both logically mean "no age is set", and if yes, isn't "null" a better indication?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -114,14 +114,17 @@
     public void writeJSON(Patient patient, JSONObject json, Collection<String> selectedFieldNames)
     {
         PatientData<T> data = patient.getData(getName());
+        String jsonPropertyName = getJsonPropertyName();
+
         if (data == null) {
```
allasm: it seems we still wont add anything to generated JSON if `data == null` but `selectedFieldNames` has a field this controller controls (as opposed to adding empty values that we should in this case)

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
veronikaslc: to overwrite

allasm: as discussed, +1

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
allasm: @marta, is that right that `field` can not be null unless `dataHolder`, which is `doc.getXObject(Patient.CLASS_REFERENCE)` is incorrect and does not have the correct `propertyName` (which is supposedly should, if the version of the code matches the version of the classes in DB)? In which case we probably do want a `NullPointerException` so that we see there is a problem in the installation (class mismatches code)

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/APGARController.java
allasm: as discussed, we probably want a separate issue to consistently represent missing data with empty strings or nulls, and not a random mix of 0-s, null-s and empty strings like we do now

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ParentalAgeController.java
```diff
@@ -181,7 +184,11 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
         for (String property : getProperties()) {
             if (data.has(property)) {
                 int age = data.getInt(property);
```
allasm: we should always use optInt(property,0) to correctly handle misformatted data (will also handle empty strings ans nulls instead of 0s as indicators for the missing age)

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -491,7 +494,10 @@ private JSONArray joinArrays(JSONArray jsonOne, JSONArray jsonTwo)
     private void updateDisordersFromJSON(XWikiDocument doc, BaseObject data, XWikiContext context, JSONObject json)
     {
         try {
-            JSONArray inputDisorders = json.optJSONArray(JSON_KEY_DISORDERS);
+            if (!json.has(JSON_KEY_DISORDERS)) {
+                return;
```
allasm: why this change? optJSONArray(JSON_KEY_DISORDERS) would return null if either there is no key or key value is not an array (e.g. misformatted JSON)

with new code, no key is covered, by value not being a proper JSON array will result in a JSONException

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -202,7 +202,7 @@ private Object format(String key, Object value)
     /** For converting JSON into internal representation. */
     private Object inverseFormat(String key, Object value)
     {
-        if (value != null) {
+        if (value != null && !value.equals(null)) {
```
allasm: just curious how does value.equals(null) work, is it for JSONObject.NULL?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -290,23 +306,25 @@ public void save(Patient patient)
             XWikiDocument doc = (XWikiDocument) this.documentAccessBridge.getDocument(patient.getDocument());
             BaseObject dataHolder = doc.getXObject(getXClassReference());
             PatientData<T> data = patient.getData(this.getName());
-            if (dataHolder == null && data != null) {
+            if (dataHolder == null || dataHolder == null) {
```
allasm: dataHolder == null || dataHolder == null ?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AllergiesController.java
```diff
@@ -175,15 +175,15 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
     public PatientData<String> readJSON(JSONObject json)
     {
         List<String> result = new ArrayList<>();
+        if (!json.has(DATA_NAME)) {
+            return null;
+        }
```
allasm: minor: move this before you create result = new ArrayList()

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
allasm: I think with that change - https://github.com/phenotips/phenotips/pull/2047/commits/9aa4f6ae0423e0765ea08eb3988df7de22b5d30f#diff-a359c7d8e774fa88f6b015042ad1f9a0R179 - data will actually be null, and the old code was correct?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
veronikaslc: yup

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
veronikaslc: had to be data, typo

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
veronikaslc: actually yes, didn't think about not array case

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ParentalAgeController.java
allasm: bump

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AbstractComplexController.java
```diff
@@ -287,26 +306,25 @@ public void save(Patient patient)
...
-                    continue;
-                }
-                if (this.getCodeFields().contains(propertyName) && this.isCodeFieldsOnly()) {
-                    @SuppressWarnings("unchecked")
-                    List<VocabularyProperty> terms = (List<VocabularyProperty>) propertyValue;
-                    List<String> listToStore = new LinkedList<>();
-                    for (VocabularyProperty term : terms) {
-                        String name = StringUtils.isNotBlank(term.getId()) ? term.getId() : term.getName();
-                        listToStore.add(name);
+                if (isInKeySet(data, propertyName)) {
```
allasm: after PT-2670 is merged in, we can just write if (data.hasKey(propertyName)) .... and remove isInKeySet()

sdumitriu:
Manually merged.

PR #2048, opened 2016-07-05
PT-1538: Date of birth/death in export does not capture "fuzziness"

allasm:
Also fixes PT-2604: Unable to update patient's DOB and DOD through REST

Requires extensive testing, @AlexLapena please test and review, @veronikaslc please review


components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsDate.java
```diff
@@ -0,0 +1,273 @@
...
+import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.json.JSONObject;
+
+/**
+ * Implementation of a fuzzy date for internal use in Phenotips.
+ *
+ * TODO: if approved, separate into an interface and an implementation?
```
sdumitriu: No, just a class is enough.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsDate.java
```diff
@@ -0,0 +1,273 @@
...
+            return null;
+        }
+    }
+
+    private boolean valueInRange(Integer value, Integer validRangeMin, Integer validRangeMax)
+    {
+        if (validRangeMin != null && value < validRangeMin) {
+            return false;
+        }
+        if (validRangeMax != null && value > validRangeMax) {
```
veronikaslc: Can be simplified to `if (validRangeMax != null && (value > validRangeMax || value < validRangeMin) {`

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsDate.java
```diff
@@ -0,0 +1,273 @@
...
+    private Integer stringToIntegerInRange(String str, Integer validRangeMin, Integer validRangeMax)
+    {
+        try {
+            Integer value = Integer.parseInt(str);
+            if (!valueInRange(value, validRangeMin, validRangeMax)) {
+                return null;
+            }
+            return value;
+        } catch (Exception ex) {
+            return null;
```
veronikaslc: I personally got used to cunture that `return null;` is the last line of method, up to dev I guess

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsDate.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    private boolean valueInRange(Integer value, Integer validRangeMin, Integer validRangeMax)
+    {
+        if (validRangeMin != null && value < validRangeMin) {
```
veronikaslc: can be simplified to ` if ( validRangeMin != null && value < validRangeMin || validRangeMax != null && value > validRangeMax )`

veronikaslc:
General question whether we should validate for some months having less days than 31.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4439,24 +4443,42 @@ document.observe("xwiki:dom:loaded", function() {
...
+              values.push({"value" : year, "selected" : true });
+          }
+          var yearInCentury = (year != null) &amp;&amp; (year &gt;= centuryStart) &amp;&amp; (year &lt; centuryEnd) &amp;&amp; (range &gt; 1);
+          values.push({"value": centuryStart + "s", "cssClass" : "century", "selected" : yearInCentury});
+      }
+      insertCentury(1800);
+      insertCentury(1700);
+      insertCentury(1600);
+      insertCentury(1500);
+      if (year != null &amp;&amp; year &lt; 1500) {
```
veronikaslc: so we push if `year < 1500`?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsDate.java
allasm: yes, but having two separate simple logical expressions is more clear (for the programmer) than one long one (especially without grouping brackets). In terms of performance/runtime, compiler will simplify to the above anyway.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
allasm: this "push" just adds a custom year, and if the year is less than 1500, it must be the last one in the list, so we push it last. If it is before that, it would be pushed before

allasm: a very smart date picker can probably change the range of years available for selection based on the currently selected year, but that is another story, and since we want to replace this fuzzy date picker with matt's new picker (eventually) it is irrelevant

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4439,24 +4443,42 @@ document.observe("xwiki:dom:loaded", function() {
...
+
     createYearDropdown : function() {
       this.yearSelector = new widgets.FuzzyDatePickerDropdown({name: "year"});
 
       var today = new Date();
       var crtYear = today.getYear() + 1900;
       var startYear = 1900;
 
+      var range = (this.__date.hasOwnProperty("range") &amp;&amp; this.__date.range.hasOwnProperty("years")) ? this.__date.range.years : 1;
+      var year = this.__date.hasOwnProperty("year") ? this.__date.year : null;
```
allasm: probably best to replace the variable name with "input_year" to avoid confusion

PR #2056, opened 2016-07-15
PT-2648: Support exporting of Clinically Normal status.

veronikaslc:
Fixed.


veronikaslc:
@AlexLapena @allasm please test/review

allasm:
This fixes only the Excel part, right? The original issue also said "When exporting into JSON, the value is always blank"

allasm:
Update: apparently JSON was fixed here: https://github.com/phenotips/phenotips/commit/e552a042ff2806805114f084d883d9ce64e80752, but Sasha tested development.phenpotipos which apparently does not have the fix

PR #2058, opened 2016-07-18
PT-2662: Disable self-registration

marta-:
This removes registration rights from guests and only allows Admins to create new accounts.


marta-:
@AlexLapena please review & test

AlexLapena:
+1

PR #2059, opened 2016-07-18
[misc] Checking user for being null.

veronikaslc:
One of cases is when user is XWiki.XWikiGuest.


components/security/authorization/pom.xml
```diff
@@ -55,4 +55,7 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <properties>
+    <coverage.instructionRatio>0.98</coverage.instructionRatio>
```
sdumitriu: Please add a test.

veronikaslc:
ready for merge?

PR #2060, opened 2016-07-22
PT-2670: Improve PatientData interface

allasm:
@itaiGershtansky @veronikaslc please review

### DIFF
--- a/components/patient-data/api/src/main/java/org/phenotips/data/PatientData.java
+++ b/components/patient-data/api/src/main/java/org/phenotips/data/PatientData.java
@@ -100,6 +100,15 @@ public interface PatientData<T> extends Iterable<T>
     boolean isNamed();
 
     /**
+     * For dictionary data only, checks if the given key is present in the dictionary data.
+     *
+     * @param key the key to be checked for
+     * @return {@code true} if the key-value based data structure has the specified key
+     * @since 1.3M2
+     */
+    boolean containsKey(String key);
+
+    /**
      * For dictionary data only, return an iterator over the dictionary keys.
      *
      * @return iterator containing all the keys, or an empty iterator if there are no keys or this is not a dictionary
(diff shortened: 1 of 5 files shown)


components/patient-data/api/src/main/java/org/phenotips/data/PatientData.java
```diff
@@ -100,6 +100,15 @@
     boolean isNamed();
 
     /**
+     * For dictionary data only, checks if the given key is present in the dictionary data.
+     *
+     * @param key the key to be checked for
+     * @return {@code true} if the key-value based data structure has the specified key
+     * @since 1.0M13
```
veronikaslc: @since 1.3M2?

veronikaslc: +1 other than that minor stuff

veronikaslc:
+1

itaiGershtansky:
+1

sdumitriu:
`Map` uses `containsKey` instead of `hasKey`. And while we're at it, why not also add `containsValue`?

allasm:
containsKey() probably makes sense for consistency - will amend

...but I don't see a use for containsValue() in the context of PatientData, as of right now (when the need arises we'll add it?). That, and the fact that containsKey() is probably more efficient when implemented using equivalent internal map method as opposed to using keyIterator, while containsValue is implemented with the same efficiency internally as it can be using existing valueIterator.

sdumitriu:
Not really, `contains` also uses hashcode and equals, so for a large map it will be faster.

What's different is that I can't find a usecase for it.

allasm:
I updated the method name.

I checked a few implementations of `containsVallue()` in standard maps in java.util, and all of them seem to be O(n) or O(hash size) or alike, which is generally comparable to O(n)... given that and the current lack of use for `constainsValue()`, we should probably not clutter the interface with an extra method, which only makes sense for Dictionary implementation of `PatientData` to start with.

By the way, I think we over-engineered PatientData, and we should only have one implementation, the dictionary one, and `Indexed` should use dictionary with keys being numbers, and `Simple` use dict. with one controller-defined key.  (possibly an extension of Dictionary implementation with a convenient constructor)

PR #2061, opened 2016-07-22
Fixed DateTimePicker for NIAID

Ruxuan:
@veronikaslc @danielpgross Please review.

This change fixes a bug where the DateTimePicker was not appearing on first load for date fields in a new table (project for a Gene42 client).

### DIFF
--- a/components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
+++ b/components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
@@ -2344,7 +2344,7 @@ CalendarDateSelect = Class.create({
       this.options = Object.extend(Object.clone(this.options), options || {});
       this.attachPickers();
       document.observe('xwiki:dom:updated', function(event) {
-        this.attachPickers(event.memo.element);
+        this.attachPickers(event.memo.elements);
       }.bindAsEventListener(this));
     },
     attachPickers : function (container) {
--- a/components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
+++ b/components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
@@ -2343,6 +2343,9 @@ CalendarDateSelect = Class.create({
     initialize : function(options) {
       this.options = Object.extend(Object.clone(this.options), options || {});
       this.attachPickers();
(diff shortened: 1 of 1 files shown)


components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
```diff
@@ -2343,6 +2343,9 @@ CalendarDateSelect = Class.create({
     initialize : function(options) {
       this.options = Object.extend(Object.clone(this.options), options || {});
       this.attachPickers();
+      document.observe('xwiki:dom:updated', function(event) {
+        this.attachPickers(event.memo.element);
```
sdumitriu: `event.memo.elements` is the right variable.

veronikaslc:
Is there any way to test?

Ruxuan:
I've tested this with a new patient section called biospecimens (from the Gene42 NIAID project). I'll send you the .xar on slack.

veronikaslc:
Tested. +1

danielpgross:
FTR, the obvious question here is, why are datepickers not already broken for other dynamic tables that use datepickers?
The answer is that the responsibility is currently on the consumer of the datepicker, rather than the datepicker itself, to make sure that new date fields have the picker attached to them.
Since we are shifting the responsibility here into the datepicker, we should be careful about other consumers of the datepicker that are still handling the attachment on DOM update. The only one I could find is in [old measurements](https://github.com/phenotips/phenotips/blob/3c753da4ff202463c0a1dcce3d021d36096f4fc4/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml#L1720).
Speaking more generally, because this widget is used in multiple places, we should be looking out for regressions introduced in different parts of the application.

sdumitriu:
Other tables work because:
1. sometime between XWiki 5 and XWiki 7, the datepicker has become part of the platform, including some improvements, so that's why the biospecimes work after the first entry is added and the page is reloaded
2. the measurements JS code has a custom listener that does exactly what Ru proposes, attaching pickers when new rows are added

sdumitriu:
Manually [merged](https://github.com/phenotips/phenotips/commit/374dc7c93f761eb50d0662500a584e0a9c2cc7c2) into master.

PR #2066, opened 2016-08-15
Implemented filtering numbers by min/max in patients table

Ruxuan:
Please review @marta @danielpgross 

Filtering number properties in the patients data table by specifying a min and max.
For task https://phenotips.atlassian.net/browse/NIAID-52.

Intended use example:
`// DatabaseFilters.xml`
`...`
`<input type="text" name="number_property/min" value="" />`
`<input type="text" name="number_property/max" value='" />`
`...`

### DIFF
--- a/components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
+++ b/components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
@@ -398,8 +398,10 @@
       #set ($discard = $filterValueList.add($i))
     #end
   #end
-  #set ($filterValueLowLimit = "$!{request.get(""${colname}/after"")}")
-  #set ($filterValueHighLimit = "$!{request.get(""${colname}/before"")}")
+  ## High/Low value filters (currently used for date and numbers)
+  #set ($filterValueLowLimit = "$!{request.get(""${colname}/after"")}$!{request.get(""${colname}/min"")}")
+  #set ($filterValueHighLimit = "$!{request.get(""${colname}/before"")}$!{request.get(""${colname}/max"")}")
+
   #set($tableAlias = $colname)
   ## only safe versions should be used in SQL queries
   #if (!$filterValueList &amp;&amp; "$filterValueLowLimit$filterValueHighLimit" == '')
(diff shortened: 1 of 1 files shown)


components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
```diff
@@ -474,8 +476,46 @@
...
+    #elseif($propType == 'NumberClass')
+      #livetable_getTableName($colname)
+      #set($fromSql = "${fromSql}, ${tableName} ${safe_tableAlias}")
+      #set($whereSql = "${whereSql} and ${targetedObject}.id=${safe_tableAlias}.id.id and ${safe_tableAlias}.id.name = ?")
+      #set($discard = $whereParams.add("${colname}"))
+      ## Converts each used value into correct type
+      #set($numberType = $propClass.get($colname).getProperty('numberType').getValue())
+      #if($numberType == 'integer')
+        #set($typedFilterValue = $mathtool.toInteger($filterValue))
+        #set($typedFilterValueLowLimit = $mathtool.toInteger($filterValueLowLimit))
```
marta-: If the user enters min = `5.5` for an `integer` or a `long` property, this code will include the results where the property value is `5`. This is counterintuitive for users. We can't expect them to know the property is an integer and they're only supposed to enter integer values in the filters. In my opinion min/max filters should *not* be rounded down to the nearest integer in any of the 4 cases.

components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
```diff
@@ -474,8 +476,46 @@
...
+    ## Numbers, exact match or min/max filters
+    #elseif($propType == 'NumberClass')
+      #livetable_getTableName($colname)
+      #set($fromSql = "${fromSql}, ${tableName} ${safe_tableAlias}")
+      #set($whereSql = "${whereSql} and ${targetedObject}.id=${safe_tableAlias}.id.id and ${safe_tableAlias}.id.name = ?")
+      #set($discard = $whereParams.add("${colname}"))
+      ## Converts each used value into correct type
+      #set($numberType = $propClass.get($colname).getProperty('numberType').getValue())
+      #if($numberType == 'integer')
+        #set($typedFilterValue = $mathtool.toInteger($filterValue))
```
marta-: I'm not convinced that rounding to the nearest integer is the best approach here. Again, it may be counterintuitive for users (who are not technical) and may wonder what kind of magic we're using to produce results. An exact match to the user input is probably what users would expect. If the user does not enter an integer value, then simply do not return any results.

components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
```diff
@@ -474,8 +476,46 @@
...
+        #set($typedFilterValue = $mathtool.toDouble($filterValue).longValue())
+        #set($typedFilterValueLowLimit = $mathtool.toDouble($filterValueLowLimit).longValue().valueOf($filterValueLowLimit))
+        #set($typedFilterValueHighLimit = $mathtool.toDouble($filterValueHighLimit).longValue().valueOf($filterValueHighLimit))
+      #end
+      ## Checks present filter value
+      #if ($filterValue != '')
+        #set($whereSql = "${whereSql} and ${safe_tableAlias}.value = ?")
+        #set($discard = $whereParams.add($typedFilterValue))
+      #else ## filters by min and max values
+        #if ($filterValueLowLimit != '')
```
danielpgross: Shouldn't this instead be `$typedFilterValueLowLimit` to cover the case when the input is not a valid number?

components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
```diff
@@ -474,8 +476,46 @@
...
+          #set($whereSql = "${whereSql} and ${safe_tableAlias}.value &gt;= ?")
+          #set($discard = $whereParams.add($typedFilterValueLowLimit))
+        #end
+        #if ($filterValueHighLimit != '')
```
danielpgross: Shouldn't this instead be `$typedFilterValueHighLimit` to cover the case when the input is not a valid number?

components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml
```diff
@@ -474,8 +476,46 @@
...
+        #set($typedFilterValue = $mathtool.toDouble($filterValue))
+        #set($typedFilterValueLowLimit = $mathtool.toDouble($filterValueLowLimit))
+        #set($typedFilterValueHighLimit = $mathtool.toDouble($filterValueHighLimit))
+      #else ## Long
+        #set($typedFilterValue = $mathtool.toDouble($filterValue).longValue())
+        #set($typedFilterValueLowLimit = $mathtool.toDouble($filterValueLowLimit).longValue().valueOf($filterValueLowLimit))
+        #set($typedFilterValueHighLimit = $mathtool.toDouble($filterValueHighLimit).longValue().valueOf($filterValueHighLimit))
+      #end
+      ## Checks present filter value
+      #if ($filterValue != '')
```
danielpgross: Shouldn't this instead be `$typedFilterValue` to cover the case when the input is not a valid number?

Ruxuan: Number filtering stops working when I replace `$filterValue*` with `$typedFilterValue*`
I'm looking into it. @sdumitriu @marta- Any ideas on why this could be happening?

sdumitriu: `typed` is a number, which is different from a string. Try `"$!typedFilterValue" != ''`

danielpgross:
Looks good, +1

PR #2068, opened 2016-08-25
Spanish translation of the HPO.

josecv:
Including both the human and machine translated elements, etc.

Could @itaiGershtansky and @allasm review please? Also, I wasn't sure which branch to merge it into, so if master isn't the right one feel free to redirect.


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyInputTerm.java
```diff
@@ -31,8 +36,13 @@
...
  * @version $Id$
  */
-public class SolrVocabularyInputTerm extends SolrVocabularyTerm implements VocabularyInputTerm
+public class SolrVocabularyInputTerm extends AbstractSolrVocabularyTerm implements VocabularyInputTerm
 {
+    /* Inexplicably, SolrInputDocument and SolrDocument aren't (in version 5.3.2) in the same class hierarchy,
+     * which is why there has to be a common parent to this and the solrvocabularyterm instead of just
+     * extending from it. This is particularly serious in the get() method which has a bug as described
+     * below. */
+
```
josecv: Actually the bug is gone now, and was the result of an oversight on my part. Fixing this comment.

sdumitriu:
Redone as #2202

PR #2070, opened 2016-08-26
PT-2669: Make the patient record form easier to configure

AlexLapena:
@sdumitriu @marta- Please review


marta-:
The change to PatientSheet.xml, which removes the hardcoded consent filtering, should also be committed.

a-gvo:
@sdumitriu @marta- ready for review

sdumitriu:
@sashaandjic @AlexLapena @a-gvo Please review and test this.

a-gvo:
I have a question about a hypothetical scenario. 

Say a consent is configured to disallows entering gene and phenotype data if not granted. Suppose I create a new patient, grant this consent, enter some genotype and phenotype data, then ungrant the consent and save the patient data. Then,

1. should "Matching Disorders in OMIM" and "Gene Panels" show anything based on that data, and
2. should filtering by gene in "Advanced Data Filters" show the patient with the non-consented data?

sdumitriu:
Currently, that is supposed to happen, since unchecking a consent doesn't discard existing data, and consents only affect what's shown in the UI, it doesn't hide the data when accessed through other means: queries, JSON, excel export, push, API access...

I'm not sure it's OK to discard data, that may lead to accidental data loss, and unchecking a consent is not something that will happen in the vast majority of cases, if ever.

In the future, once we have access rights per piece of data, instead of per whole patient record, we may add an authorization module that rejects access to data that hasn't been consented through any means, while still storing it for future use.

marta-:
@a-gvo that's a very good point. The answer is no, the two should not be shown.

The good news is:
- I believe the behaviour has not changed since the former implementation, so this refactoring does not introduce a regression
- The chance of being consented for genes but not phenotypes in the real world is almost null.

However, this does reveal some issues with how we handle dependent data in general. Something to think about in the future.

AlexLapena:
+1

PR #2076, opened 2016-08-31
PT-2763: Suggest dialogs should hide when they lose focus.

veronikaslc:
Better fix takes into account native xwiki suggest widget and works for both administration and patient form.


veronikaslc:
@itaiGershtansky @a-gvo please, review

a-gvo:
+1

PR #2082, opened 2016-09-19
PT 2782 and PT-2786

allasm:
Fixed patient creation via REST and improved patient controller performance by eliminating multiple unnecessary document saves


allasm:
Tested by @veronikaslc (for PT-2786)
Tested by @allasm (for PT-2782)

PR #2085, opened 2016-09-22
Modify the specificity meter to not fire an event on initialization

danielpgross:
Currently, on initialization of the specificity meter (on page load), an event is always fired with value "0", even though this is not the actual specificity score. That event should not be fired because it does not represent an actual change in value, nor does it contain the actual specificity score.

This change is necessary for a downstream module that triggers certain behaviour when the specificity score changes. With the current "false positive" behaviour of the specificity meter, the downstream module does not work properly.


danielpgross:
@itaiGershtansky @veronikaslc please review.

itaiGershtansky:
+1

veronikaslc:
+1

PR #2090, opened 2016-10-06
PT-2808: Implemented parseToJSONObject and parseToJSONArray

a-gvo:
@veronikaslc @itaiGershtansky please review

### DIFF
--- a/components/base-war/src/main/java/org/xwiki/velocity/tools/JSONTool.java
+++ b/components/base-war/src/main/java/org/xwiki/velocity/tools/JSONTool.java
@@ -113,9 +113,9 @@ public class JSONTool
      * @return the parsed JSON, either a {@link net.sf.json.JSONObject} or a {@link net.sf.json.JSONArray}, or
      *         {@code null} if the argument is not a valid JSON
      * @since 5.2M1
+     * @deprecated use {@link #parseToJSONArray(String)}} or {@link #parseToJSONObject(String)} instead
      */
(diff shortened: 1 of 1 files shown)


itaiGershtansky:
+1

components/base-war/src/main/java/org/xwiki/velocity/tools/JSONTool.java
```diff
@@ -127,6 +127,46 @@ public JSON parse(String json)
...
+        } catch (org.json.JSONException ex) {
+            this.logger.info("Tried to parse invalid JSON: [{}], exception was: {}",
+                    StringUtils.abbreviate(json, 32), ex.getMessage());
+            return null;
+        }
+    }
+
+    /**
+     * Parse a serialized JSON into a real JSONArray. Only valid strings can be parsed, and doesn't support
+     * JSONP. If the argument is not valid JSON, then {@code null} is returned.
```
allasm: if the argument is not a vlaid JSON  array or not a valid JSON in general? E.g. is "[1,2,2]" supported and is "{"a": [1,2,3]} ? I presume it only supportes input which represents a valid JSON array (not any valid JSON) and we should document that

a-gvo: @allasm Something like "[1, 2, 2]" should work, but yes, the second one will throw an exception.

PR #2092, opened 2016-10-08
PT-2230: Allow exporting pedigree as an image from within pedigree editor.

veronikaslc:
Fixed.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/filesaver/FileSaver.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/filesaver/FileSaver.js
@@ -238,12 +238,12 @@ define([
   // with an attribute `content` that corresponds to the window
 
   var saveTextAs = saveTextAs
-  || (function (textContent, fileName, charset) {
+  || (function (textContent, fileName, type) {
       fileName = fileName || 'pedigree.txt';
-      charset = charset || 'utf-8';
       textContent = (textContent || '').replace(/\r?\n/g, "\r\n");
+      type = type || "text/plain;charset=utf-8";
       if (saveAs && Blob) {
-          var blob = new Blob([textContent], { type: "text/plain;charset=" + charset });
+          var blob = new Blob([textContent], { type: type });
           saveAs(blob, fileName);
           return true;
       } else {//IE9-
@@ -279,9 +279,19 @@ define([
       }
   })
 
+  var saveTextFile = function (textContent, fileName) {
+	  saveTextAs(textContent, fileName, "text/plain;charset=utf-8");
+  }
+
+  var saveSVGFile = function (textContent, fileName) {
+	  saveTextAs(textContent, fileName, "data:image/svg+xml;charset=utf-8");
+  }
+
   var FileSaver = {};
   FileSaver.saveAs = saveAs;
   FileSaver.saveTextAs = saveTextAs;
+  FileSaver.saveTextFile = saveTextFile;
+  FileSaver.saveSVGFile = saveSVGFile;
 
   return FileSaver;
 });
--- a/components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
@@ -32,6 +32,7 @@ define([
             typeListElement.insert(_addTypeOption(true,  "PED", "ped"));
             typeListElement.insert(_addTypeOption(false, "BOADICEA", "BOADICEA"));
             typeListElement.insert(_addTypeOption(false, "Simple JSON", "simpleJSON"));
+            typeListElement.insert(_addTypeOption(false, "Image", "image"));
             //TODO: typeListElement.insert(_addTypeOption(false, "Phenotips Pedigree JSON", "phenotipsJSON"));
 
             var fileDownload = new Element('a', {"id": 'downloadLink', "style": 'display:none'});
@@ -54,9 +55,15 @@ define([
             configListElementPED.insert(this._addConfigOption(false, "ped-options", "export-subconfig-label", "Generate new numeric IDs", "newid"));
             configListElementPED.insert(this._addConfigOption(false, "ped-options", "export-subconfig-label", "Using Names (when available)", "name"));
 
+            var configListElementImage = new Element('table', {"id": "imageOptions", "style": 'display:none'});
+            var label = new Element('label', {'class': 'export-config-header'}).insert("Select type of generated pedigree image");
+            configListElementImage.insert(label.wrap('td').wrap('tr'));
+            configListElementImage.insert(this._addConfigOption(true,  "image-options", "export-subconfig-label", "Raster image (PNG)", "png"));
+            configListElementImage.insert(this._addConfigOption(false, "image-options", "export-subconfig-label", "Scalable image (SVG)", "svg"));
+
             var promptConfig = new Element('div', {'class': 'import-section'}).update("Options:");
             var dataSection3 = new Element('div', {'class': 'import-block'});
-            dataSection3.insert(promptConfig).insert(configListElementJSON).insert(configListElementPED);
+            dataSection3.insert(promptConfig).insert(configListElementJSON).insert(configListElementPED).insert(configListElementImage);
             mainDiv.insert(dataSection3);
 
             var buttons = new Element('div', {'class' : 'buttons import-block-bottom'});
@@ -86,6 +93,7 @@ define([
             var pedOptionsTable = $("pedOptions");
             var pedDisorderOptions = $$('[name="ped-disorders-options"]');
             var jsonOptionsTable = $("jsonOptions");
+            var imageOptionsTable = $("imageOptions");
 
             if (exportType == "ped" || exportType == "BOADICEA") {
                 pedOptionsTable.show();
@@ -102,20 +110,20 @@ define([
                     pedDisorderOptions.each( function(item) {item.up('tr').show();});
                 }
                 jsonOptionsTable.hide();
+                imageOptionsTable.hide();
             } else {
                 pedOptionsTable.hide();
                 pedDisorderOptions.each( function(item) {item.up('tr').hide();});
-                jsonOptionsTable.show();
+                if (exportType == "simpleJSON") {
+                    jsonOptionsTable.show();
+                    imageOptionsTable.hide();
+                } else {
+                    jsonOptionsTable.hide();
+                    imageOptionsTable.show();
+                }
             }
         },
 
-        /**
-         * Loads the template once it has been selected
-         *
-         * @param event
-         * @param pictureBox
-         * @private
-         */
         _onExportStarted: function() {
             this.hide();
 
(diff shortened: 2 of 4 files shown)


veronikaslc:
@allasm please, review

components/pedigree/resources/src/main/resources/pedigree/filesaver/FileSaver.js
```diff
@@ -238,12 +238,13 @@ define([
   // with an attribute `content` that corresponds to the window
 
   var saveTextAs = saveTextAs
-  || (function (textContent, fileName, charset) {
+  || (function (textContent, fileName, charset, type) {
       fileName = fileName || 'pedigree.txt';
       charset = charset || 'utf-8';
       textContent = (textContent || '').replace(/\r?\n/g, "\r\n");
+	  type = type || "text/plain;charset=" + charset;
```
allasm: identation

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -124,7 +132,44 @@ define([
             var exportType = $$('input:checked[type=radio][name="export-type"]')[0].value;
             //console.log("Import type: " + exportType);
 
-            if (exportType == "simpleJSON") {
+            if (exportType == "image") {
+                var imageType = $$('input:checked[type=radio][name="image-options"]')[0].value;
+
+				var _bbox = $('canvas').down().getBBox();
```
allasm: more identation

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -124,7 +132,44 @@ define([
...
+                var svgEl = $$("#canvas svg")[0];
+                var background = $$('.panning-background')[0];
+                svgEl.insertBefore(
+                    new Element('rect', {"id": "white-bbox-background",
+                                         "width": _bbox.width,
+                                         "height": _bbox.height,
+                                         "x": _bbox.x,
+                                         "y": _bbox.y,
+                                         "fill": "white"}),
+                                svgEl.firstChild
```
veronikaslc: this one is actually intentional :)

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -148,7 +193,11 @@ define([
 
             console.log("Export data: >>" + exportString + "<<");
             if (exportString != "") {
-                FileSaver.saveTextAs(exportString, fileName);
+                if (exportType != "image") {
+                    FileSaver.saveTextAs(exportString, fileName);
+                } else if (exportType == "image" && imageType == "svg") {
+                    FileSaver.saveTextAs(exportString, fileName, 'base64', 'data:image/png;base64');
```
allasm: I would instead add a FileSaver.saveAsBinary() method, which would set type ("base64 etc etc")  internally. This would make it clear SaveAsText is for text, SaveAsBinary is for binary, and implementational details will be inside the class

(and both SaveAsText() and SaveAsBinary() would call some othe rmethod "SaveAs" which in turn takes the extra required parameter for type)

allasm: ..otherwise the name "SaveAsText" is confusing, e.g. a programmer may say "wtf they 'save image as text' ???"

veronikaslc: it's not "SaveAsText". It is "saveTextAs". This line is a Safari fix, in this case we saving text as 'data:image/png'.

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeImageService.xml
```diff
@@ -37,13 +37,36 @@
...
+#if ("$!request.action" == 'export')
+  {{html wiki="false" clean="false"}}
+  $response.setContentType('application/json')##
+  #if ("$!request.image" == '')
+    $response.setStatus(404)##
+    {"error" : "$services.localization.render('phenotips.PedigreeImageService.noPedigreeImage')"}
+    #break
+  #end
+  ## Fix broken SVG generated by a broken version of the pedigree editor
+  #if ($stringtool.countMatches($request.image, 'xmlns=') &gt; 1)
```
allasm: we should do this inside pedigree editor instead. If this is a "broken" SVG we don't want to save it to family as well

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
allasm: but shouldn't it align with "new" since it "new Element" is first parameter, svgEl is second to the same function?

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
allasm: my bad for misreading function name, but the general argument stands that IMHO it makes ore sense to have two functions which only take user-relevant parameters (data and filename) while what we export should be somehow encoded in function name. The original idea was not "save text as <TYPE>" but "save text as <FILENAME>". I guess bad naming to start with

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeImageService.xml
allasm: in workspace.js we do very similar things so should be a copy-paste operation to move this one there I hope. I was sure we remove extra xmlns-ses already there (I reclal this problem!), if we do not it is a bug

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
veronikaslc: Disagree. "saveTextAs" already takes a charset as a third parameter.  If we need a separate function to set it to base64  why this parameter is here in a first place?
Having file type parameter there as an option seems pretty logical and clear. 
Very often file constructors take a file type as a standard parameter, like 
`var file = new File(["Hello, world!"], "hello world.txt", {type: "text/plain;charset=utf-8"});`
Why not having a file saving function with an option to specify type?

allasm: well, it is of course arguable what is bette, there is no clear better way.

To start with, it is there because for the generic library it may make sense, but for us file type is a very technical point that "I, the pedigree developer" couldn't care less about, I just want to "save the file", whatever way current technology/browsers support it. E.g. I want to expose as high a level API as reasonable, and since we generally dont deal with file here I think abstractin away low-level MIME details is reasonable. Of course it is ok to leave a sis since it works :), just saying my 2c

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeImageService.xml
veronikaslc: I think we do it in workspace.js here `var svgText = image.innerHTML.replace(/xmlns(:xlink)?=".*?"/g, '')`

veronikaslc: https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/view/workspace.js#L153

allasm: yeah. If it does not work and we have to ad new rules I would place it as a separate rule a few lines below where we do more replacings

btw you can always check in browser console whats the result of replacement to figure out at which point it fails

veronikaslc: There should be a reason @sdumitriu placed those lines in this image service XML file and not in workspace.js in this commit https://github.com/phenotips/phenotips/commit/0c74fa5c69bad684606ce6da9048b6575e8f0896
Especially if two days later there was your commit to workspace.js with line doing pretty much same with replacing xmlns-ses to fix generating SVG in IE9 https://github.com/phenotips/phenotips/commit/736ec819ffa93d4fceb4c4538e342cf1d2b7204f
From the header of @sdumitriu 's commit "PT-2249: The pedigree image service fails to generate PNGs for pedigrees created with 1.2.1" and the comment in commit code " Fix broken SVG generated by a broken version of the pedigree editor" am guessing that those fixes are placed in image service XML and not in pedigree code for the reason of fixing issues if someone will install old broken version of pedigree editor 1.2.1and placing fixes in newer pedigree code wont help in this case. Just a thought..

sdumitriu: There is a reason, indeed. Putting the fix only in the JavaScript runtime means that a pedigree that's already broken has to be re-opened and re-saved to get it fixed, while this is just a workaround so that the users have a more pleasant experience. But I agree that the proper fix has to be in the JS code, so that broken SVGs aren't generated at all.

components/pedigree/resources/src/main/resources/pedigree/filesaver/FileSaver.js
```diff
@@ -238,12 +238,13 @@ define([
   // with an attribute `content` that corresponds to the window
 
   var saveTextAs = saveTextAs
-  || (function (textContent, fileName, charset) {
+  || (function (textContent, fileName, charset, type) {
       fileName = fileName || 'pedigree.txt';
       charset = charset || 'utf-8';
       textContent = (textContent || '').replace(/\r?\n/g, "\r\n");
+      type = (type || "text/plain;charset=") + charset;
```
sdumitriu: This assumes that the `type` argument ends with `;charset=`, which isn't always the case. Either put the brackets around the last 2 things instead of the first 2, or add a check if the value so far ends with `charset=`.

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeImageService.xml
```diff
@@ -37,13 +37,29 @@
...
+  #end
+  {"url": "$pedigreeRasterUrl.serialize()"}
+  {{/html}}
+  #break
+#end
 #if ($xcontext.action != 'get')
   **Usage**: //http[s]:~/~/&lt;hostname&gt;[:&lt;port&gt;]/&lt;appname&gt;/bin/get/PhenoTips/PedigreeImageService?[e]id=&lt;identifier&gt;[&amp;format=raster]//
   #break
 #end
+{{include reference="PhenoTips.PedigreeMacros" /}}
```
sdumitriu: The include will happen after all of the Velocity code has executed, which means that the macros used after this line will fail.

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -124,7 +132,43 @@ define([
...
+                var svg = editor.getWorkspace().getSVGCopy();
+                var exportString = svg.getSVGText();
+                $('white-bbox-background').remove();
+
+                if (imageType == "png") {
+                    // generate PNG image on the back end
+                    var pedigreeImageExportServiceURL = editor.getExternalEndpoint().getPedigreeImageExportServiceURL();
+                    new Ajax.Request(pedigreeImageExportServiceURL, {
+                        method: 'POST',
+                        onSuccess: function(response) {
```
sdumitriu: And what if it fails? There should be a way of telling the user that the export failed, otherwise they may wait forever for something to happen.

components/pedigree/resources/src/main/resources/pedigree/filesaver/FileSaver.js
allasm: I think we should just have two method, saveTextFile and savePNGFile, with hardcoded charsets inside, and no checks, since that is the functionality we need, and developers reading h code using those methods will know exactly whats happening, without caring for hairy charset details

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
veronikaslc: as of per conversation with @allasm it was decided not to inform user

allasm: yes, we did. But on the second thought @sdumitriu is right, I kind of though that if you click and it doesn't export you realize right away it doe snot work, but in theory users who are used to slow software may not realize it fails but assume it just takes forever.... so yeah, I guess need to tell the user

veronikaslc: then we need to decide which way of telling the user that the export failed would be best here

allasm: just a popup, even javascript `alert` (but better an `okCancelDialogue`, which has a special "showError() or showWarning() method) saying "Internal error - image export failed".

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeImageService.xml
sdumitriu: Including `PedigreeMacros` shouldn't produce any extra markup, since it uses `output=false`, so the extra markup comes from the macros.

PR #2093, opened 2016-10-08
[misc] Pedigree Genotype label change from 'rejected' to 'negative'

veronikaslc:
For consistency with Genotype information section in patient form


allasm:
+1

PR #2097, opened 2016-10-11
Fixes for PT-2825 and PT-2829

allasm:


allasm:
@sdumitriu @veronikaslc please review

components/pedigree/resources/src/main/resources/pedigree/view/cancersLegend.js
```diff
@@ -51,7 +51,7 @@ define([
                                    "Myeloma" : "Myeloma",
                                    "Leukemia" : "Leukemia"
                                  };
-            $super('Cancers', true);
```
sdumitriu: The rest of the changes don't seem to be related to PT-2825, are they? Can this commit be split in two?

allasm: well, it is kind of related as this enables better "rejection" message. PT-2825 is not a bug, just the message is a bit unclear,and in case of group node there is no message at all. All of these changes are related to improving the messages. I guess a better "fix" is to re-word PT-2825 itself

PR #2098, opened 2016-10-12
PT-2723: Users should be able to select what traits are reflected in the affected column when exporting the PED file of a pedigree

sdumitriu:
Added phenotypes, cancers, genes


veronikaslc:
+1

PR #2101, opened 2016-10-12
[misc] Option to use custom file name for file download

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -185,7 +185,7 @@ define([
                         on0 : function(response) {
                             response.request.options.onFailure(response);
                         },
-                        parameters: {"image": exportString}
+                        parameters: {"image": exportString, "filename": patientDocument + ".png"}
```
sdumitriu: I wonder if it would be a good idea to also include a timestamp in the name; not just for the image, but for all export formats.

components/base-war/src/main/java/com/xpn/xwiki/web/TempResourceAction.java
```diff
@@ -117,7 +118,11 @@ public String render(XWikiContext context) throws XWikiException
         }
         response.setContentType(contentType);
         if ("1".equals(request.getParameter("force-download"))) {
-            response.addHeader("Content-disposition", "attachment; filename*=utf-8''" + tempFile.getName());
+            String fileName =
+                (!StringUtils.isBlank(request.getParameter("force-filename"))) ? request.getParameter("force-filename")
```
sdumitriu: Shorter: `StringUtils.defaultIfBlank(request.getParameter("force-filename"), tempFile.getName())`

PR #2103, opened 2016-10-13
[misc] Make section expand/collapse global and improve UX by importing it first

teyden:
- Make it also available for sections in the FamilySheet 
  @sdumitriu @marta please review


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4886,6 +4886,130 @@ document.observe("xwiki:dom:loaded", function() {
   </object>
   <object>
     <name>PhenoTips.Widgets</name>
+    <number>14</number>
```
allasm: Since JS extensions are executed in the order they are imported, we should not have this as a #14 widget since this is the first thing a user sees (sections collapsing/expanding), and which shoul dbe done ASAP to improve user experience

Instead we should place it into a separate document, and import first before all widgets and before PatientSheetCode (or FamilySheetCode)

I wanted to do this for a long time just had other things to do - since you started, maybe you can do that please?

teyden:
I see, that makes sense. But in terms of the setup, where does the import order get specified for the different pages? As in, how do I specify the order page's get imported?

allasm:
We import javascript extensions from PatientSheertCode in PatientSheet:

https://github.com/phenotips/phenotips/blob/master/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheet.xml#L53

...and I guess on FamilySheet as well (we just don't import much there yet):

https://github.com/phenotips/phenotips/blob/master/components/family-studies/ui/src/main/resources/PhenoTips/FamilySheet.xml#L40

So IMHO the proper way for this expand/collapse code is to have it in a separate document (maybe named smth like a better version of "FirstUICodeToLoad") and have it as the very first $xwiki.jsx.use() in both

teyden:
Ohh, ok - got it. I'll import "FirstUICodeToLoad" in both FamilySheet and PatientSheet and place them **first**.

teyden:
I have two concerns: 
- Are JS extensions imported differently for FamilySheet? It appears that it may be imported through a "uix" service that I am not familiar with.
- Where is the best place to put the separate, "FirstUICodeToLoad" document?

allasm:
1) UIX_Fields may import their own jsx, where do you know the code that uses widgets that work? We can look around that code to find where widgets are actually imported

2) I'm not sure, @sdumitriu is th eperosn to ask, but I suppose same location as our Widgets.xml

3) I would `git rebase -i HEAD~2` and `git push -f ...` and squash the two commits to make it simpler to see the changes and where stuff was moved

teyden:
1) Correct, however from what I've seen, none of the Family UIX fields (or equivalent pages, the naming scheme is for families is FamilyRecordField_*) import the widgets jsx. I'll search more today for the widgets import in family sheet - and perhaps @sdumitriu can advise?

2) Okay, good - I had put it in the same place as Widgets.xml. 

3) Thanks, done. Final and revised commit is above.

teyden:
@allasm, could you please review the changes when you get a chance?

allasm:
+1

I think this looks good. I'm not sure if `UISheetCode` is the best name, since most of the velocity code is UI-related, but I don't have a better name in mind (@sdumitriu?)

components/widgets/ui/src/main/resources/PhenoTips/UISheetCode.xml
```diff
@@ -0,0 +1,165 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+-->
+
+<xwikidoc version="1.1">
+  <web>PhenoTips</web>
+  <name>UISheetCode</name>
```
marta-: I agree with @allasm about `UISheetCode` not being the best name. @sdumitriu : would you say this js code's use is wide enough for it to be put as a jsx object in `PhenoTips.Skin`?

sdumitriu: If it's placed in `Skin`, then it will also affect family sheets. Do we want that?

sdumitriu: As I see in the pull request, yes, apparently we want that. Then yes, putting it in [`PhenoTips.Skin`](https://github.com/phenotips/phenotips/blob/master/components/skin/ui/src/main/resources/PhenoTips/Skin.xml) is better. We already have a bit of code there.

marta-: Ok. @teyden please note that the JS extensions from `PhenoTips.Skin` don't need to be pulled. They are among loaded on all pages before extensions pulled with `$xwiki.jsx.use()`.

marta-: Shouldn't you also remove the [code from `PatientSheetCode`](https://github.com/phenotips/phenotips/blob/master/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml#L121)?

teyden:
Oh, I thought I did. I must've left it out when I did a git rebase. Will
add the changes.

Best,

Teyden Nguyen

teyden:
^

PR #2104, opened 2016-10-14
[misc] Adding missing translations

veronikaslc:


allasm:
+1

components/patient-tools/src/test/java/org/phenotips/tools/FormSectionTest.java
```diff
@@ -42,13 +53,29 @@
...
+        try {
+            field.setAccessible(true);
+            @SuppressWarnings("unchecked")
+            Provider<ComponentManager> cmp = mock(Provider.class);
+            field.set(null, cmp);
+            ComponentManager cm = mock(ComponentManager.class);
+            when(cmp.get()).thenReturn(cm);
+            TranslationManager tm = mock(TranslationManager.class);
+            when(cm.getInstance(TranslationManager.class)).thenReturn(tm);
+            when(tm.translate("Phenotips.FormSection.suggestTermPlaceholder")).thenReturn("");
```
allasm: I would return "TRANSLATION1" instead of "" and correspondingly used TRANSLATION1 instead of "" a few lines below @ PT-2723: Users should be able to select what traits are reflected in the "affected" column when exporting the PED file of a pedigree

components/patient-tools/src/main/java/org/phenotips/tools/FormSection.java
```diff
@@ -98,11 +102,17 @@ private String generateSuggestionsField(DisplayMode mode, String[] fieldNames)
         String displayedLabel = "Other";
         result += String.format("<label for='%s' class='label-other label-other-%s'>%s</label>", id, fieldNames[YES],
             displayedLabel);
-
+        String placeholder = "";
```
sdumitriu: I'd put the original text here, as a fallback in case the translation fails.

components/patient-tools/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1 @@
+Phenotips.FormSection.suggestTermPlaceholder=enter free text and choose among suggested ontology terms
```
sdumitriu: While we're refactoring this, can you also replace `ontology` with `vocabulary please? Same for the Spanish translation below.

veronikaslc:
commit amended

PR #2106, opened 2016-10-17
PT-2838: Set the same font size on both consent warnings.

veronikaslc:
Fixed.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
@@ -677,6 +677,9 @@
 .error_updating_consents {
   margin-top: 1em;
   margin-left: 2em;
+}
+.view-title-required-not-granted {
+  font-size: 1.15em;
 }</code>
     </property>
     <property>


veronikaslc:
@allasm please review

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
```diff
@@ -677,6 +677,9 @@
 .error_updating_consents {
   margin-top: 1em;
   margin-left: 2em;
+}
+.view-title-required-not-granted {
+  font-size: 1.15em;
```
allasm: 1.15em is a magic value and is kind of a coincidence that it matches the other warning font size (which is not specified in em). A better fix is to make sure both fonts are explicitly set to the same value, whatever it is

marta-: [The whole 'before-patient-data' block (containing consents, permissions, studies) has `font-size: .87em`.](https://github.com/phenotips/phenotips/blob/master/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml#L3791) I'm not sure it's needed or even helps in any way. Should we maybe remove that instead of adding a new rule?

sdumitriu: +1 for removal instead of addition.

veronikaslc: removal of 0.87em increases sizes of all elements in the whole section, which contains both permissions and consents. Section looks better with it when everything is smaller.

marta-: Whether it looks better or not is subjective. We're all used to having that section in smaller fonts so that may influence the perception a lot. But I think there are too many font sizes on the screen already and having that section in "normal" font can make everything look cleaner. Just my 2 cents. Other opinions are welcome.

veronikaslc: Small version looks cleaner to me 
@allasm wdyt? 
small
big

veronikaslc: I would not display the top warning in a view mode at all. It is a duplication of information that not all consents are granted.

marta-: That's a good point.

marta-: Hmm, i'd remove the bottom one actually.

veronikaslc: I'll make a new PR

marta-: @allasm This is all deprecated now. Please see https://github.com/phenotips/phenotips/pull/2120 instead.

allasm: This is deprecated, but 2120 does not cover everything this was supposed to fix, does it?

PR #2108, opened 2016-10-19
PT-2851: After linking an existing patient record to an individual in…

marta-:
… the pedigree, any phenotypes, diseases and genes in that report will not appear in the node menu don't appear in the node menu


marta-:

components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
```diff
@@ -724,6 +724,7 @@ define([
                      //Event.fire(patientPicker, 'custom:selection:changed', { "useValue": { "phenotipsid": event.memo.id } });
                     Event.fire(patientPicker, 'custom:selection:changed', { "useValue": event.memo.id });
                     _this.reposition();
+		    _this.refresh();
```
sdumitriu: No tabs please.

allasm:
proper solution is to refresh disorder names when they change for any reason, e.g. after a user edit as well (not supported yet, but at some point it was discussed). The fix is also simple, I will implement it quickly now

PR #2109, opened 2016-10-19
PT-2822: In Chrome, the "Advanced data filters" accumulate duplicate entries

alexhenrie:
Fixed duplication of phenotype field

There were actually two problems here:
1. Selections were copied to the URL from both the textbox and the checkboxes, creating duplicate entries in the URL and breaking the Back button.
2. The checkboxes were being duplicated, which rewrote the URL several more times (breaking the Back button more) before returning to the same URL with all entries duplicated.

The fix is to prevent either step from creating a URL with duplicate entries.

Of the problems mentioned in [PT-2822](https://phenotips.atlassian.net/browse/PT-2822), this patch only fixes the phenotype duplication. Separate fixes will be required for the other duplicated fields.

@marta-

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
@@ -158,16 +158,22 @@ document.observe('xwiki:dom:loading', function() {
               item._suggestPicker = suggestPicker;
 
               // Format the predefined value
-              item.value.split(',').each(function(value) {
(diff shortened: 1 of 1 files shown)


alexhenrie:
@marta- @AlexLapena Have you had a chance to look at this?

marta-:
@alexhenrie yes, briefly.  I'm hoping we can fix the root cause or at least patch it more generically. We're preparing a change where filters are configurable by the user, so different types of filters will appear. It's important not to be forced to write a specific fix like this for each one.

PR #2112, opened 2016-10-21
PT-2858: Push dialog displays text in language that is not chosen

marta-:


marta-:
@veronikaslc

veronikaslc:
+1

marta-:
@a-gvo

PR #2115, opened 2016-10-25
Pt-2640-entities

itaiGershtansky:
@sdumitriu, @danielpgross please review


components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
```diff
@@ -53,7 +55,14 @@
...
      */
-    DocumentReference getDocument();
+    DocumentReference getDocumentReference();
+
+    /**
+     * Returns the document where the entity is stored.
+     *
+     * @return a valid document reference
+     */
+    XWikiDocument getDocument();
```
danielpgross: I don't think changing the return type of `DocumentReference getDocument()` (to `XWikiDocument getDocument()`) is a good choice, because it breaks code that already uses that method. The `DocumentReference getDocument()` API has already been in use for a few years (e.g. [here](https://github.com/phenotips/phenotips/blob/61b9c338cb3ee3b2bb064d5b68f9b1affa2946fc/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/ClinicalStatusController.java#L74)), so this change is likely to affect a lot of downstream code. I already encountered code breakage resulting from this change in a downstream Gene42 extension. 

The fact that this PR mostly consists of small changes within PT to make things work with the `getDocument()` API change is a good indication that lots of downstream code will likely need to be tweaked as well. I think the harm of breaking downstream code in this case far outweighs the benefit of introducing a slightly more convenient version of the method.

Fortunately, we can easily maintain backwards compatibility while still adding this more convenient method. How about keeping `DocumentReference getDocument()` and adding a new method `XWikiDocument getXWikiDocument()` instead?

components/entities/api/src/main/java/org/phenotips/entities/internal/AbstractPrimaryEntityGroupManager.java
```diff
@@ -17,77 +17,225 @@
...
+    {
+        this.groupEntityReference = groupEntityReference;
+        this.memberEntityReference = memberEntityReference;
+
+        this.membersManager = (PrimaryEntityManager<E>) getManager(memberEntityReference);
+        this.groupManager = (PrimaryEntityManager<G>) getManager(groupEntityReference);
+    }
+
+    private PrimaryEntityManager<PrimaryEntity> getManager(EntityReference entityReference)
+    {
```
danielpgross: This implementation fails to find `PrimaryEntityManager`s with a type specified in angle brackets, e.g. `PrimaryEntityManager<Family>`. [Here](https://github.com/Gene42/phenotips-niaid/commit/bf58b9a65e7d3443d66951a349f003d99e0f722b#commitcomment-19613681), I needed to specify the role explicitly as the class without the type.

Can this be improved to find `PrimaryEntityManager`s with a type specified in angle brackets?

components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
allasm: I agree that downstream code will be affected, but in a simple search-and-replace manner, as ALL existing `getDocument()`-s return a `documentReference`. PT itself is small, and for now all the derivative code bases are small, and all can just user search-and-replace.

...and it is one of the most confusing things in our APIs, so IMHO we should just bite the bullet and rename everything. As @sergiu explained, we avoided returning XWikiDocument because XWiki was supposed to get rid of this class/refactor it, but it never happened in the last ~5 years, so why not take a chance and make it less confusing now, that we have to refactor a lot of related code anyway. It was a bad decision to start with, and IMHO better to fix it than to stick with it for compatibility reasons, when compatibility can easily be fixed (and it takes about the same amount of work both ways, as @itaiGershtansky already fixed things one way on the PT side)

danielpgross: To clarify, in some cases compatibility _cannot_ easily be fixed. For example, we have provided JARs to a client a few months ago containing a module that has a controller based on the old method. We do not support their servers or provide ongoing updates to their modules. 

If we change the return type of the existing method:
They will see a `NoSuchMethodException` the next time they upgrade PT and will likely ask us why it broke and what needs to be done to fix it, at which time we'll need to step in and help them.

If we create a new method:
None of the above will happen. The module will continue to work as expected.

danielpgross: That is a good point that the controllers need to be changed anyway. I did consider this and I think you are correct. So even though it in this case it is only a theoretical concern, we should be careful going forward about making major changes to widely-used APIs for all of the reasons we discussed in this thread.

As for maintaining custom versions into the future: we are definitely capable of doing this, and for some clients we already do. However, we want to minimize the amount of work required to make old modules work with new versions of PT, because the work that's required to keep these modules up-to-date presents another strain on our already limited resources (at both Gene42 and CCM).

allasm: so, what is the conclusion for now? Just the old API was misleading, and if we have to make a change anyway, I would vote to fix the API _iff_ it is the same amount of work.

I agree about being careful going forward, just if you have to break things for a serious reason, I think it makes sense to fix/improve all the related things (if any) if it is not much more work

danielpgross: Considering that the upgrade would break things anyway, and considering that changing the API would make it easier to understand and use in the future, I'm ok with the return type of `getDocument()` being changed to `XWikiDocument`. So there doesn't need to be any change to this PR.

sdumitriu: -1 for changing the return type, only because this method is also exposed in the old `Patient` interface, which is usable not just from Java code, but also from scripts. While we can detect broken Java code at compile time, there's no check for Velocity.

Another thing to keep in mind: `Patient` is returned by `$services.patients`, but the internal `com.xpn.xwiki.document.XWikiDocument` should __never__ be exposed in scripts, since this one bypasses all permission checks. So `SecurePatientRepository` should return a new `Patient` implementation which returns `null` whenever `getXWikiDocument` is invoked.

allasm: +1 for `SecurePatientRepository` returning a special Patient which returns `null` for `getXWikiDocument()`.

...but, for scripts, do we expect any clients to have velocity code? if yes then I guess we are back to square 0 with keeping the old API, if no we should change those as well, and possibly optimize scripts the same way we have to optimize old uses for `getDocument()` in Java

sdumitriu: It's not just about end-user scripts, but our own scripts. If you search for `patient.document`, it's used in several places in our codebase.

sdumitriu: New idea: deprecate `getDocument` as it is, introduce `getDocumentReference` and `getXDocument`.

danielpgross: +1 for
> deprecate getDocument as it is, introduce getDocumentReference and getXDocument.

components/entities/api/src/main/java/org/phenotips/entities/internal/AbstractPrimaryEntityGroupManager.java
```diff
@@ -17,77 +17,235 @@
...
+        PrimaryEntityManager<PrimaryEntity> manager = null;
+        if (memberEntityReference != null) {
+            ComponentManager cm = ComponentManagerRegistry.getContextComponentManager();
```
danielpgross: This will fail if `ApplicationStartedEvent` has not been sent yet, because the Component Manager Registry is not initialized yet. 

We learned this because we built a document deletion listener utilizing this method that was being initialized on startup, before `ApplicationStartedEvent` was sent. The application was failing to start and throwing Component Manager exceptions. We eventually narrowed down the problem to this line.

@sdumitriu says that the solution should probably be to use the Component Manager injected directly, as follows, instead of using `ComponentManagerRegistry`.
`@Inject @Named("context") private static Provider<ComponentManager> cmProvider;`

allasm:
Anyone remembers what happens with this? I don't think it got merged/cherry-picked?

@danielpgross, I remember Gene42 wanted to have Entities in a more developed form?  @sdumitriu?

danielpgross:
AFAIK this has not been merged or cherry-picked into master.

allasm:
yeah, I randomly checked some places and it is not in master - but did it get used anywhere by Gene42 (e.g. for that new search-using-entities feature)?

danielpgross:
Yes, we did use it there. I can give you more details today at dev meeting.

components/entities/api/src/main/java/org/phenotips/entities/internal/AbstractPrimaryEntityGroupManager.java
allasm: This can also be addressed by lazy initialisation of the manager, since it is not needed until something actually happens, which should be after `ApplicationStartedEvent`

sdumitriu:
Rebased as #2415

PR #2116, opened 2016-10-25
[misc] Dependency optimization

veronikaslc:


allasm:
I think it is a good idea to have this on in before for the "release candidate" version

PR #2117, opened 2016-10-25
PT-2870: Delete dialog looks incomplete on Mac (chrome, safari)

veronikaslc:
Fixed.

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
@@ -106,10 +106,7 @@
   margin-top: 0px;
 }
 .xdialog-box .button {
-  width: 100px;
-}
-.xdialog-box .question {
-  padding: 4px 0 6px 25px;
+  min-width: 5em;
 }
 .push-fields.singlepatient{
   margin-left: 1.5em !important;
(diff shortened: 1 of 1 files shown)


components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
```diff
@@ -110,7 +110,6 @@
 }
 .xdialog-box .question {
```
sdumitriu: The whole ruleset can actually be removed. It's supposed to be applied to one dialog only: from the _browse all patients_ table, select _Push to PhenomeCentral_, then _Next: Select patients_, then _Push to PhenomeCentral_. There should be a dialog asking to confirm that the user wants to push the following patients. The padding override doesn't really change much, and the whitespace doesn't need to be preserved. since all that's displayed is a list of patient identifiers.

marta-: @sdumitriu +1, however I had a discussion with @allasm about removing thaat block and about how the yes/no buttons are too small and difficult to click, and we agreed on setting a `min-width` for buttons (as a better version of the `width: 100px` rule. I suggest something in the lines of `5em'. Wdyt?

sdumitriu: OK, but that's not related to this ruleset or this pull request.

veronikaslc:
codereview comments were adressed

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
allasm: as discussed, i would use padding in addition to min-width, with long-text buttons it becomes too condensed again, IMHO. Also maybe  a bit wider than 5em? I tried and the following seems to look good:

    min-width: 6em;
    padding-right: 1em;
    padding-left: 1em;

PR #2118, opened 2016-10-25
PT-2785: A gene is stored as Ensembl ID if it exist, and as a gene sy…

a-gvo:
…mbol or non-standard ID otherwise.

@buske @marta- Please review.

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
@@ -125,6 +125,7 @@
       geneNames = geneRefs.pluck('value');
     }
     var i = 0;
+    var notmapped = 'notmapped';
     geneRefs.each(function (element) {
       if (element.next() &amp;&amp; element.next(1) &amp;&amp; element.next(1).hasClassName('xHelpButton gene-info'))
           element.next().next('span.xHelpButton').remove();
@@ -137,16 +138,21 @@
         var infoTool = new Element('span', {'class' : 'fa fa-info-circle xHelpButton gene-info', 'title' : geneNames[i]});
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
-        if (isSuggestSelected)
+        if (isSuggestSelected) {
+          element.removeClassName(notmapped);
           element.insert({after: infoTool});
-        else
+        } else if (/^ENSG[0-9]{11}$/.test(geneNames[i])) {
+          element.addClassName(notmapped);
+        } else {
+          element.addClassName(notmapped);
           element.up().insert({top: triangle});
+        }
         ++i;
       }
     });
   }
 
-  var updateInfo = function(el, value, isSuggestSelected) {
+  var updateInfo = function(el, id, value, isSuggestSelected) {
     value = value || el.value;
     if (el.next() &amp;&amp; el.next().className.startsWith('LV_validation_message'))
         el.next().remove();
@@ -162,18 +168,25 @@
     }
     //update genesymbol label for input field
     el.previous().innerHTML = value;
+    // update geneid label for the hidden input field
+    el.next("input.gene-id").value = id || value;
     //update info icon
     fetchGeneDetails([el],[value], isSuggestSelected);
   }
 
-  fetchGeneDetails($$('input.gene-name'), null, true);
+  // Fetch gene details, marking any non-standardized ids.
+  $$('input.gene-name').each(function(elt) {
+    fetchGeneDetails([elt], null, !elt.className.endsWith('notmapped'));
+  });
 
   $$('input.gene-name').invoke('observe', "ms:suggest:selected", function (event) {
     fetchGeneDetails([event.findElement()], event.memo &amp;&amp; [event.memo.value] || [], true);
+    // Update the id element.
+    event.element().next("input.gene-id").value = event.memo.id || event.memo.value;
   });
 
   $$('input.gene-name').invoke('observe', 'change', function (event) {
-    updateInfo(event.element(), null, false);
+    updateInfo(event.element(), null, null, false);
   });
 
   document.observe("xwiki:dom:updated", function(event) {
@@ -183,10 +196,10 @@
           fetchGeneDetails(elt.select('input.gene-name'), false);
           var inputGene = elt.select('input.gene-name');
           inputGene.invoke('observe', "ms:suggest:selected", function (event) {
-            updateInfo(event.element(), event.memo.value, true);
+            updateInfo(event.element(), event.memo.id, event.memo.value, true);
           });
           inputGene.invoke('observe', 'change', function (event) {
-            updateInfo(event.element(), null, false);
+            updateInfo(event.element(), null, null, false);
           });
         }
       });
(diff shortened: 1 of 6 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
```diff
@@ -162,18 +166,29 @@
...
   }
 
-  fetchGeneDetails($$('input.gene-name'), null, true);
+  // Fetch gene details, marking any non-standardized ids.
+  $$('input.gene-name').each(function(elt) {
+    if (!elt.className.endsWith('nonstandard')) {
+      fetchGeneDetails([elt], null, true);
+    } else {
+      fetchGeneDetails([elt], null, false);
+    }
```
marta-: The last 5 lines could be replaced with `fetchGeneDetails([elt], null, !elt.className.endsWith('nonstandard'));`

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/VocabularyTerm.java
```diff
@@ -44,6 +44,13 @@
     String getId();
 
     /**
+     * The preferred term identifier for internal storage, for example {@code MIM: 260540} or {@code ENSG00000230986}.
+     *
+     * @return the term identifier, or {@code null} if the term does not have an associated identifier
+     */
+    String getUnderlyingId();
```
marta-: I am not a fan of this method name. This is the kind of method that should speak for itself, but I wouldn't know what it does without reading the javadoc. Suggestions: `getStoredId`, `getPreferredId`. @sdumitriu thoughts?

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
```diff
@@ -130,6 +130,17 @@ public String getId()
     }
 
     @Override
+    public String getUnderlyingId()
+    {
+        if ("hgnc".equals(this.ontology.getIdentifier())) {
+            @SuppressWarnings("unchecked")
+            List<String> ensemblList = (List<String>) this.get("ensembl_gene_id");
+            return ensemblList != null && !ensemblList.isEmpty() ? ensemblList.get(0) : (String) this.get("symbol");
```
buske: I believe the SolrVocubularyTerm should be abstracted from the specific vocabularies as much as possible.

marta-:
I entered SRCAP without selecting it from the dropdown. It is shown to the user as standard in view mode (i.e. the info button popover displays the right Ensembl id), however in the database it is stored as SRCAP, not ENSG00000080603. Is that intentional? @a-gvo @buske

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/VocabularyTerm.java
a-gvo: getPreferredId is probably the most self-explanatory one?

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/SolrVocabularyTerm.java
```diff
@@ -130,6 +130,17 @@ public String getId()
     }
 
     @Override
+    public String getUnderlyingId()
+    {
+        if ("hgnc".equals(this.ontology.getIdentifier())) {
```
sdumitriu: I don't like this hardcoding here.

a-gvo:
@marta- This is not an issue in PT-2854, where selection on loss of focus is implemented. I considered adding an additional check for this on patient load, but it would probably be redundant? What do you think?

components/vocabularies/hgnc/ui/src/main/resources/PhenoTips/GeneNameService.xml
```diff
@@ -57,6 +57,7 @@
       #set ($term = $services.vocabularies.hgnc.getTerm($i))
       #if ($term)
         #set ($result = {
+          'id' : $term.underlyingId,
```
sdumitriu: Since we're manually building this, you can just use `'id': $stringutils.defaultIfBlank($term.ensembl_gene_id, $term.symbol)` and get rid of the `getUnderlyingId` method completely.

components/vocabularies/hgnc/ui/src/main/resources/PhenoTips/GeneSymbolDisplayer.xml
```diff
@@ -0,0 +1,71 @@
...
+    {{html wiki="false" clean="false"}}&lt;input type='text' name='${prefix}symbol' class='suggested suggest-gene gene-name $!termStatus' placeholder='$placeholder' size='16' value='$!{gene.symbol}'/&gt;&lt;input type='hidden' name='${prefix}${name}' class='gene-id' value='$!{gene.id}'/&gt;{{/html}}
+  #else
+    ${gene.symbol}#if("$!{gene.symbol}" != ''){{html wiki="false" clean="false"}}&lt;input type='hidden' name='gene_name' class='gene-name' value="$!{gene.symbol}"/&gt;{{/html}}#end
+  #end
+#end
+##
+##
+#macro (_validateGeneId $geneData $termStatus)
+  #if (!$geneData.id.matches('^ENSG[0-9]{11}$'))
+    #set ($termStatus = "nonstandard")
```
sdumitriu: I think this one should be set regardless of the format of the identifier, as long as the term is unknown.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
```diff
@@ -137,16 +138,19 @@
...
         var icon_helpController = new PhenoTips.widgets.HelpButton(infoTool);
         var triangle = new Element('span', {'class': "fa fa-exclamation-triangle", title: '$services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardGenesymbol')'}).update(' ');
-        if (isSuggestSelected)
+        if (isSuggestSelected) {
+          element.removeClassName(nonstandard);
           element.insert({after: infoTool});
-        else
+        } else {
+          element.addClassName(nonstandard);
           element.up().insert({top: triangle});
```
buske: If they type something that looks like an Ensembl ID, I think it should validate even if they didn't select it from the suggestion list.

components/vocabularies/hgnc/ui/src/main/resources/PhenoTips/GeneSymbolDisplayer.xml
marta-: I overheard Alina and Orion talking about this, and how anything that matches the Ensembl id should be accepted, was it was because HGNC is not always updated? @buske please jump in with more details.

marta-:
> @marta- This is not an issue in PT-2854, where selection on loss of focus is implemented. I considered adding an additional check for this on patient load, but it would probably be redundant? What do you think?

@a-gvo I checked out PT-2854 and it works as expected. Thanks!

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene-variants.xml
marta-: @buske OK, but are we all fine with the popover that shows up when clicking the info button for such an entry? (See screenshot below)
 
ensgxxx

marta-: I'm fine not to show a warning sign, but imo we should not show an info button either @buske

a-gvo: Fixed :). No info button on Ensembl IDs that are not in vocab.

marta-:
@buske @sdumitriu : @a-gvo made changes addressing your comments, please review again. Thanks!

PR #2119, opened 2016-10-25
PT-2777: Create a migrator from gene symbols to Ensembl IDs

a-gvo:
@veronikaslc @allasm Please review

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
@@ -271,13 +271,13 @@ public class GeneListController extends AbstractComplexController<Map<String, St
             JSONObject solvedGene = json.optJSONObject(JSON_SOLVED_KEY);
 
             List<Map<String, String>> accumulatedGenes = new LinkedList<>();
-            List<String> geneSymbols = new ArrayList<>();
+            List<String> geneNames = new ArrayList<>();
 
-            parseGenesJson(genesJson, geneSymbols, accumulatedGenes, enumValues);
+            parseGenesJson(genesJson, geneNames, accumulatedGenes, enumValues);
 
             // 1.2.* json compatibility
-            parseRejectedGenes(rejectedGenes, geneSymbols, accumulatedGenes);
-            parseSolvedGene(solvedGene, geneSymbols, accumulatedGenes);
+            parseRejectedGenes(rejectedGenes, geneNames, accumulatedGenes);
+            parseSolvedGene(solvedGene, geneNames, accumulatedGenes);
 
             return new IndexedPatientData<>(getName(), accumulatedGenes);
         } catch (Exception e) {
@@ -286,7 +286,7 @@ public class GeneListController extends AbstractComplexController<Map<String, St
         return null;
     }
 
-    private void parseGenesJson(JSONArray genesJson, List<String> geneSymbols, List<Map<String, String>> allGenes,
+    private void parseGenesJson(JSONArray genesJson, List<String> geneNames, List<Map<String, String>> allGenes,
         Map<String, List<String>> enumValues)
     {
         if (genesJson != null) {
@@ -295,7 +295,7 @@ public class GeneListController extends AbstractComplexController<Map<String, St
 
                 // discard it if gene symbol is not present in the geneJson, or is whitespace, empty or duplicate
                 if (!geneJson.has(JSON_GENE_KEY) || StringUtils.isBlank(geneJson.getString(JSON_GENE_KEY))
-                    || geneSymbols.contains(geneJson.getString(JSON_GENE_KEY))) {
+                    || geneNames.contains(geneJson.getString(JSON_GENE_KEY))) {
                     continue;
                 }
 
@@ -310,12 +310,12 @@ public class GeneListController extends AbstractComplexController<Map<String, St
                 }
 
                 allGenes.add(singleGene);
-                geneSymbols.add(geneJson.getString(JSON_GENE_KEY));
+                geneNames.add(geneJson.getString(JSON_GENE_KEY));
             }
         }
     }
 
-    private void parseRejectedGenes(JSONArray rejectedGenes, List<String> geneSymbols,
+    private void parseRejectedGenes(JSONArray rejectedGenes, List<String> geneNames,
         List<Map<String, String>> allGenes)
     {
         if (rejectedGenes != null && rejectedGenes.length() > 0) {
@@ -325,7 +325,7 @@ public class GeneListController extends AbstractComplexController<Map<String, St
                 // discard it if gene symbol is not present in the geneJson, or is whitespace, empty or duplicate
                 if (!rejectedGeneJson.has(JSON_GENE_KEY)
                     || StringUtils.isBlank(rejectedGeneJson.getString(JSON_GENE_KEY))
-                    || geneSymbols.contains(rejectedGeneJson.getString(JSON_GENE_KEY))) {
+                    || geneNames.contains(rejectedGeneJson.getString(JSON_GENE_KEY))) {
                     continue;
                 }
                 Map<String, String> singleGene = new LinkedHashMap<>();
@@ -338,16 +338,16 @@ public class GeneListController extends AbstractComplexController<Map<String, St
                 }
 
                 allGenes.add(singleGene);
-                geneSymbols.add(rejectedGeneJson.getString(JSON_GENE_KEY));
+                geneNames.add(rejectedGeneJson.getString(JSON_GENE_KEY));
             }
         }
     }
 
-    private void parseSolvedGene(JSONObject solvedGene, List<String> geneSymbols, List<Map<String, String>> allGenes)
+    private void parseSolvedGene(JSONObject solvedGene, List<String> geneNames, List<Map<String, String>> allGenes)
     {
         if (solvedGene != null && solvedGene.has(JSON_GENE_KEY)
             && !StringUtils.isBlank(solvedGene.getString(JSON_GENE_KEY))
-            && !geneSymbols.contains(solvedGene.getString(JSON_GENE_KEY))) {
+            && !geneNames.contains(solvedGene.getString(JSON_GENE_KEY))) {
             Map<String, String> singleGene = new LinkedHashMap<>();
             singleGene.put(INTERNAL_GENE_KEY, solvedGene.getString(JSON_GENE_KEY));
             singleGene.put(INTERNAL_STATUS_KEY, INTERNAL_SOLVED_KEY);
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
@@ -79,7 +79,7 @@ public class VariantListController extends AbstractComplexController<Map<String,
 
     private static final String INTERNAL_VARIANT_KEY = "cdna";
 
-    private static final String INTERNAL_GENESYMBOL_KEY = "genesymbol";
+    private static final String INTERNAL_GENE_KEY = "gene";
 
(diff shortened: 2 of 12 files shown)


marta-:
Replacing @allasm with @itaiGershtansky to balance review workload.

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R71498PhenoTips2777DataMigration.java
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    public String getDescription()
+    {
+        return "Migrate all existing gene values from HGNC symbols to Ensembl IDs";
+    }
+
+    @Override
+    public XWikiDBVersion getVersion()
+    {
+        return new XWikiDBVersion(71498);
```
sdumitriu: This one is already in use. Try `71499`.

a-gvo: Fixed.

a-gvo:
Amended to account for variants. @itaiGershtansky @veronikaslc please review again.

components/patient-data/ui/src/main/resources/PhenoTips/GeneVariantClass.xml
```diff
@@ -66,19 +66,19 @@ $!{value}
       <validationRegExp/>
       <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
     </cdna>
-    <genesymbol>
+    <gene>
```
marta-: @buske Just a heads-up about this change in the name of the field that references the gene in GeneVarientClass, which may break some PhenomeCentral functionality (e.g. https://github.com/phenotips/patient-network/blob/cb3b482e33b70ac8ccb63ccafd937fb6b9c9cf5d/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java#L150 ).

buske: Thank you for the heads-up!

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -78,7 +78,7 @@
 
     private static final String INTERNAL_VARIANT_KEY = "cdna";
 
-    private static final String INTERNAL_GENESYMBOL_KEY = "genesymbol";
+    private static final String INTERNAL_GENESYMBOL_KEY = "gene";
```
buske: These are no longer gene symbols, so the name of this constant should be changed.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/VariantListControllerTest.java
```diff
@@ -88,7 +88,7 @@
 
     private static final String VARIANT_KEY = "cdna";
 
-    private static final String GENESYMBOL_KEY = "genesymbol";
+    private static final String GENESYMBOL_KEY = "gene";
```
buske: These are no longer gene symbols, so the name of this constant should be changed.

a-gvo:
@buske I changed the naming as you requested (I tried to find all such occurrences).

components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm
```diff
@@ -173,15 +173,15 @@ $doc.display($prop.getName(), 'view', $fakeObj )##
 #set($fakeObj = $doc.newObject($geneClassName))
 #set($geneNumber = 'ZZGENE_NUMBER_PLACEHOLDERZZ')
 #set($geneObjNumber = 'ZZGENE_INDEX_PLACEHOLDERZZ')
-#set($genesymbol = 'ZZGENE_SYMBOL_PLACEHOLDERZZ')
+#set($geneName = 'ZZGENE_SYMBOL_PLACEHOLDERZZ')
```
buske: I would update this to `ZZGENE_ID_PLACEHOLDERZZ`

a-gvo: Fixed.

components/patient-data/impl/src/test/java/org/phenotips/data/internal/controller/VariantListControllerTest.java
```diff
@@ -470,7 +470,7 @@ public void writeJSONAddsContainerWithOnlySelectedFields() throws ComponentLooku
 
         Map<String, String> item = new LinkedHashMap<>();
         item.put(VARIANT_KEY, "variantName");
-        item.put(GENESYMBOL_KEY, "geneSymbol");
+        item.put(GENE_KEY, "geneSymbol");
```
sdumitriu: Shouldn't the value be updated as well? I see two other places in this file with the same problem.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/VocabularyManager.java
```diff
@@ -44,6 +44,16 @@
...
 
     /**
+     * Retrieve a term given its owner vocabulary. The owner vocabulary's id is specified.
+     * @param vocabularyId the identifier for the owner vocabulary, for example {@code HGNC}
+     * @param termId the term identifier, in the format {@code <vocabulary prefix>:<term id>}, for example
+     *               {@code HP:0002066}
+     * @return the requested term, or {@code null} if the term doesn't exist in the vocabulary, or no matching
+     *         vocabulary is available
+     */
+    VocabularyTerm resolveTerm(String vocabularyId, String termId);
```
sdumitriu: I'd rather not add a new method here, for API stability. It's easy to just use this instead:

    final VocabularyTerm term = vocabularies.getVocabulary(HGNC).getTerm(geneSymbol);

Or, instead of working with the vocabulary manager, you can just do:

    @Inject @Named('hgnc') private Vocabulary hgnc;
    ....
    final VocabularyTerm term = hgnc.getTerm(geneSymbol);

a-gvo: I tried the latter before, and it seems to result in component look-up failure when building. The vocabulary manager works well though.

a-gvo:
@sdumitriu Fixed

marta-:
Just had a chat with @allasm and he pointed out that we also need to migrate all the pedigree JSONs to store Ensemble ids instead of gene names. This, together with [PT-2942](https://phenotips.atlassian.net/browse/PT-2942) and [PT-2929](https://phenotips.atlassian.net/browse/PT-2929) should cover everything wrt genes in the pedigree editor.

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R71499PhenoTips2777DataMigration.java
```diff
@@ -0,0 +1,250 @@
...
+    /**
+     * Gets EnsemblID corresponding to the HGNC symbol.
+     * @param geneSymbol the string representation of a gene symbol (e.g. NOD2).
+     * @return the string representation of the corresponding Ensembl ID.
+     */
+    private String getEnsemblId(final String geneSymbol) {
+        final VocabularyTerm term = this.hgnc.getTerm(geneSymbol);
+        @SuppressWarnings("unchecked")
+        final List<String> ensemblIdList = term != null ? (List<String>) term.get("ensembl_gene_id") : null;
+        final String ensemblId = ensemblIdList != null && !ensemblIdList.isEmpty() ? ensemblIdList.get(0) : null;
```
allasm: is get(0) always guaranteed to be the right term? E.g. is the first element of the results array always the correct choice? What if there is NO exact match, and the first is just the least inaccurate match? We may have free-text genes (right?) which have no exact match but do have some partial match

E.g. http://localhost:8080/get/PhenoTips/GeneNameService?outputSyntax=plain&rand=0.38996313003383465&json=true&q=GS

returns stuff, but none of it is the correct ensemble ID

i think the correct thing to do is to compare the `symbol`, if the match is exact use the ID, otherwise keep free text as is

allasm:
I think here is another very important problem with EnsembleIDs not related to migrations.

Since we changed patient JSON, push will now be broken when pushing to an older instance, e.g. to PhenomeCentral (or from an older instance for that matter). Since PC is the most important instance and pushing is important, I think we should fix that as well before releasing.

Two possible fixes are:
1) (correct complete fix) make a new push protocol version and add the change in gene representations to the list of incompatibilities between versions, like it was done for dates

See PT-2643 / [https://github.com/phenotips/phenotips/blob/master/components/patient-data-sharing/share-protocol/src/main/java/org/phenotips/data/shareprotocol/ShareProtocol.java] 

2) (quick easy fix) temporarily revert back patient JSON change, so that patientJSON generated still uses gene symbols (and patient JSON importer can read gene symbols).

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R71499PhenoTips2777DataMigration.java
buske: This is a very good point. A lot of the candidate genes are garbage, e.g., `Microarray` or `Joubert panel`

a-gvo: @allasm From what I understand, ```ensemblIdList.get(0)``` should be right. 
This: ```final VocabularyTerm term = this.hgnc.getTerm(geneSymbol);``` gets the vocabulary term that has ```geneSymbol``` as its ID or as one of its aliases (or null if it doesn't exist). If the term exists, it may have one or more Ensembl IDs associated with it. I'm just taking the first one. Is that what you meant, or am I misunderstanding?

For GeneNameService, your example actually uses the ```search```, not the ```getTerm``` method, which does return a result array of possible terms.

veronikaslc: [PT-2942](https://phenotips.atlassian.net/browse/PT-2942) is taking care of it

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -271,13 +271,13 @@ private JSONObject assebleGene(Map<String, String> internalToJSONkeys, Map<Strin
             JSONObject solvedGene = json.optJSONObject(JSON_SOLVED_KEY);
 
             List<Map<String, String>> accumulatedGenes = new LinkedList<>();
-            List<String> geneSymbols = new ArrayList<>();
+            List<String> geneNames = new ArrayList<>();
 
-            parseGenesJson(genesJson, geneSymbols, accumulatedGenes, enumValues);
+            parseGenesJson(genesJson, geneNames, accumulatedGenes, enumValues);
 
             // 1.2.* json compatibility
```
allasm: version 1.2 does not use ensemle IDs, so I think there should be a translation layer which first attempts to convert symbols to IDs (since in most cases it is possible)

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R71499PhenoTips2777DataMigration.java
allasm: OK, this is not an issue indeed then

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/VocabularyManager.java
allasm: BTW, just curious, what is the overall design, what is an "owner vocabulary", e.g. when is someone supposed to use `resolveTerm(String termId)` and when `resolveTerm(String vocabularyId, String termId)`?

allasm:
This pull request has been superseded by PT-2012 branch, which will have a pull request of its own

PR #2121, opened 2016-10-27
PT-2833: "Affected" and the first "Confirmed gene" use the exact same…

marta-:
… color in the pedigree

Removed the first two predefined colors, which are shades of yellow. It's unlikely that there will be that many confirmed causal genes in real life anyway.


marta-:
@allasm please let me know if you agree.

PR #2124, opened 2016-10-28
[misc] Correct Data Tables view

veronikaslc:


veronikaslc:
@marta- I found out that stylesheet extension in `DBConfigurationClass` was configured to be `on demand only` but was not imported. In this PR I was trying to guesse how Data Tables were intended to look.

components/application-configuration/ui/src/main/resources/PhenoTips/DBConfigurationClass.xml
```diff
@@ -103,7 +107,7 @@ $!value
 {{/velocity}}
 
 {{velocity}}{{html wiki=false clean=false}}
-&lt;ul style="list-style-type: none; margin: 0 0 0 16px"&gt;
+&lt;ul class="custom-display-list col-2"&gt;
```
marta-: I'm not a fan of the result of adding the `col-2` css class. Users have been complaining about our two column layouts and how they're really difficult to follow. I'd suggest removing it from [DBConfigurationClassDisplayer](https://github.com/phenotips/phenotips/blob/dd204fcf8fc9c7f4b32c34886a8c38740f288bd1/components/application-configuration/ui/src/main/resources/PhenoTips/DBConfigurationClassDisplayer.xml#L93) as well.

veronikaslc:
@marta- amended commit

components/navigation/ui/src/main/resources/data/WebHome.xml
```diff
@@ -119,6 +119,10 @@ $!value
         <cache>0</cache>
         <customDisplay>**{{translation key="phenotips.DBConfigurationClass.metaInfo" /}}**
 
+{{velocity}}
+$xwiki.ssx.use('PhenoTips.DBConfigurationClass')##
```
marta-: this can go in the velocity block below, no need for a new block.

PR #2126, opened 2016-10-31
NIAID-79: added support for absolute paths

sebialex:
@sdumitriu @itaiGershtansky This is needed for the entities search rest api for niaid. Please review.

### DIFF
--- a/components/base-war/src/main/webapp/templates/macros.vm
+++ b/components/base-war/src/main/webapp/templates/macros.vm
@@ -1418,7 +1418,13 @@ ${Msz}Mb##
     #if ("$!resultPage" == '')
       #set ($resultPage = 'XWiki.LiveTableResults')
     #end
-    #set ($dataurl = $xwiki.getURL($resultPage, 'get', "$escapetool.url($parameters)&$!options.extraParams"))
+    #if ($resultPage.startsWith('/'))
+       #set ($hostname = "$request.getServerName()")
+       #set ($hostport = "$request.getServerPort()")
+       #set ($dataurl = "http://$hostname:$hostport$resultPage?$escapetool.url($parameters)&$!options.extraParams")
+    #else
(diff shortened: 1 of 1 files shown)


components/base-war/src/main/webapp/templates/macros.vm
```diff
@@ -1418,7 +1418,13 @@ ${Msz}Mb##
     #if ("$!resultPage" == '')
       #set ($resultPage = 'XWiki.LiveTableResults')
     #end
-    #set ($dataurl = $xwiki.getURL($resultPage, 'get', "$escapetool.url($parameters)&$!options.extraParams"))
+    #if ($resultPage.startsWith('/'))
+       #set ($hostname = "$request.getServerName()")
+       #set ($hostport = "$request.getServerPort()")
+       #set ($dataurl = "http://$hostname:$hostport$resultPage?$escapetool.url($parameters)&$!options.extraParams")
```
sdumitriu: You don't need to make it an absolute URL, just let it start with `/`.

For example, this is wrong since it may use the wrong protocol.

marta-: Wdym? this line shoulf simply read `#set ($dataurl = "$resultPage?$escapetool.url($parameters)&$!options.extraParams")`

marta-: What @sdumitriu is saying is basically:
* If your `$resultsPage` is already a relative url (starting with `/`, e.g. `/rest/entities`), then just use that relative url with the appropriate parameters
* If `$resultsPage` is a document reference, build the relative url of the document and append the parameters.

```
#if ($resultPage.startsWith('/'))
  #set ($hostname = "$request.getServerName()")
  #set ($hostport = "$request.getServerPort()")
  #set ($dataurl = "$resultPage?$escapetool.url($parameters)&$!options.extraParams")
#else
  #set ($dataurl = $xwiki.getURL($resultPage, 'get', "$escapetool.url($parameters)&$!options.extraParams"))
#end
```

PR #2128, opened 2016-11-01
PT-2871: Delete patient and family functionality is broken

marta-:
Fixed 3 problems:
- `deleteFamily` workflow was wrong when the deletion of all members was requested
- Family Id got "lost" before actually sending the request to delete the family
- CSRF token was not sent from client side for the `deletefamily` action


marta-:

PR #2129, opened 2016-11-02
PT-2896: Switching to another language makes name fields hidden in th…

marta-:
…e pedigree node form

Fixed. Made the code more generic, and as a positive "side effect" now other fields (date of birth and death, sex, gestational age, phenotypes, disorders) will be disabled in the pedigree editor menu if they are disabled if the PhenoTips patient form.  Genes cannot be disabled yet because of an issue with the `RecordConfiguration` implementation (`getAllFieldNames()` and `getEnabledFieldNames()` only return fields from the `PatientClass`, and genes are stored separately) .

### DIFF
--- a/components/pedigree/ui/src/main/resources/PhenoTips/PedigreeInterface.xml
+++ b/components/pedigree/ui/src/main/resources/PhenoTips/PedigreeInterface.xml
@@ -47,14 +47,26 @@
 #end
 #macro(__getDisabledFields)
   #set ($recordConfiguration = $services.recordConfiguration.getActiveConfiguration())
-  #set ($disabled = ["first_name","last_name","last_name_birth"])
-  #foreach ($section in $recordConfiguration.enabledSections)
-    #foreach ($element in $section.enabledElements)
-      #if ($element == "Patient name")
-        #set ($disabled = [])
-        #break
-      #end
-    #end
+  #set ($disabledFieldNames = [])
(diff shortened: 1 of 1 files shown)


marta-:

components/pedigree/ui/src/main/resources/PhenoTips/PedigreeInterface.xml
```diff
@@ -47,13 +47,22 @@
...
+  #set ($linkedFields = {
+    "last_name": ["last_name_birth"],
+    "phenotype": ["hpo_positive"],
+    "negative_phenotype" : ["hpo_negative"],
+    "omim_id" : ["disorders"],
+    "gestation" : ["gestation_age"],
+    "gene" : ["candidate_genes", "causal_genes", "rejected_genes"]
+  })
+  #set ($disabled = [])
+  #set ($discard = $disabled.addAll($disabledFieldNames))
```
allasm: as a side effect `disabled` will contain all disabled field names as configured in the patient form, e.g. `phenotype`, which are not the fields we want to disable in pedigree.

It does not hurt at the moment, but in theory it may overlap with a totally different field in pedigree (well, that is a bit stretching it, but still this is not a clean conversion). We should either rename pedigree fields, or have a clean conversion or - hopefully - get rid of pedigree node menu and use the same UI. Other than that, +1

marta-:
@allasm your comment was addressed in the latest commit

allasm:
+1

PR #2130, opened 2016-11-02
PT-2155 (old family studies migration) and PT-2890, PT-2904

allasm:
Fixed and re-implemented old family studies migrator + supporting code


allasm:
@veronikaslc @itaiGershtansky please review

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_pedigree.xml
```diff
@@ -117,7 +117,7 @@
   #end
 #end
 #if ("$!{pedigreeHtml}" != '')
-  (% id='pedigree' class='chapter clinical-info pedigree-family-section' %)(((
+  (% id='pedigree' class='clinical-info pedigree-family-section' %)(((
```
marta-: This causes a merge conflict because ` FamilyRecordField_pedigree.xml` has been changed. Please update.

components/family-studies/ui/src/main/resources/PhenoTips/FamilySheet.xml
```diff
@@ -115,7 +116,7 @@
     #end
   #end
   #if ("$!{pedigreeHtml}" != '')
-    (% id='pedigree' class='chapter clinical-info pedigree-family-section' %)(((
+    (% id='pedigree' class='clinical-info pedigree-family-section' %)(((
```
marta-: This code is actually commented out and should have been removed.The change has no effect.

components/pedigree/resources/src/main/resources/pedigree/model/import.js
```diff
@@ -705,6 +705,7 @@ define([
      *                            First a match versus an existing ID is checked, if not found a check against "externalId",
      *                            if not found a check against "name" and finally "firstName".
      *                            If one of the parents is given and the other one is not a virtual new node is created
+     *   - "phenotipsId": The id of the PhenoTips document this node is linked to (default: none)
```
veronikaslc: why not `documentId`?

allasm:
rebased on master and addressed comments (cancelled some of the FamilySheet changes + better comment for aunt_uncle and niece_nephew)

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__family.xml
```diff
@@ -113,26 +228,30 @@
...
 
 {{velocity}}
+$xwiki.ssx.use('PhenoTips.UIX_Field__family')##
+
 #set ($targetClassName = 'PhenoTips.RelativeClass')
 #if ($xcontext.action == 'edit' || $doc.getObject($targetClassName))
-  {{error}}This component is deprecated. Please specify relevant family relationships in the pedigree.{{/error}}
 
   (% class="clear relatives-info" %)(((
+    {{error}}This component is deprecated. Please specify relevant family relationships in the pedigree.{{/error}}
```
veronikaslc: should be translated?

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/PhenotipsFamilyMigrations.java
```diff
@@ -104,38 +104,39 @@
...
+            this.setCollaborators(newFamilyXDocument, patientXDoc);
+
+            newFamilyXDocument.setAuthorReference(patientXDoc.getAuthorReference());
+            newFamilyXDocument.setCreatorReference(patientXDoc.getCreatorReference());
+            newFamilyXDocument.setContentAuthorReference(patientXDoc.getContentAuthorReference());
+            newFamilyXDocument.setParentReference(this.familyParentReference);
+
+            return newFamilyXDocument;
+        } catch (Exception ex) {
+            return null;
```
veronikaslc: adding log to show `IllegalArgumentException("The new family id was already taken.")` would be bonus, otherwise won't be seen

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/AbstractBasePedigree.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param data pedigree data
+     * @param image SVG 'image'
+     */
+    public AbstractBasePedigree(JSONObject data, String image)
+    {
+        if (data == null || data.length() == 0) {
+            throw new IllegalArgumentException();
+        }
+        this.data = data;
+        this.image = image;
```
allasm: not sure, probably not - when pedigree will be imported via REST there will be pedigree data but no image (yet)

..at that point we will have to make sure nothing crashes in that case downstream, but having no image is fine in itself

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/NewFormatPedigree.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Checks that the provided JSON objects represents a pedigree in the supported format.
+     *
+     * @param data JSON object ot check
+     * @return true if JSON represents a pedigree format supported by this class
+     */
+    public static boolean isSupportedPedigreeFormat(JSONObject data)
+    {
+        if (data.optJSONArray(PEDIGREE_JSON_DATA_KEY) != null) {
```
allasm: +1

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__family.xml
allasm: doesn't matter, component is dperecated and will only be enabled temporarily to check migrations. Marta added it without translation so I'm fine with it for sure

components/pedigree/resources/src/main/resources/pedigree/model/import.js
allasm: good point, but not sure what is better. DocumentId may give ideas that you can link to some other documents, which wont work correctly (since this link only works correctly in pedigree when it links to a specifically PT document)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/NewFormatPedigree.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+        return Pair.of(null, null);
+    }
+
+    @Override
+    protected void removeLinkFromPedigreeJSON(String linkedPatientId)
+    {
+        JSONArray nodeList = data.optJSONArray(PEDIGREE_JSON_DATA_KEY);
+        if (nodeList != null) {
```
veronikaslc: shell we check ` && nodeList.length() > 0` as well?

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/PhenotipsFamilyMigrations.java
```diff
@@ -176,17 +181,78 @@ public JSONObject processPedigree(JSONObject data, String patientId)
...
+     * @param commentAddition text of comment addition
+     */
+    public void updatePedigreeComment(JSONObject pedigree, String patientId, String commentAddition)
+    {
+        JSONArray gg = pedigree.getJSONArray("GG");
+        for (int pedigreeId = 0; pedigreeId < gg.length(); pedigreeId++) {
+            JSONObject node = gg.getJSONObject(pedigreeId);
+
+            JSONObject properties = node.optJSONObject("prop");
+            if (properties == null) {
```
veronikaslc: shell we check `|| properties.length() == 0` as well?

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/NewFormatPedigree.java
allasm: we don't, why bother, if length() == 0 the for loop simply will execute 0 times

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/PhenotipsFamilyMigrations.java
allasm: as above, there is no difference for us if legth is 0 or not 0, even if it is not 0 it may not contain what we are looking for, and that is taken care for below. I'm not sure JSONObject even has length() (unlike JSONArray)

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71498PhenoTips2155DataMigration.java
```diff
@@ -132,264 +202,638 @@ public Object doInHibernate(Session hSession) throws HibernateException, XWikiEx
...
-            if (relativesFamilyRef == null || patientFamilyRef != null && StringUtils.isNotBlank(relativesFamilyRef)) {
-                this.migrator.logger.debug("More than one family exists for patient and relatives. Patient Id: {}.",
-                    patientXDocument.getId());
+            try {
+                XWikiDocument familyXDocument =
+                        this.context.getWiki().getDocument(this.migrator.resolver.resolve(familyId), this.context);
+
+                BaseObject pedigreeObj = familyXDocument.getXObject(Pedigree.CLASS_REFERENCE);
+                if (pedigreeObj == null) {
+                    this.migrator.logger.error("Error updating existing family: no pedigree object for {}", familyId);
```
veronikaslc: Is it an error? in same situation in `importPedigreeToFamily()` we have `this.migrator.logger.debug("Patient does not have pedigree...`

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71498PhenoTips2155DataMigration.java
```diff
@@ -132,264 +202,638 @@ public Object doInHibernate(Session hSession) throws HibernateException, XWikiEx
...
+                Map<String, String> relativesToLink, Map<String, String> relativesToMention)
+        {
+            try {
+                String pedigreeCommentAddition = this.generatePedigreeCommentForRelatives(relativesToMention);
+
+                // Generate pedigree based on relativesToLink
+                JSONObject pedigree = this.generatePedigree(probandId, probandData, relativesToLink,
+                        pedigreeCommentAddition);
+
+                // generat enew family document and set llinks
```
veronikaslc: some typos in comment line

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71498PhenoTips2155DataMigration.java
allasm: it is an error. Should be an error there as well (but that was not done by me for this issue, so I did not notice/did not know)

allasm: well, in a way in that other case it can be considered not a big error, since if there is no pedigree there is no thing to migrate, end of story (for that other issue)

In our case, the fact that there supposedly is a pedigree changes how we migrate other data, and it might not get migrated properly if pedigree is broken. So it is OK to have `debug` there and we do need an `error` here

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_pedigree.xml
```diff
@@ -136,4 +136,117 @@ fields=warning,warning_message</parameters>
...
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>.pedigree-wrapper {
+  margin-top: 1em;
```
veronikaslc: Looks strange when to have a CSS property we need 110 lines. In both `PhenoTips.UIX_Field__family` and `FamilyRecordField_pedigree` we create a StyleSheetExtension to hold one class.

allasm: Valid point, but I think this CSS logically belongs here, and sinc eit is already there developer time is more valuable than removing extra 100 lines which are not wrong per se :)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/NewFormatPedigree.java
```diff
@@ -0,0 +1,120 @@
...
+            for (Object nodeObj : nodeList) {
+                JSONObject node = (JSONObject) nodeObj;
+                String nodeLinkId = node.optString(NewFormatPedigree.PATIENT_LINK_JSON_KEY);
+                if (StringUtils.equals(linkedPatientId, nodeLinkId)) {
+                    node.remove(NewFormatPedigree.PATIENT_LINK_JSON_KEY);
```
allasm: No, the pedigree may not make sense without the node (it may become disconnected, or there may be no partner in a relationship - which is not supported yet. ON top of that it is extremely difficult to do from Java)

..and even if we can solve the all the problems, it still does not make sense e.g. a person (almost) always has two parents, a mother and a father. The fact that father patient profile is gone or unlinked does not remove the fact that there is still a father, which should be recorded in the pedigree

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_pedigree.xml
marta-: FWIW I agree with @veronikaslc . Can't we put this in [PedigreeMacros](https://github.com/phenotips/phenotips/blob/master/components/pedigree/ui/src/main/resources/PhenoTips/PedigreeMacros.xml#L542)? Does it have to be different on family page and patient page?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__family.xml
```diff
@@ -38,6 +38,121 @@
...
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>.relatives-info {
```
sdumitriu: Since this is deprecated, why bother adding new code to it?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__family.xml
```diff
@@ -38,6 +38,121 @@
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
   <content/>
+ <object>
```
sdumitriu: Run `mvn xar:format` before committing xdoc sources.

components/family-studies/ui/src/main/resources/PhenoTips/FamilySheet.xml
```diff
@@ -41,6 +41,7 @@
 
 {{velocity}}
 $xwiki.ssx.use('PhenoTips.PatientSheetCode', {'colorTheme' : "$xwiki.getSpacePreference('colorTheme')"})##
+$xwiki.ssx.use('PhenoTips.FamilyRecordField_pedigree')##
```
sdumitriu: This should be in `PhenoTips.FamilyRecordField_pedigree`, since it only applies to content generated by that extension. That is, if the SSX survives the review.

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_pedigree.xml
sdumitriu: You can add the following in `PedigreeMacros`:

```
h2 + .pedigree-wrapper {
  margin-top: 1em;
}
```

This way, all the style stays in one place, and this only applies when needed.

sdumitriu: Although I think it can be removed completely, the extra padding isn't that useful. If you think more space is needed, then that's a global problem for all sections, including in the patient record, and should be solved globally by [increasing the bottom margin of every header](https://github.com/phenotips/phenotips/blob/1a77df5863363e147ee506c4f2754c54d857f8c3/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml#L3909).

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__family.xml
allasm: Because everyone who had data in that field will have to enable it at least once to check migration and/or to manually transfer data that has not been transferred. And i wanted to make this part more clear/obvious

components/family-studies/api/src/main/java/org/phenotips/studies/family/Pedigree.java
```diff
@@ -79,14 +79,14 @@
     List<String> extractIds();
 
     /**
-     * @return phenotipsID of the patient which is the proband of the family, or null if pedigree has no proband or
```
allasm: Hmm, what was wrong on this line?

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
```diff
@@ -259,12 +260,22 @@ public Pedigree getPedigree()
...
+                    JSONObject pedigreeJSON = new JSONObject(data.toText());
+                    if (DefaultPedigree.isSupportedPedigreeFormat(pedigreeJSON)) {
+                        return new DefaultPedigree(pedigreeJSON, image.toText());
+                    } else if (NewFormatPedigree.isSupportedPedigreeFormat(pedigreeJSON)) {
+                        return new NewFormatPedigree(pedigreeJSON, image.toText());
+                    }
                 }
             } catch (XWikiException e) {
-                this.logger.error("Error reading data from pedigree. {}", e.getMessage());
+                this.logger.error("Error reading data from pedigree: [{}]", e);
```
allasm: if just e is used wouldn't there be just the stack trace?

allasm: I will update, just curious

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71498PhenoTips2155DataMigration.java
```diff
@@ -132,264 +202,638 @@ public Object doInHibernate(Session hSession) throws HibernateException, XWikiEx
             Query q =
                 this.session.createQuery("select distinct o.name from BaseObject o, BaseObject ro where"
                     + " o.className = :patclass and o.name <> 'PhenoTips.PatientTemplate'"
-                    + " and ro.name = o.name and ro.className = :relclass");
```
allasm: Because I need to get all patients, including those without family studies data.

To be precise, I need to get all patients that may be referred TO by a family studies relation, but was not sure how to quickly do the query which gets all patients that either have a relationship, or are referred to by a relationship. I assumed this part will be fast enough and only done once per instance, so not a big deal to optimize the query to get only those patients I need.

components/family-studies/api/src/main/java/org/phenotips/studies/family/Pedigree.java
allasm: I did it manually because it looked "better" to me for this particular text...

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
sdumitriu: There will be the stacktrace, but not the right message. It looks like this:

```
[main] ERROR some.package.SomeClass - Error reading data from pedigree: [{}]
java.lang.Exception: x
	at some.package.SomeClass.main(SomeClass.java:49)
```

Notice how `[{}]` is printed uninterpolated.

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71498PhenoTips2155DataMigration.java
sdumitriu: Then remove `ro` completely from the query, it doesn't do anything except slow things down.

allasm: Will do

allasm:
Addressed all @sdumitriu comments made up to this point

PR #2131, opened 2016-11-03
PT-2014: A&W in the pedigree.

veronikaslc:


veronikaslc:
@allasm @a-gvo please, review

components/pedigree/resources/src/main/resources/pedigree/view/person.js
```diff
@@ -1505,6 +1544,13 @@ define([
                 } else {
                     this.setLifeStatus('alive');
                 }
+                if(info.hasOwnProperty('aliveandwell')) {
+                    if (!this.getAliveAndWell()) {
```
allasm: this should be something ismilar to how"evaluated status" is set below:

```
if (this.getAliveAndWell() != info.aliveandwell) {
   this.setAliveAndWell(info.aliveandwell);
}
```

...the point is to set only if it is different from what it is set to

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -125,6 +139,10 @@ define([
             $super();
 
             this.generateMenuBtn();
+            if (this.getNode().getLifeStatus() == "alive" || this.getNode().getLifeStatus() == "deceased") {
+                this.generateAliveWell();
+            }
+            this.getBoxOnHover().attr({'height': PedigreeEditorParameters.attributes.personHoverBoxRadius * 2 + this.getLabelsOffset()});
```
allasm: this happens even if life status is not alive or dead when no radio buttons are generated

also it should be the other way around, labelsoffsetshould be based on the sizeof extended howerbox, not howerbox on the size of labels which seem random withotu digging into the code

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -23,8 +23,22 @@ define([
...
+
+        /**
+         * Returns the gray hover box height extension
+         *
+         * @method getLabelsOffset
+         * @return {int} 0 by default or PedigreeEditorParameters.attributes.personHoverBoxRadius if we display alive and well radio buttons
+         */
+        getLabelsOffset: function() {
+            if (this.getNode().getLifeStatus() == "alive" || this.getNode().getLifeStatus() == "deceased") {
+                return PedigreeEditorParameters.attributes.personHoverBoxRadius - 20;
```
allasm: instead of some non-obvious computation better add a new parameter to PedigreeEditorParameters.attributes with a comment explaining what it is (which should be done in any case, so that if e.g. font size for labels is changed there the howerbox can be sized accordingly at the same place)

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -158,6 +176,82 @@ define([
...
+                    tick.radioSelected = true;
+                }
+                ticks[i] = tick;
+
+                var text = editor.getPaper().text(this.getX()+25, height+20*i, labeles[i]).attr(PedigreeEditorParameters.attributes.awLabel);
+                var rect = editor.getPaper().rect(this.getX()+5, height+20*i-10, this._width-20, 20, 1).attr(PedigreeEditorParameters.attributes.awRect);
+                rect.tick = tick;
+                rects[i] = rect;
+                if (this._hidden && !this.isMenuToggled()) {
+                    circle.hide();
```
allasm: not very nice, better add all shapes to one set and just use set.hide() which auto hides all elements in the set.

on a more general note you don't need these ticks, you may generate none, and only generate one in `customShow()` below (while destroying all others, if present)

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -294,7 +388,14 @@ define([
...
             }
             this.animateHideHoverZone();
-        }
+        },
+
+        generateCircle: function(x, y, tick) {
+            if (tick)
+                return editor.getPaper().circle(x, y, 5).attr({"fill": "#000"});
+            else
+                return  editor.getPaper().circle(x, y, 5).attr({'stroke': '#000', 'stroke-width': '1', "fill": "#fff"});
```
allasm: attributes should be added to PedigreeEditorParameters.attributes instead of hardcoding

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
```diff
@@ -995,7 +995,8 @@ define([
         },
 
         _labelSelectionOffset: function() {
-            var selectionOffset = this.isSelected() ? PedigreeEditorParameters.attributes.radius/1.4 : 0;
+            var labelsOffset = this.getHoverBox().getLabelsOffset();
+            var selectionOffset = this.isSelected() ? PedigreeEditorParameters.attributes.radius/1.4 + labelsOffset : 0;
```
allasm: ideally getLabelsOffset() should incorporate this magical `PedigreeEditorParameters.attributes.radius/1.4` if we have the method now

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
veronikaslc: Made to happen always by intension. By default we generate extended hover box because the default live status is 'alive'. When live status changes to anything that's is not alive or deceased we need this line to shrink hover box. If life status changes again to alive we need this line to extend hover box.

AlexLapena:
Looks good, will check again when the changes are made.

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
allasm: well, `getLabelesOffset()` is supposed to say somethign about labels, not about hoverboxes. If you treat it as described, you shoudl rename it to `hoverBoxSize()`

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
allasm: actually, the fastest way is to have just one tick, and hide/show as usual, and on tick just move position. Less drawing, less hassle. Bsically you wont even need special hide/show code at all, it will hide and show together with everything. And inlyon state change code will move the position - even if it is hidden. And once shown it will show and in the correct place, and nothign special to hide or show

components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
allasm: just fix other issues first and we'll discuss this separately

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
veronikaslc: +1

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -158,6 +176,67 @@ define([
...
+         * @method generateAliveWell
+         * @return {Raphael.el} Raphael element set
+         */
+        generateAliveWell: function() {
+            var node = this.getNode();
+            var lifeStatus = node.getLifeStatus();
+            var aliveandwellStatus = node.getAliveAndWell();
+            //generate 3 radio buttons
+            var labeles = ["Alive", "Alive & Well", "Deceased"];
+            var ticks = [];
```
allasm: you don't use `ticks` any more i think

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -158,6 +176,67 @@ define([
...
+            var rects = [];
+            var height = this.getY()+this._height-this.getHoverBoxHeightExtension()-10;
+            var yTickIndex = 0;
+            var aliveAndWell = editor.getPaper().set();
+
+            for (var i = 0; i < 3; i++) {
```
allasm: `i < labeles.length()`

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -158,6 +176,67 @@ define([
...
+                var circle = this.generateCircle(this.getX()+15, height+20*i, false);
+                if ((lifeStatus == "alive" && !aliveandwellStatus && i == 0)
```
allasm: for clarity, i woudl replace all references to `i == 0` with `labels[i] == "Alive"` etc.

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -158,6 +176,67 @@ define([
...
+                    || (aliveandwellStatus && i == 1)
+                    || (lifeStatus == "deceased" && i == 2) ) {
+                    yTickIndex = i;
+                }
+                var text = editor.getPaper().text(this.getX()+25, height+20*i, labeles[i]).attr(PedigreeEditorParameters.attributes.awLabel);
+                var rect = editor.getPaper().rect(this.getX()+5, height+20*i-10, this._width-20, 20, 1).attr(PedigreeEditorParameters.attributes.awRect);
+                rects[i] = rect;
+                var rowSet = editor.getPaper().set().push(circle, text, rect);
+                if (this._hidden && !this.isMenuToggled()) {
```
allasm: it is better to do this check once outside the loop after the tick is created and do `aliveAndWell.hide()` which will hide everything

allasm: ..it seems then we wont need rowSet as well

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -294,7 +373,16 @@ define([
...
             this.animateHideHoverZone();
-        }
+        },
+
+        /**
+         * Generate circles to draw radio buttons, white under-circle if tick if false and black if true
+         *
+         * @method generateCircle
+         */
+        generateCircle: function(x, y, tick) {
```
allasm: rename to `_generateRadioTickCircle`?

veronikaslc:
@allasm commit amended w.r.t. codereview comments

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
sdumitriu: That looks like a translatable label, whixh won't work nicely when adding support for other languages.

allasm:
@veronikaslc please review, @sashaandjic please test

What needs to be tested: A&W itself, pedigree node grey hoverbox behaviour in general (appears on mouseover, hides on mouseout) plus dragging (parent/child/relationship handles), plus dragging from a legend. E.g. one of the bugs I discovered and fixed was that after using any of the handles A&W links became unclickable, stuff like that, interplay between various things you can do with the hoverbox

sashaandjic:
Regression: changed status does not get saved in pedigree editor (e.g. click on deceased, save and close pedigree. when you open the pedigree the node is alive again). this worked when I tested this branch previously.) 

Second issue is that in the pedigree editor form "Alive & Well" is separated from other statuses.  It should be the first option (before "Alive") in the pedigree editor form.

AlexLapena:
@allasm I also had the same regression issue as Sasha. I also agree with the "Alive & Well" checkbox issue in the pedigree editor form -- it's not super intuitive for the user.

sashaandjic:
When I click on the A&W check box in the node form, the radio buttons on the node menu do not get updated.

allasm:
Addressed comments:
- fixed A&W checkbox in the menu sometimes not being reflected in new A&W hoverbox widget
- fixed life status not being saved when selected using the new widget
- fixed undo/redo issues (bugs and UI glitch) when A&W was involved
- made life status radio buttons non-disabled in node menu when A&W is selected

Undo/redo system was slightly altered in pedigree and needs a bit of attention when testing

sdumitriu:
Wasn't "A&W" supposed to be clearly printed somewhere on/near the targeted nodes?

sashaandjic:
Yes, we'll do it for the next milestone.

veronikaslc:
@allasm Life status is still not being saved when selected using the new widget. (more correctly, the bug is that it is stored but not displayed)
My guess is that `aliveandwell` node property gets filtered out [here](https://github.com/phenotips/phenotips/blob/PT-2014/components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js#L253) because it is not in the `keepProperties` list.
May be better to close PR and do a new one with fixes of this and PT-2945 all together.

sashaandjic:
@veronikaslc can you give me more details. i am unable to reproduce your findings. so far, I haven't found any issues.

veronikaslc:
like previous time, status does not get saved in pedigree editor (e.g.
click on deceased, save and close pedigree. when you open the pedigree the
node is alive again).

allasm:
@sashaandjic, @veronikaslc - there was indeed a bug that Alive&Well status was not saved (technically, not loaded) for nodes linked to PT patients - this has been fixed just now. But plain "alive" and "deceased" statuses are saved and loaded correctly for all nodes, as far as I can tell, @veronikaslc can you provide more details?

allasm:
P.S. in response to an earlier comment: since this PR has a bit more than A&W (includes some refactoring of pedigree hoverboxes) I would prefer this PR to go in so that other code could be based off it without extra merging

veronikaslc:
When I was testing yesterday, clicking on the alive and well tick or deceased tick does not get loaded when redigree saved and reopened.

veronikaslc:
+1, now works.

sashaandjic:
When I uncheck the "Date of death unknown", should the life status go back to what it was prior to checking the "dod unknown" or should it be "Alive"? Right now it goes back to the previous status.

allasm:
As discussed with @sashaandjic what Sasha found is the correct behaviour

sashaandjic:
+1

PR #2133, opened 2016-11-03
Various Excel fixes

sdumitriu:
PT-2906, PT-2907, PT-2908, PT-2909, PT-2910, PT-2911 and PT-2914.


veronikaslc:
PT-2906: if I select All field for export, Genes are shifted one cell to the right
PT-2908: there is no Apgar header

sdumitriu:
@veronikaslc: The problem you spotted in PT-2906 was in fact PT-2914.

For PT-2908 it looks like I forgot to push the latest rebased code before merging PT-2099.

Both issues fixed, please retest.

PR #2134, opened 2016-11-03
PT-2854

a-gvo:
@marta- @buske Please review.

### DIFF
--- a/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
@@ -141,6 +141,8 @@
     method : "get",
     // The name of the request parameter holding the input stub
     varname : "input",
+    // The name of the request parameter holding the input stub for exact term search
+    matchVarname : "id",
     // The CSS classname of the suggest list
     className : "ajaxsuggest",
     timeout : 2500,
@@ -208,6 +210,10 @@
   aSuggestions : [],
   iHighlighted : null,
   isActive : false,
+  // True iff the suggest dropdown is displayed, or an item from it was just selected.
+  suggestionSelected: false,
+  // Input that has been submitted.
+  lastSubmitted: "",
 
   /**
    * Initialize the suggest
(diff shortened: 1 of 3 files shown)


sashaandjic:
I am unable to tab out of the GENE field when I type exact geneID.

sdumitriu:
Instead of adding a new widget, why not make this standard behavior for `PhenoTips.widgets.Suggest`?

sashaandjic:
+1 Works as expected in: genes, phenotypes and OMIM disorder

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -367,6 +382,222 @@
...
+    if (!matchSubmitted) {
+      // Our input didn't change. No need to search again.
+      if (query == this.lastSubmitted &amp;&amp; query.length &gt; 1) {
+        return;
+      }
+      this.latestRequest++;
+      var requestId = this.latestRequest;
+      var pointer = this;
+      clearTimeout(this.ajID);
+      this.ajID = setTimeout( function() { pointer.submitAjaxSearchMatch(query, requestId) }, this.options.delay );
```
sdumitriu: You don't need to use the delay here, since its purpose is to not trigger too many requests while the user is still typing. But here, the input is not focused anymore, we know the user isn't going to be typing anything right away.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -367,6 +382,222 @@
...
   /**
+   * Attaches mousedown observers on newly created suggests.
+   *
+   * @param event the event being observed
+   */
+  onSuggestCreated : function (event) {
+    this.suggestionSelected = false;
+    var _this = this;
+    // A mousedown event occurred in the suggest. Don't want onInputBlurred to perform a search on this loss of focus
+    $(this.container).observe("mousedown", function(event) {
```
sdumitriu: `this.container` doesn't exist. Are you referring to the suggestions list? That's in `event.memo.container`.

PR #2136, opened 2016-11-03
PT-2883: Add chromosome, start position, end position and reference genome as fields for variants.

veronikaslc:
Fixed.

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -277,7 +277,7 @@ public class DataToCellConverter
         String[] fieldIds =
             { "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity",
             "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence",
-            "variants_segregation", "variants_sanger" };
+            "variants_segregation", "variants_sanger", "variants_coordinates" };
         // FIXME These will not work properly in different configurations
         String[][] headerIds =
             {
@@ -294,8 +294,9 @@ public class DataToCellConverter
             { "segregation", "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp", "transcript",
                 "protein", "cdna" },
             { "sanger", "segregation", "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp",
-                "transcript", "protein", "cdna" }
-            };
+                "transcript", "protein", "cdna" },
+            { "reference_genome", "end_position", "start_position", "chromosome", "cdna" }
+        };
         Set<String> present = addHeaders(fieldIds, headerIds, enabledFields);
         this.enabledHeaderIdsBySection.put(sectionName, present);
     }
@@ -364,8 +365,10 @@ public class DataToCellConverter
         section.addCell(cell);
         hX++;
 
-        List<String> fields = Arrays.asList("cdna", "protein", "transcript", "dbsnp", "zygosity", "effect",
-            "interpretation", "inheritance", "evidence", "segregation", "sanger");
+        List<String> fields =
+            Arrays.asList("cdna", "protein", "transcript", "dbsnp", "zygosity", "effect", "interpretation",
+                "inheritance", "evidence", "segregation", "sanger", "chromosome", "start_position",
+                "end_position", "reference_genome");
 
         for (String field : fields) {
             if (!present.contains(field)) {
@@ -410,7 +413,8 @@ public class DataToCellConverter
         }
 
         List<String> fields = Arrays.asList("protein", "transcript", "dbsnp", "zygosity", "effect",
-            "interpretation", "inheritance", "evidence", "segregation", "sanger");
+            "interpretation", "inheritance", "evidence", "segregation", "sanger", "chromosome", "start_position",
+            "end_position", "reference_genome");
         List<String> translatables =
             Arrays.asList("zygosity", "effect", "interpretation", "inheritance", "segregation", "evidence", "sanger");
         List<String> evidenceTranslates = Arrays.asList("rare", "predicted", "reported");
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
@@ -45,6 +45,7 @@ import javax.inject.Provider;
 import javax.inject.Singleton;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.slf4j.Logger;
@@ -100,6 +101,14 @@ public class VariantListController extends AbstractComplexController<Map<String,
 
     private static final String INTERNAL_SANGER_KEY = "sanger";
 
+    private static final String INTERNAL_CHROMOSOME_KEY = "chromosome";
+
+    private static final String INTERNAL_START_POSITION_KEY = "start_position";
+
+    private static final String INTERNAL_END_POSITION_KEY = "end_position";
+
+    private static final String INTERNAL_REFERENCE_GENOME_KEY = "reference_genome";
+
     private static final String JSON_VARIANT_KEY = INTERNAL_VARIANT_KEY;
 
     private static final String JSON_GENESYMBOL_KEY = INTERNAL_GENESYMBOL_KEY;
@@ -124,6 +133,14 @@ public class VariantListController extends AbstractComplexController<Map<String,
 
     private static final String JSON_SANGER_KEY = INTERNAL_SANGER_KEY;
 
+    private static final String JSON_CHROMOSOME_KEY = INTERNAL_CHROMOSOME_KEY;
+
+    private static final String JSON_START_POSITION_KEY = INTERNAL_START_POSITION_KEY;
+
+    private static final String JSON_END_POSITION_KEY = INTERNAL_END_POSITION_KEY;
+
+    private static final String JSON_REFERENCE_GENOME_KEY = INTERNAL_REFERENCE_GENOME_KEY;
+
     private static final List<String> ZYGOSITY_VALUES = Arrays.asList("heterozygous", "homozygous", "hemizygous");
 
     private static final List<String> EFFECT_VALUES = Arrays.asList("missense", "nonsense", "insertion_in_frame",
@@ -141,6 +158,11 @@ public class VariantListController extends AbstractComplexController<Map<String,
     private static final List<String> SEGREGATION_VALUES = Arrays.asList("segregates", "not_segregates");
 
     private static final List<String> SANGER_VALUES = Arrays.asList("positive", "negative");
+
+    private static final List<String> CHROMOSOME_VALUES = Arrays.asList("1", "2", "3", "4", "5", "6", "7", "8", "9",
+        "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "X", "Y");
+
+    private static final List<String> REFERENCE_GENOME_VALUES = Arrays.asList("GRCh37", "GRCh38", "NCBI36");
 
     @Inject
     private Logger logger;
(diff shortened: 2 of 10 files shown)


veronikaslc:
@a-gvo please, review

a-gvo:
@veronikaslc  Looks good. One scenario though: if I enter something invalid for start and/or end positions -> Save and View Summary, then try to edit, the "Start position" and/or "End position" input fields disappear altogether.

veronikaslc:
@a-gvo added one more commit, please, review

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +354,25 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
...
+          var theEvent = evt || window.event;
+          var key = theEvent.keyCode || theEvent.which;
+          key = String.fromCharCode( key );
+          var regex = /[0-9]|\./;
+          if( !regex.test(key) ) {
+            theEvent.returnValue = false;
+            if(theEvent.preventDefault) theEvent.preventDefault();
+          }
+        });
+      });
```
marta-: This disallows ALL keys that are not numbers. It's impossible to use BACKSPACE for corrections, CTRL+V for pasting a value from a lab report, TAB for navigating to the next control etc, which makes the input quite unusable.

veronikaslc:
@marta- amended last commit wrt codereview comments

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +353,18 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
...
       variantFooter.insert({before : newMoreInfoRow});
 
+      // restrict only numerical input for start_position and end_position variant fiellds
+      newMoreInfoRow.select('input:[id$=_start_position]', 'input:[id$=_end_position]').each( function(input) {
+        input.observe('keyup', function(event){
+          this.value=this.value.replace(/[^\d]/,'');
+        });
+      });
+
+      // get the last value of reference_genome of variants related to current gene symbol to create a new variant row with this value
```
marta-: Minor comment: IMO it should look for the last value of reference_genome for any variants in the table, not just the current gene.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +353,18 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       newMoreInfoRow.insert(varMoreInfoInner);
       variantFooter.insert({before : newMoreInfoRow});
 
+      // restrict only numerical input for start_position and end_position variant fiellds
+      newMoreInfoRow.select('input:[id$=_start_position]', 'input:[id$=_end_position]').each( function(input) {
+        input.observe('keyup', function(event){
+          this.value=this.value.replace(/[^\d]/,'');
+        });
+      });
```
marta-: Problems with this:
1/ I can paste any text, not just numbers, and only the first non-digit characters will be removed. This is a big issue because:
* The users are likely to copy & paste positions from a lab report rather than type them
* It's easy to accidentally copy some extra spaces around the number
* A numerical value with spaces around it unfortunately is not parsed as a number and not saved in the number field, so the data will be lost upon saving.

2/ If I hold a letter key to type multiple letters, it will only remove the first one.

Both can probably  be fixed if you do a global replace: `this.value = this.value.replace(/[^\d]/g,'');`.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
veronikaslc: In that scenario, if user creates 2 genes and first adds a variant to the second gene, then starts adding variants to the first gene, then all variants will be created with the reference-genome value if the second gene, just bacause it us allways the last one, even if for tge first gene user wants different one.

marta-: That is true, however IMO it's quite unlikely users would actually jump from the 2nd gene back to the first one. We should do our best make it easy for users follow the steps in the order they'd be 95% of the time, which is transferring data from a report to the electronic record top to bottom.

veronikaslc:
@marta- amended last commit again

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -427,6 +439,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       });
     },
 
+    restrictNumericInput : function () {
+      this._geneTable.select('input:[id$=_start_position]', 'input:[id$=_end_position]').each( function(input) {
+        input.observe('keyup', function(event){
+          this.value=this.value.replace(/[^\d]/,'');
```
marta-: Duplicated code + bug: this copy of the code which restricts inputs to numerical values doesn't have the global flag and will have the issues I reported in the previous review.

Must change: `this.value = this.value.replace(/[^\d]/g,'');`.

Suggestion for further improvement:

```
restrictNumericInput : function (element) {
  (element || this._geneTable).select('input:[id$=_start_position]', 'input:[id$=_end_position]').invoke('observe', 'keyup', function(event) {
    this.value = this.value.replace(/[^\d]/g,'');
  });
}
```

Call `this.restrictNumericInput()` (no parameter, defaults to `this._geneTable`) in the initialization, and `this.restrictNumericInput(newMoreInfoRow)` when adding a new variant.

sdumitriu: `keyup` is not enough, since our users are likely to paste with the mouse. The [`input` event](https://developer.mozilla.org/en-US/docs/Web/Events/input) seems to be supported by our target browsers, and would cover any kind of change.

veronikaslc: +1, that sounds optimal, will test

veronikaslc:
@marta- amended commit with better solution

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -142,6 +159,11 @@
 
     private static final List<String> SANGER_VALUES = Arrays.asList("positive", "negative");
 
+    private static final List<String> CHROMOSOME_VALUES = Arrays.asList("1", "2", "3", "4", "5", "6", "7", "8", "9",
```
marta-: All property values should be read from the XClass, not duplicated in the Java class. Since existing code has exactly the same problem, this doesn't have to be fixed for this PR. But it looks like a refactoring will be needed in the future.

components/export/ui/src/main/resources/PhenoTips/ExportJSON.xml
```diff
@@ -41,7 +41,7 @@
 
 {{velocity wiki="true"}}{{{
 #set ($codeProperties = ['phenotype', 'negative_phenotype', 'prenatal_phenotype', 'omim_id'])
-#set ($nonPatientClassProperties = ['doc.name', 'referrer', 'author', 'creationDate', 'date', 'genes', 'genes_status', 'genes_strategy', 'genes_comments', "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity", "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence", "variants_segregation", "variants_sanger"])
+#set ($nonPatientClassProperties = ['doc.name', 'referrer', 'author', 'creationDate', 'date', 'genes', 'genes_status', 'genes_strategy', 'genes_comments', "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity", "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence", "variants_segregation", "variants_sanger", "variants_chromosome", "variants_start_position", "variants_end_position", "variants_reference_genome"])
```
sdumitriu: These four fields only make sense together. Instead of adding 4 new checkboxes, add only one: `Variant coordinates`.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +352,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       newMoreInfoRow.insert(varMoreInfoInner);
       variantFooter.insert({before : newMoreInfoRow});
 
+      this.restrictNumericInput(newMoreInfoRow);
+
+      // get the last value of reference_genome of all variantsto create a new variant row with this value
+      refGenomes = this._geneTable.select('select:[id$=_reference_genome]');
```
sdumitriu: Weird selector syntax... Use this: `select('select[id$="_reference_genome"]')`.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +352,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       newMoreInfoRow.insert(varMoreInfoInner);
       variantFooter.insert({before : newMoreInfoRow});
 
+      this.restrictNumericInput(newMoreInfoRow);
+
+      // get the last value of reference_genome of all variantsto create a new variant row with this value
```
sdumitriu: Typo: `variantsto`

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +352,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
...
+      refGenomes = this._geneTable.select('select:[id$=_reference_genome]');
+      if (refGenomes && refGenomes.length > 0) {
```
sdumitriu: `select` is guaranteed to return an array, no need to check if it's define. Use only the second part of the condition.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -353,6 +352,14 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
...
+        newMoreInfoRow.down('select:[id$=_reference_genome]').value = refGenomes[refGenomes.length -1].value;
```
sdumitriu: `select[id$="_reference_genome"]`

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
```diff
@@ -427,6 +434,18 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
       });
     },
 
+    // restrict only numerical input for start_position and end_position variant fiellds
+    restrictNumericInput : function (container) {
+      if (!container) {
+        var container = this._geneTable;
+      }
+      container.select('input:[id$=_start_position]', 'input:[id$=_end_position]').each( function(input) {
```
sdumitriu: `container.select('input[id$="_start_position"]', 'input[id$="_end_position"]')`

components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm
```diff
@@ -107,13 +107,13 @@ $doc.display($prop.getName(), 'view', $fakeObj )##
 #foreach ($prop in $variantObj.xWikiClass.properties)
 #if ($prop.getName() != 'genesymbol' && !$options.defaultProperties.contains($prop.getName()) && $variantObj.getProperty($prop.getName()))
 #set($rawValue = $variantObj.getProperty($prop.getName()).value)
-#set($isValueEmpty = $rawValue == '' || $rawValue == 'NA' || $rawValue == [])
+#set($isValueEmpty = $rawValue == '' || $rawValue == 'NA' || $rawValue == [] || !$!{rawValue})
```
sdumitriu: Are you adding a check for `false` or `null`? If `null`, then it's safer to change `$rawValue == ''` into `"$!{rawValue}" == ''`.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -295,8 +296,17 @@ public void variantsSetup(Set<String> enabledFields) throws Exception
             { "segregation", "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp", "transcript",
                 "protein", "cdna" },
             { "sanger", "segregation", "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp",
-                "transcript", "protein", "cdna" }
-            };
+                "transcript", "protein", "cdna" },
+            { "chromosome", "sanger", "segregation", "evidence", "inheritance", "interpretation", "effect",
```
sdumitriu: All this code has been wrong for some time, and you're just perpetuating the mistake. The intention is to allow a single checkbox in the export dialog to select multiple fields.

Add `variant_coordinates` in `fieldIds`, and `{ "reference_genome", "end_position", "start_position", "chromosome"}` in `headerIds`.

veronikaslc:
@sdumitriu corederview comments addressed

veronikaslc:
@sdumitriu corederview comments addressed

PR #2137, opened 2016-11-03
[bug] Classname 'selected' to be appended only once when input is checked

teyden:
Requesting change for line 476 to simplify applying CSS rules to show only file attachment items that are actually `selected` (when `input.checked == true`). 

The current behaviour does not remove the `selected` class name from the `attachment-item` div when `input.checked` becomes `false`, resulting in extra `selected` class names appended. Unclicking and reclicking the same attachment item produces this.

Please review @veronikaslc. Thanks.


sdumitriu:
Applied a better fix in 776e008

PR #2139, opened 2016-11-04
PT-2785: Broken variants -- fixed

a-gvo:
@sdumitriu @marta- Broken variant fix. Please review.

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/extraGeneVariantData.js
@@ -270,6 +270,7 @@ var ExtraGeneVariantData = (function (ExtraGeneVariantData) {
 
       var className = addTrigger.up('td.variant').className;
       var geneIndex = className.substring(className.lastIndexOf('-') + 1);
+      var geneId = $$('[name="' + this.geneClassName + '_' + geneIndex + '_gene"]')[0].value;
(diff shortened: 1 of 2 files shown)


marta-:
Changes look like they solve the problem (I haven't actually tested yet, but I will soon). One problem with semantics: we're now storing the Ensembl Id in a field called `genesymbol` in the variants class. I think it would make sense to rename this field to `gene` or `geneid` as part of [PT-2777 (Write a migrator to translate existing gene data from HGNC id to Ensembl Id)](https://phenotips.atlassian.net/browse/PT-2777). @sdumitriu thoughts?

sdumitriu:
It's still possible to lose variants:

- add a new gene
- type T
- don't select the suggestion, but focus out
- enter variants for this gene; at this moment, the gene is identified by `T` instead of `ENSG00000164458`, variants reference `T` as their gene
- save, edit again
- re-type T in the gene input, but this time select the suggestion
- save, variants are now lost

This is just an instance of another already present bug: changing the gene symbol to something else also discard existing variants for that gene.

I guess PT-2854 may partially fix this.

marta-:
@sdumitriu This scenario actually reveals a bug in the original code for genes & variants. When you add a variant for a gene, the input for the gene name becomes "locked" (the gene name is not editable). However if you save&view then edit again, the gene name is displayed as an editable input until you add another variant. This behavior is inconsistent and kind of wrong. When loading the genes and variants form the database for editing, all genes with variants should have names that are not editable.

sdumitriu:
I agree with @marta-, rename `genesymbol` to `gene`.

PR #2140, opened 2016-11-07
PT-2712: Phenotype Selector Bug when using Browse Related Terms

sdumitriu:
Fixed

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
@@ -1324,9 +1324,8 @@ document.observe('xwiki:dom:loaded', function()
                 elt.__termCategory &amp;&amp;  elt.__termCategory.insert(negative) || (elt.__termCategory = negative);
             } else if (selected == 'yes') {
               var element = event.findElement('.entry');
-              if (!element.down('.descendents') || element.hasClassName('collapsed')) {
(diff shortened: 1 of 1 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -1325,8 +1325,7 @@ document.observe('xwiki:dom:loaded', function()
             } else if (selected == 'yes') {
               var element = event.findElement('.entry');
               if (!element.down('.descendents') || element.hasClassName('collapsed')) {
```
marta-: @sdumitriu As per my earlier comment about parents being expanded when we click 'Y' and producing a possibly confusing tree, this like could be changed to
`if ((!element.down('.descendents') || element.hasClassName('collapsed')) && element.down('.expand-tool')) {`

sashaandjic:
When I go to see related terms and press "Y" for the parent, my original term gets listed twice. 
To reproduce: 
- type "slow" in the search and look for related terms for "Slow-growing hair",
- select "Y" for the very first term listed "Abnormality of hair growth rate", 
- notice that another "Slow-growing hair" appears

marta-:
@sashaandjic Yep, that's exactly what I mentioned above. Your comment is a confirmation that it may indeed bother/confuse users. The proposed change should fix it. Thanks!

sdumitriu:
@marta- @sashaandjic Applied the suggestion, please retest.

PR #2141, opened 2016-11-07
[bugfix] Small UI fix: the header doesn't align on the right side with the page body

sdumitriu:


veronikaslc:
+1 to changes
A wish: if we could lift`Search for a patient record...` input to be on one level with elements in a row, that would IMHO improve a look even more.

components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -599,10 +603,10 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
 }
 #phenotips-globalTools .search input {
   margin-left: 0;
-  margin-right: 5px;
+  margin-right: 0;
   margin-top: 0.05em;
   min-height: 20px;
```
marta-: I think removing `margin-top` and setting `min-height: 1.6em` may address Veronika's comment.

PR #2142, opened 2016-11-07
PT-2916: Gene names are editable after save&view+edit even if the gene has associated variants, which may cause the variants to disappear

veronikaslc:


marta-:
@sdumitriu can you be more specific?

sdumitriu:
The order of css classes should not matter, so this indicates a deeper problem somewhere else.

veronikaslc:
Agreed that this fix is a queck one, the better but longer would be us regex to ger gene/variant index

marta-:
@sdumitriu This is a quick fix to get the feature to work sooner rather than later. Veronika will refactor the code to remove the dependency on CSS classes for storing and retrieving the gene/variant index, and store this kind of information in hidden inputs instead. This is not a small task. It will take some time and the first attempts may introduce further regressions, so I thought it would be OK to postpone until the a future sprint/release so that we can take our time reviewing and testing it. WDYT?

PR #2146, opened 2016-11-08
PT-2865: Pedigree export dialog does not fit on the screen if PED export is selected and there are many traits recorded in the family; 	PT-2973: Pedigree export dialogue lists PED export options for other export types in some cases

veronikaslc:


veronikaslc:
adjusted based on the client height

components/pedigree/resources/src/main/resources/pedigree/view/exportSelector.js
```diff
@@ -255,11 +255,17 @@ define([
...
-                hasCausalGenes && this._addPedOption("causalGenes", causalGenes, pedContainer, "confirmed causal genes");
+                var traitsContainner = new Element('table');
+
+                hasDisorders && this._addPedOption("disorders", disorders, traitsContainner);
+                hasPhenotypes && this._addPedOption("phenotypes", hpos, traitsContainner);
+                hasCancers && this._addPedOption("cancers", cancers, traitsContainner);
+                hasCandidateGenes && this._addPedOption("candidateGenes", candidateGenes, traitsContainner, "candidate genes");
+                hasCausalGenes && this._addPedOption("causalGenes", causalGenes, traitsContainner, "confirmed causal genes");
+
+                pedContainer.insert(traitsContainner.wrap('div', {'class': 'scrollable-container', 'style': 'max-height:'+Math.floor(window.innerHeight/2)+'px;'})
```
marta-: Setting the max height to be half of the viewport isn't a reliable way to solve the problem. There's no guarantee that the rest of the dialog will fit in the other half. If we're going to try to fit the whole dialog on the screen, then a better option is to:
* get the modal's height and the dcrollable container's height when fully displayed
* compute how big the scrollable container needs to be so that the modal can fit on the screen (something like `sHeight - Math.max (0, modalHeight - viewportHeight)`)
* adjust the height of  the scrollable container to this value

Disadvantages:
- you'd need to recompute on each window resize
- on REALLY small screens it will be impossible to fit because the rest of the modal is already too big

There are still options:
- make the whole content of the modal scrollable, not just that part (while keeping the modal fixed & fully on screen)
- make the modal not fixed, allowing it to scroll with the background content.

veronikaslc: Found out we already adjust modal's height for printing [here](https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/view/printDialog.js#L208) However, we do not recompute on each window resize there. If we recompute in one case but don't recompute in the other this seems inconsistent.

allasm: We do recompute on each window resize for print - https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/view/printDialog.js#L153

veronikaslc: As far as I tested a dialog stays the same size if you don't close it while resizing a window. I guess I misunderstood the context of recomputing thinking it should be recomputed when opened.

veronikaslc: Appeared there was a bug in print dialog resizing function, which is fixed in second commit.

veronikaslc:
@allasm please, review

components/pedigree/resources/src/main/resources/pedigree/view/printDialog.js
```diff
@@ -236,7 +217,7 @@ define([
                                                                     this._moveHorizontally,
                                                                     options);
             this.previewContainer.update(previewHTML);
-            this._adjustPreviewWindowHeight();
+            GraphicHelpers.adjustPreviewWindowHeight("pedigree-print-dialog", 'printPreview', this._minPreviewHeight, this._maxPreviewHeight);
```
allasm: to avoid repetition of arguments `"pedigree-print-dialog", 'printPreview', this._minPreviewHeight, this._maxPreviewHeight"` I would keep `this._adjustPreviewWindowHeight()` method, and internally just called `GraphicHelpers.adjustPreviewWindowHeight`

Same with exportSelector.js (where i would add a new `_adjustExportWindowHeight()`)

allasm:
If it is quick, I would add a fix for PT-2973 to this pull request

veronikaslc:
@allasm PT-2973 fix added

components/pedigree/resources/src/main/resources/pedigree/pedigreeEditorParameters.js
```diff
@@ -126,6 +126,8 @@ define([], function(){
         hoverboxHighlightColor: "green",
         mainGlowColor: "green",
         secondaryGlowColor: "#556B2F"
+        maxPEDexportContainerHeight: 600,
```
marta-: `,` missing on previous line, this causes the build to fail.

veronikaslc: indeed, refactoring regression. made a separate commit with it, now should build.

veronikaslc:
@sashaandjic please, retest

components/pedigree/resources/src/main/resources/pedigree/view/graphicHelpers.js
```diff
@@ -315,6 +315,34 @@
...
+	 * @param {Number} minResizableElementHeight
+	 * @param {Number} maxResizableElementHeight
+     */
+    GraphicHelpers.adjustPreviewWindowHeight = function(dialogSelectorClass, resizableElementId, minResizableElementHeight, maxResizableElementHeight) {
+        var pedigreeDialogue = $$('.'+dialogSelectorClass)[0];
+		var resizableElement = $(resizableElementId);
+        if (!pedigreeDialogue || !resizableElement) {
+            return;
+        }
+        var screenHeight = window.innerHeight - 10;
```
veronikaslc: because it was here https://github.com/phenotips/phenotips/pull/2146/files#diff-3a30a4eb55105151589d0845319bc80cL214

allasm:
One note - I would squash all commits into one since most of them have "codereview comments addressed" as the commit comment, which is not useful after pull request has been approved. I think we discussed that we generaly want to do it after pull rqs are approved

PR #2148, opened 2016-11-09
PT-2925 and PT-2860 (OLD)

allasm:
TO BE REMOVED ONCE https://github.com/phenotips/phenotips/pull/2305 IS MERGED IN

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -599,9 +599,7 @@ public class DataToCellConverter
 
     public DataSection documentInfoBody(Patient patient) throws Exception
     {
-        @SuppressWarnings("deprecation")
-        DocumentAccessBridge dab = Utils.getComponent(DocumentAccessBridge.class);
-        XWikiDocument patientDoc = (XWikiDocument) dab.getDocument(patient.getDocument());
+        XWikiDocument patientDoc = patient.getXDocument();
         String sectionName = "documentInfo";
         Set<String> present = this.enabledHeaderIdsBySection.get(sectionName);
         if (present == null || present.isEmpty()) {
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyRepository.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyRepository.java
@@ -150,7 +150,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
         try {
             XWikiContext context = this.provider.get();
             XWiki xwiki = context.getWiki();
-            xwiki.deleteDocument(xwiki.getDocument(family.getDocument(), context), context);
+            xwiki.deleteDocument(xwiki.getDocument(family.getXDocument(), context), context);
         } catch (XWikiException ex) {
             this.logger.error("Failed to delete family document [{}]: {}", family.getId(), ex.getMessage());
             return false;
@@ -210,7 +210,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
             return null;
         }
         String patientId = patient.getId();
-        XWikiDocument patientDocument = getDocument(patient);
+        XWikiDocument patientDocument = patient.getXDocument();
         if (patientDocument == null) {
             return null;
         }
@@ -256,7 +256,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
 
         String patientId = patient.getId();
         XWikiContext context = this.provider.get();
-        XWikiDocument patientDocument = getDocument(patient);
+        XWikiDocument patientDocument = patient.getXDocument();
         if (patientDocument == null) {
             throw new PTInvalidPatientIdException(patientId);
         }
@@ -268,7 +268,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
             throw new PTPedigreeContainesSamePatientMultipleTimesException(patientId);
         }
 
-        if (!this.setFamilyReference(patientDocument, family.getDocument(), context)) {
+        if (!this.setFamilyReference(patientDocument, family.getXDocument(), context)) {
             throw new PTInternalErrorException();
         }
         if (!savePatientDocument(patientDocument, "added to family " + family.getId(), context)) {
@@ -277,7 +277,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
 
         // Add member to the list of family members
         members.add(patientLinkString(patient));
-        BaseObject familyObject = family.getDocument().getXObject(Family.CLASS_REFERENCE);
+        BaseObject familyObject = family.getXDocument().getXObject(Family.CLASS_REFERENCE);
         familyObject.set(PhenotipsFamily.FAMILY_MEMBERS_FIELD, members, context);
 
         // only save family document if this add() is not performed as a part of a batch update
@@ -316,7 +316,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
 
         String patientId = patient.getId();
         XWikiContext context = this.provider.get();
-        XWikiDocument patientDocument = getDocument(patient);
+        XWikiDocument patientDocument = patient.getXDocument();
         if (patientDocument == null) {
             throw new PTInvalidPatientIdException(patientId);
         }
@@ -349,7 +349,7 @@ public class PhenotipsFamilyRepository implements FamilyRepository
 
         // Remove patient from family's members list
         members.remove(patientLinkString(patient));
-        BaseObject familyObject = family.getDocument().getXObject(Family.CLASS_REFERENCE);
+        BaseObject familyObject = family.getXDocument().getXObject(Family.CLASS_REFERENCE);
         familyObject.set(PhenotipsFamily.FAMILY_MEMBERS_FIELD, members, context);
 
         if (!batchUpdate) {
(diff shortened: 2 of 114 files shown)


allasm:
@itaiGershtansky @veronikaslc please review

@AlexLapena , @sashaandjic please test.

Since there are changes in all areas related to patient data, a few different things should be tested:
1) PT-2860 as described in the issue
2) pushing in general (smoke test)
3) setting patient access rights, via UI and via REST (if we already have REST for that?)
4) patient data: setting via UI, getting/setting via REST (smoke test)
5) smoke test family studies

allasm:
@sdumitriu should probably review once the initial code review is done and any obvious problems/omissions/typos are ironed out

allasm:
Note for code reviewers: `getDocument()` - which got deprecated now -  is used just about everywhere.

I only touched parts that had to be changed in other ways (e.g. not just rename the method used), mostly in `org.phenotipos.data` package/namespace, all the other places should be updated as part of PT-2640

sashaandjic:
@AlexLapena: smoke test https://phenotips.org/Release/ReleasePHENOTIPS13M3

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatientRepository.java
```diff
@@ -114,9 +114,9 @@ public synchronized Patient create(DocumentReference creator)
         try {
             XWikiContext context = this.xcontextProvider.get();
             Patient patient = super.create(creator);
-            XWikiDocument doc = (XWikiDocument) this.bridge.getDocument(patient.getDocument());
+            XWikiDocument doc = patient.getXDocument();
             doc.getXObject(Patient.CLASS_REFERENCE).setLongValue("identifier",
-                Integer.parseInt(patient.getDocument().getName().replaceAll("\\D++", "")));
+                Integer.parseInt(patient.getDocumentReference().getName().replaceAll("\\D++", "")));
```
itaiGershtansky: patient.**getId()**.replaceAll("\\D++", "");

components/patient-data-sharing/receiver-api/src/main/java/org/phenotips/data/receive/internal/DefaultReceivePatientData.java
```diff
@@ -241,7 +241,7 @@ protected JSONObject generateFailedActionResponse(String jsonKeyToSet)
 
     protected XWikiDocument getPatientDocument(Patient patient) throws Exception
```
itaiGershtansky: Should we call it getPatientXDocument?

sashaandjic:
Works well when pushing bulk of patients. For the single patient record push I don't even get this option as it mimics the behaviour of updating existing remote record.

Bulk push: 
screen shot 2016-11-10 at 3 02 29 pm

Single patient push: 
screen shot 2016-11-10 at 3 02 03 pm

components/patient-data-sharing/receiver-api/src/main/java/org/phenotips/data/receive/internal/DefaultReceivePatientData.java
allasm: should just remove the method altogether, will do

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatientRepository.java
allasm: +1, will update

allasm:
@sashaandjic you don't get this option when you try to push patients that have been pushed already (you can see the "can't be changed when updating an existing patient" text). Just create a new patient and push, you will get the group select option

sashaandjic:
@allasm - @AlexLapena and I finally figured it out. Here are the steps:
1. push a single record
2. then go to the remote server and delete it
3. try to push again - you will not get the option to assign it to the remote group as it thinks you are updating
4. select fields and proceed with push
5. now it tells you "Unable to update this patient - stored remote GUID is incorrect. Maybe the patient was deleted on the remote server". PUSH AS A NEW PATIENT.

I think it should recognize this odd situation where the remote patient was deleted and offer the group assignment in the step 3.

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_memberReports.xml
```diff
@@ -116,7 +116,7 @@
 #end
 #foreach ($patient in $family.members)
   #if ($patient)
-    #set ($patientDoc = $xwiki.getDocument($patient.document))
```
veronikaslc: should we also change [here](https://github.com/phenotips/phenotips/blob/master/components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml#L493) and [here](https://github.com/phenotips/phenotips/blob/master/resources/ui/src/main/resources/PhenoTips/PatientPhenotypeExportService.xml#L64) ?

sashaandjic:
Showstopper discovered by @AlexLapena : Global Visibility cannot be modified throught UI

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_memberReports.xml
allasm: we should, will do

allasm:
Confirmed: there is a bug in visibility code, will be fixed

allasm:
Fixed visibility bug and addressed all comments

veronikaslc:
missed a [spot](https://github.com/phenotips/phenotips/blob/master/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultDomainObjectFactory.java#L99)?

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -595,9 +595,7 @@ public DataSection documentInfoHeader(Set<String> enabledFields) throws Exceptio
 
     public DataSection documentInfoBody(Patient patient) throws Exception
     {
-        @SuppressWarnings("deprecation")
-        DocumentAccessBridge dab = Utils.getComponent(DocumentAccessBridge.class);
-        XWikiDocument patientDoc = (XWikiDocument) dab.getDocument(patient.getDocument());
+        XWikiDocument patientDoc = patient.getXDocument();
```
veronikaslc: Useless import of [DocumentAccessBridge](https://github.com/phenotips/phenotips/blob/5bf1c40014f2e5de26835a4dcd71fc0803b75161/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java#L30) was not removed.
You can configure your IDE to automatically remove unused imports. This simply saves costly developers time and code-reviewers time.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamilyRepository.java
```diff
@@ -154,7 +154,7 @@ public synchronized boolean deleteFamily(Family family, User updatingUser, boole
         try {
             XWikiContext context = this.provider.get();
             XWiki xwiki = context.getWiki();
-            xwiki.deleteDocument(xwiki.getDocument(family.getDocument(), context), context);
+            xwiki.deleteDocument(xwiki.getDocument(family.getXDocument(), context), context);
```
veronikaslc: same change [here](https://github.com/phenotips/phenotips/blob/PT-2860/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java#L173) and [here](https://github.com/phenotips/phenotips/blob/PT-2860/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientResourceImpl.java#L175) ?

allasm:
@itaiGershtansky @sdumitriu please review again

A top-level description of the changes:

1) SecurePatientRepository now returns a (newly-inctroduced) SecurePatient instead of a Patient, which is read-only and does not expose XDocument (which was a security problem). One side-effect is that it is much harder to write tests for that now, so there are a lot of changes in the tests

2) Entities got a `getSecureDocument()` method (is it a good name?) which returns the equivalent of `$xwiki.getDocument($object.document)`, but is necessary since `$xwiki.getDocument()` will return a Document based on the on-disk copy of the object, not the `$object` in memory, which in theory may have some changes

3) In the previous set of commits patient-access code was modified to work with the Patient and its XDocument (instead of reading stuff from disk from scratch every time). That fixed some bugs, but now scripts which use it only have a SecurePatient instead of a Patient, which is read-only. So some changes were made to take this into account

allasm:
P.S. To save reviewers time, commits before Nov 22, 2016 have been reviewed already, none of the new commits have been squashed with older ones (yet) to make reviewing easier

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
allasm: Note: it seems like excel export is initiated from velocity, and `patient` in an instance of `SecurePatient` here, which breaks things. Need to fix

allasm: Note: it seems like excel export is initiated from velocity, and `patient` in an instance of `SecurePatient` here, which breaks things. Need to fix

allasm:
@sashaandjic found two issues: Excel export is broken (confirmed, caused by this pull request), and BOADICEA export does not work in some cases (to be investigated, sample JSON was sent to me on slack, probably unrelated)

allasm:
All comments have been addressed and all issue discovered while testing so far are fixed

Reviewers (@sdumitriu, @veronikaslc) - is it ok to squash commits that need to be squashed and rebase on top of latest master?

sdumitriu:
@allasm OK for rebase & squash.

allasm:
squashed and rebased on master

allasm:
when do we plan to merge it into master? In the past 4 days there were no changes in the branch and the code as rebased on top of latest master as of that time, and now we got new conflicts again due to some cosmetic changes in master - IMHO we should apply those only if nothing large is pending commits, otherwise better to work on merging

allasm:
Comment by @sdumitriu on slack: "There's a security issue with it: Families return insecure patients"

sdumitriu:
Redone in #2305

PR #2150, opened 2016-11-10
PT-2893: Adjust fields alignment and font size in the extended variant fields

veronikaslc:


veronikaslc:
@marta- @a-gvo please review

PR #2159, opened 2016-11-15
PT-2944: Write a migrator to translate existing gene data in pedigrees.

veronikaslc:


components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
```diff
@@ -0,0 +1,311 @@
...
+        List<String> docs = q.list();
+
+        this.logger.debug("Found {} family documents", docs.size());
+
+        for (String docName : docs) {
+            XWikiDocument familyXDocument;
+            boolean updated = false;
+
+            try {
+                familyXDocument = xwiki.getDocument(this.resolver.resolve(docName), context);
```
allasm: given the query this - hypothetically - may be not a family document any more, so maybe we can just call it just "doc"/"document" to not give maintainers the idea that this is necesserily a family

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
```diff
@@ -0,0 +1,311 @@
...
+                    continue;
+                }
+
+                BaseObject pedigreeXObject = familyXDocument.getXObject(PEDIGREE_CLASS_REFERENCE);
+                if (pedigreeXObject == null) {
+                    continue;
+                }
+
+                this.logger.debug("Updating pedigree for family {}.", docName);
+                updated = this.updatePedigreeGenes(pedigreeXObject, context, docName);
```
allasm: just use `if (!this.updatePedigreeGenes(pedigreeXObject, context, docName)) { continue }`?

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
```diff
@@ -0,0 +1,311 @@
...
+
+        }
+        return null;
+    }
+
+    private boolean updatePedigreeGenes(BaseObject pedigreeXObject, XWikiContext context, String documentName)
+    {
+        boolean updatedGenes = false;
+        boolean updatedColors = false;
+        String dataText = pedigreeXObject.getStringValue(PEDIGREECLASS_JSONDATA_KEY);
```
allasm: I would put all of these inside a try-catch block, since pedigreeXObject.getStringValue may in theory throw something as well I think (or be null if higher-level code is incorrect). The boolean variablkes ar enot useful outside try-catch as well, as the last return can be inside

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
```diff
@@ -0,0 +1,311 @@
...
+                    updatedColors = this.convertColors(pedigreeColors, hgncToEnsemblMap);
+                }
+            }
+
+            if (updatedGenes || updatedColors) {
+                String pedigreeData = pedigree.toString();
+                pedigreeXObject.set(PEDIGREECLASS_JSONDATA_KEY, pedigreeData, context);
+            }
+        } catch (Exception e) {
+            this.logger.error("Family pedigree data is not a valid JSON for family {}: [{}]", documentName, e);
```
allasm: we don't know the exact problem, maybe conversion fails for some other reason (e.g. vocabulary throws an exception) so I would just say "error converting"

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
```diff
@@ -0,0 +1,311 @@
...
+     *
+     * @param geneSymbol the string representation of a gene symbol (e.g. NOD2).
+     * @return the string representation of the corresponding Ensembl ID.
+     */
+    private String getEnsemblId(String geneSymbol)
+    {
+        final VocabularyTerm term = this.hgnc.getTerm(geneSymbol);
+        @SuppressWarnings("unchecked")
+        final List<String> ensemblIdList = term != null ? (List<String>) term.get("ensembl_gene_id") : null;
+        final String ensemblId = ensemblIdList != null && !ensemblIdList.isEmpty() ? ensemblIdList.get(0) : null;
```
allasm: is `get(0)` always guaranteed to be the right term? E.g. is the first pone in array the best? What if there is No exact match, and the first is just the least accurate match? We may have free-text terms which have no exact match but do have some partial match

E.g. http://localhost:8080/get/PhenoTips/GeneNameService?outputSyntax=plain&rand=0.38996313003383465&json=true&q=GS

returns stuff, but none of it is the correct ensemble ID

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
allasm: but you don't use it anywhere else in this file?

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71500PhenoTips2944DataMigration.java
veronikaslc: fix will wait till PT-2777 will be merged.

allasm:
Actually, we need one more thing to do in the migrator. We need to add ID to name (gene symbol) mapping to the legend, similar to what is already done for phenotypes (we have both `colors` and `names`). Genes only have `colors` now, but we should also add `names` now

allasm:
This pull request is outdated and is superseded by the commits in the PT-2012 branch

PR #2162, opened 2016-11-16
PT-2899: When switching to another language while editing warn user o…

marta-:
…f potential data loss

Fixed. Covers navigating away from any other links. Also reworked the language switching UI.

### DIFF
--- a/components/skin/ui/src/main/resources/PhenoTips/Skin.xml
+++ b/components/skin/ui/src/main/resources/PhenoTips/Skin.xml
@@ -429,6 +429,129 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
   </object>
   <object>
     <name>PhenoTips.Skin</name>
+    <number>3</number>
+    <className>XWiki.JavaScriptExtension</className>
+    <guid>a847fd11-a557-4323-9940-9302a4f11f22</guid>
+    <class>
+      <name>XWiki.JavaScriptExtension</name>
+      <customClass/>
+      <customMapping/>
+      <defaultViewSheet/>
+      <defaultEditSheet/>
+      <defaultWeb/>
+      <nameField/>
+      <validationScript/>
+      <cache>
+        <cache>0</cache>
+        <disabled>0</disabled>
+        <displayType>select</displayType>
+        <multiSelect>0</multiSelect>
+        <name>cache</name>
+        <number>5</number>
+        <prettyName>Caching policy</prettyName>
+        <relationalStorage>0</relationalStorage>
+        <separator> </separator>
+        <separators>|, </separators>
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>long|short|default|forbid</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </cache>
+      <code>
+        <disabled>0</disabled>
+        <name>code</name>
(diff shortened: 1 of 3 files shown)


components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -429,6 +429,130 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
...
+      });
+      $$("input[name='action_save']").invoke('observe', 'click', function(event) {
+        window.__formData = $('inline').serialize();
+      });
+      // Upon navigating away, check if anything has changed in the form and warn the user
+      // if it did.
+      window.addEventListener('beforeunload', function(event) {
+        if (window.__formData != $('inline').serialize()) {
+          var message = "$services.localization.render('phenotips.navigation.beforeunload.warning')";
+          console &amp;&amp; console.log(message);
```
sdumitriu: This isn't needed.

components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -429,6 +429,130 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
...
+    </property>
+    <property>
+      <code>(function () {
+  var init = function() {
+    if ($('inline')) {
+      // Some form elements are still being modified after ""xwiki:dom:loaded" is fired.
+      // If the initial state of the form is recorded immediately, there will likely be
+      // differences with the state of the form when navigating the way even if the user
+      // hasn't actually changed any data.
+      // As a quick fix, wait two seconds before recording the "initial" state of the form.
```
sdumitriu: I don't think waiting for a specific amount of time is needed. The code that does the changes is usually run synchronously on `xwiki:dom:loaded`, which means that even a delay of `0` will make this execute after the other form changes are done. You could test this theory.

marta-: You're right, it works. Thanks, I'll amend the commit.

components/skin/ui/src/main/resources/PhenoTips/Skin.xml
marta-: I forgot to remove it. It was the only way to actually see the message was properly translated, since most browsers don't actually display the returned message.

marta-:
@sdumitriu your comments were addressed.

marta-:
@veronikaslc please review. @sashaandjic please test.

components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -429,6 +429,129 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
   </object>
   <object>
     <name>PhenoTips.Skin</name>
+    <number>3</number>
```
sdumitriu: Since this is attached to the skin, it affects not just the language switch, but also all other inline forms (families, studies, workgroups, users...). As such, I'd commit this with its own issue number.

marta-:
@sashaandjic This PR has already been merged into master, but it was only tested in Chrome and Firefox. Please test it in IE and Safari as soon as you get a chance to make sure
1/ the language menu looks and works fine
2/ the warning about data loss appears as expected.
Thanks!

PR #2163, opened 2016-11-17
PT-2821: Added functionality for retrieving multiple patients at once…

a-gvo:
… via REST given a list of identifiers.
Test cases will be added as a separate commit.
@itaiGershtansky @veronikaslc Please review.

### DIFF
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientsFetchResource.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientsFetchResource.java
@@ -1,0 +1,50 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.data.rest;
+
+import org.phenotips.rest.ParentResource;
+import org.phenotips.rest.Relation;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+/**
+ * Resource for working with multiple patient records, identified by their given "external" or "internal" identifiers.
+ *
+ * @version $Id$
+ * @since 1.3M5
+ */
+@Path("/patients/fetch")
+@Relation("https://phenotips.org/rel/patientRecords")
+@ParentResource(PatientsResource.class)
+public interface PatientsFetchResource
+{
+    /**
+     * Retrieve multiple patient records, identified by their given "external" or "internal" identifiers, in their JSON
+     * representation. If any of the indicated patient records don't exist, or if the user sending the request doesn't
+     * have the right to view any of the target patient records, they are excluded from the search results.
+     *
+     * @return JSON representations of the requested patients, or a status message in case of error
+     */
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    Response fetchPatients();
+}
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
@@ -1,0 +1,246 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.data.rest.internal;
+
+import org.phenotips.data.Patient;
+import org.phenotips.data.PatientRepository;
+import org.phenotips.data.rest.PatientsFetchResource;
+import org.phenotips.entities.PrimaryEntity;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.container.Container;
+import org.xwiki.query.Query;
+import org.xwiki.query.QueryException;
+import org.xwiki.query.QueryManager;
+import org.xwiki.rest.XWikiResource;
+import org.xwiki.security.authorization.AuthorizationManager;
+import org.xwiki.security.authorization.Right;
+import org.xwiki.users.User;
+import org.xwiki.users.UserManager;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.slf4j.Logger;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.google.common.base.Function;
+import com.google.common.base.Predicates;
+import com.google.common.collect.FluentIterable;
+
+/**
+ * Default implementation for {@link PatientsFetchResource} using XWiki's support for REST resources.
+ *
+ * @version $Id$
+ * @since 1.3M5
+ */
+@Component
+@Named("org.phenotips.data.rest.internal.DefaultPatientsFetchResourceImpl")
+@Singleton
+public class DefaultPatientsFetchResourceImpl extends XWikiResource implements PatientsFetchResource
+{
+    /** Jackson object mapper to facilitate array serialization. */
+    private static final ObjectMapper OBJECT_MAPPER = getCustomObjectMapper();
+
+    /** Logging helper object. */
+    @Inject
+    private Logger logger;
+
+    @Inject
+    private QueryManager qm;
+
+    @Inject
+    private PatientRepository repository;
+
+    @Inject
+    private AuthorizationManager access;
+
+    @Inject
+    private UserManager users;
+
+    @Inject
+    private Container container;
+
+    /** Function that returns an iterable containing the specified patient objects. */
+    private Function<Object, Iterable<PrimaryEntity>> patientsFromEidFunction = getPatientsFromEidFunction();
+
+    /** Function that returns an iterable containing the specified patient object. */
+    private Function<String, PrimaryEntity> patientFromIdFunction = getPatientFromIdFunction();
+
+    @Override
+    public Response fetchPatients()
+    {
+        final List<Object> eids = this.container.getRequest().getProperties("eid");
+        @SuppressWarnings("unchecked")
+        final List<String> ids = (List<String>) (List<?>) FluentIterable.from(this.container.getRequest()
+            .getProperties("id"))
+            .filter(Predicates.notNull())
+            .toList();
+
+        this.logger.debug("Retrieving patient records with external IDs [{}] and patient IDs [{}] via REST", eids, ids);
+
+        // Get a set of patients from the provided eid and/or id data.
+        final Set<PrimaryEntity> patients = new HashSet<>();
+
+        FluentIterable.from(eids)
+            .filter(Predicates.notNull())
+            .transformAndConcat(this.patientsFromEidFunction)
+            .copyInto(patients);
+
+        FluentIterable.from(getPatientsFromIds(ids))
+            .copyInto(patients);
+
+        try {
+            // Generate JSON for all retrieved patients.
+            final String json = OBJECT_MAPPER.writeValueAsString(patients);
+            return Response.ok(json, MediaType.APPLICATION_JSON_TYPE).build();
+        } catch (final JsonProcessingException ex) {
+            logger.warn("Failed to serialize patients [{}] to JSON: {}", eids, ex.getMessage());
+            return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
+        }
+    }
+
+    /**
+     * Gets the function that performs a query using the given patient eid, and returns an iterable containing the
+     * relevant patient entity objects.
+     *
+     * @return a function that gets patients associated with the provided external id
+     */
+    @Nonnull
+    private Function<Object, Iterable<PrimaryEntity>> getPatientsFromEidFunction()
+    {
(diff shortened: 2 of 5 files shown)


components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientsFetchResource.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.ws.rs.core.Response;
+
+/**
+ * Resource for working with multiple patient records, identified by their given "external" identifiers.
+ *
+ * @version $Id$
+ * @since 1.3M5
+ */
+@Path("/patients/fetch")
+@Relation("https://phenotips.org/rel/patientRecord")
```
sdumitriu: This isn't necessarily one patient record, so I would call it `"https://phenotips.org/rel/patientRecords"` (plural).

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Gets the function that gets the patient given the patient's ID.
+     *
+     * @return a function that gets the patient associated with the provided ID
+     */
+    @Nonnull
+    private Function<String, PrimaryEntity> getPatientFromIdFunction()
+    {
+        return new Function<String, PrimaryEntity>()
```
allasm: ah, right, did not notice that. It is O, but IMHO for such a simple task (at the end of the day we want to loop over patient IDs, get Patient objects, and add all non-null to a list) it does not make sense to bring in an external library and use fancy structures, even if they look a bit nicer/more modern

a-gvo: Regarding Guava, it's already included in the project. In my opinion, switching to loops, in this case, would make the code more bulky, and could potentially introduce new bugs.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+    private void addPatientFromId(@Nonnull final ImmutableSet.Builder<PrimaryEntity> patientsBuilder,
+        @Nonnull final Object id)
+    {
+        try {
+            // Try to get the patient entity.
+            final PrimaryEntity patient = this.repository.get((String) id);
+            // Get the current user.
+            final User currentUser = this.users.getCurrentUser();
+            // If user has view rights and patient with the provided ID exists, add patient to patient set.
+            if (patient != null && this.access.hasAccess(Right.VIEW, currentUser == null ? null
```
sdumitriu: Instead of manually checking rights here, you should `@Inject @Named("secure") PatientRepository`, which will throw a `SecurityException` when trying to get a patient that's not accessible.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * A custom serializer for primary entities.
+     */
+    private static final class PrimaryEntitySerializer extends JsonSerializer<PrimaryEntity>
+    {
+        @Override
+        public void serialize(final PrimaryEntity primaryEntity, final JsonGenerator jgen,
+            final SerializerProvider provider) throws IOException
+        {
+            jgen.writeRawValue(primaryEntity.toJSON().toString());
```
sdumitriu: I'd like to see the same links that are included in [the normal patient JSON](http://localhost:8080/rest/patients/P0000001). This does the trick:

    json.put("links", this.autolinker.get().forSecondaryResource(PatientResource.class, uriInfo)
        .withExtraParameters("patient-id", patient.getId()).build());

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+     * Retrieves and collects patient entities that correspond to the provided external ID.
+     *
+     * @param patientsBuilder a patient entity set builder
+     * @param eid an external patient ID, as string
+     */
+    private void collectPatientsFromEid(@Nonnull final ImmutableSet.Builder<PrimaryEntity> patientsBuilder,
+        @Nonnull final Object eid)
+    {
+        try {
+            final Query q = qm.createQuery("where doc.object(PhenoTips.PatientClass).external_id = :eid", Query.XWQL);
```
sdumitriu: Instead of running a query for every eid, you could just do:

```
q = qm.createQuery("from doc.object(PhenoTips.PatientClass) p where p.external_id in (:eids)", Query.XWQL);
q.bindValue("eids", eids);
q.execute()
```

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * A custom object mapper to facilitate serializing a list of {@link Patient} objects.
+     *
+     * @return an object mapper that can serialize {@link Patient} objects
+     */
+    private static ObjectMapper getCustomObjectMapper()
+    {
+        final ObjectMapper objectMapper = new ObjectMapper();
+        SimpleModule m = new SimpleModule(SERIALIZER_LABEL, new Version(1, 0, 0, "", SERIALIZER_LABEL, "json"));
```
sdumitriu: Still not right:

    public Version(int major, int minor, int patchLevel, String snapshotInfo, String groupId, String artifactId)

Should be something like:

    new Version(1, 0, 0, "", "org.phenotips", "PatientReferenceSerializer"));

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
a-gvo: Fixed all of the above.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+    @Before
+    public void setUp() throws ComponentLookupException, URISyntaxException
+    {
+        MockitoAnnotations.initMocks(this);
+        final Execution execution = mock(Execution.class);
+        final ExecutionContext executionContext = mock(ExecutionContext.class);
+        final ComponentManager componentManager = this.mocker.getInstance(ComponentManager.class, "context");
+        when(componentManager.getInstance(Execution.class)).thenReturn(execution);
+        doReturn(executionContext).when(execution).getContext();
+        doReturn(mock(XWikiContext.class)).when(executionContext).getProperty("xwikicontext");
```
sdumitriu: The lines from 130 to here aren't needed, they're boilerplate for an older way of getting the `XWikiContext`.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+
+        final String expectedStr = new JSONArray().put(new JSONObject().put("id", ID_1).put("links", this.uriList))
+            .toString();
+        @SuppressWarnings("unchecked")
+        final List<String> expected = (List<String>) (OBJECT_MAPPER.readValue(expectedStr, List.class));
+        @SuppressWarnings("unchecked")
+        final List<String> actual = (List<String>) (OBJECT_MAPPER.readValue((String) response.getEntity(), List.class));
+        assertTrue(expected.containsAll(actual) && actual.containsAll(expected));
+        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
```
sdumitriu: Duplicate with line 190.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+        doReturn(ID_1).when(this.patient1).getId();
+
+        final Response response = this.component.fetchPatients();
+
+        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+
+        final String expectedStr = new JSONArray().put(new JSONObject().put("id", ID_1).put("links", this.uriList))
+            .toString();
+        @SuppressWarnings("unchecked")
+        final List<String> expected = (List<String>) (OBJECT_MAPPER.readValue(expectedStr, List.class));
```
sdumitriu: I think you can do `assertTrue(expectedJson.similar(actualJson))` to simplify this kind of check.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+        final Request request = mock(Request.class);
+        doReturn(request).when(this.container).getRequest();
+        doReturn(eidList).when(request).getProperties("eid");
+        doReturn(idList).when(request).getProperties("id");
+
+        final Query query = mock(DefaultQuery.class);
+        when(this.patient1.toJSON()).thenReturn(this.patient1JSON);
+        when(this.patient2.toJSON()).thenReturn(this.patient2JSON);
+        when(this.patient3.toJSON()).thenReturn(this.patient3JSON);
```
sdumitriu: Could these lines be moved to `setup`?

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+        when(this.qm.createQuery(Matchers.anyString(), Matchers.anyString())).thenReturn(query);
+        when(query.execute()).thenReturn(ImmutableList.<Object>of(ID_1, ID_2));
+        doReturn(this.patient1).when(this.repository).get(ID_1);
+        doReturn(this.patient2).when(this.repository).get(ID_2);
+        doReturn(this.patient3).when(this.repository).get(ID_3);
```
sdumitriu: Could these 3 lines be moved to setup?

sdumitriu: You can override the behavior when needed, the last mocking call is the one that will be taken into account.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,355 @@
...
+        final String expectedStr = new JSONArray().put(new JSONObject().put("id", ID_1).put("links", this.uriList))
+            .put(new JSONObject().put("id", ID_3).put("links", this.uriList))
+            .put(new JSONObject().put("id", ID_2).put("links", this.uriList))
+            .toString();
+        @SuppressWarnings("unchecked")
+        final List<String> expected = (List<String>) (OBJECT_MAPPER.readValue(expectedStr, List.class));
+        @SuppressWarnings("unchecked")
+        final List<String> actual = (List<String>) (OBJECT_MAPPER.readValue((String) response.getEntity(), List.class));
+        assertTrue(expected.containsAll(actual) && actual.containsAll(expected));
+        assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
```
sdumitriu: Duplicate again.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
a-gvo: I think context is required for XWikiResource.

a-gvo: @sdumitriu could the logger's call to ```Utils.getXWikiUser``` in the ```initialize``` method of ```XWikiResource``` be the culprit?

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+     * @param id an internal patient ID
+     */
+    private void addPatientFromId(@Nonnull final ImmutableSet.Builder<PrimaryEntity> patientsBuilder,
+        @Nonnull final Object id)
+    {
+        try {
+            // Try to get the patient entity.
+            final PrimaryEntity patient = this.repository.get((String) id);
+            // If user has view rights and patient with the provided ID exists, add patient to patient set.
+            if (patient != null) {
```
allasm: currently there is no distinction between incorrect IDs and IDs for patients which the requestor has no permissions for. This is fine, as long as that is what we want: on lines 112-118 (https://github.com/phenotips/phenotips/pull/2163/files#diff-db33e60f31a18452c5507f7605e8385dR112) the only possibl responses are OK and INTERNAL_SERVER_ERROR.

I'm not sure we want to complicate our API with more error codes, but it may be useful as well, so worth a discussion/documentation.

...in any case this is ok for version 1

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
```diff
@@ -0,0 +1,294 @@
...
+        doReturn(executionContext).when(execution).getContext();
+        doReturn(mock(XWikiContext.class)).when(executionContext).getProperty("xwikicontext");
+
+        this.component = (DefaultPatientsFetchResourceImpl) this.mocker.getComponentUnderTest();
+
+        this.logger = this.mocker.getMockedLogger();
+        this.repository = this.mocker.getInstance(PatientRepository.class, "secure");
+
+        when(this.patient1.getId()).thenReturn(ID_1);
+        when(this.patient1.getDocument()).thenReturn(this.patientReference1);
```
allasm: Do we use these (all the patientReferenceXXX) anywhere? (I don't see any use but it may be used somewhere deeper which I don't see off the top of my head)

I'm asking because another PR (https://github.com/phenotips/phenotips/pull/2148) going into this release changes how we work with patient documents, so if we can remove this alltogether it will simplify merging the changes (or at least simplify figuring out if something has to be changed or not)

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
```diff
@@ -0,0 +1,222 @@
...
+        final ImmutableSet.Builder<PrimaryEntity> patientsBuilder = ImmutableSet.builder();
+
+        try {
+            addEids(patientsBuilder, eids);
+            addIds(patientsBuilder, ids);
+            // Generate JSON for all retrieved patients.
+            final String json = objectMapper.writeValueAsString(patientsBuilder.build());
+            return Response.ok(json, MediaType.APPLICATION_JSON_TYPE).build();
+        } catch (final JsonProcessingException ex) {
+            logger.warn("Failed to serialize patients [{}] to JSON: {}", eids, ex.getMessage());
```
allasm: I think this should be an `error` not a `warn` (which would be consistent with all other uses of error/warn in REST/controller code). Same in the other `catch`

a-gvo: Fixed.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImpl.java
a-gvo: I could add an else statement, and log it separately. What do you think?

allasm: Given that release is happening now/soon, I think we should leave it as is for now but create a JIRA issue "error reporting in REST API" or something like that and move the discussion there

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientsFetchResourceImplTest.java
sdumitriu: @a-gvo Yes, looks like it.

PR #2165, opened 2016-11-21
PT-2780: Generate list of corresponding genes based on the HPO terms entered in the patient record (PT-2879, PT-2874, PT-2876, PT-2875, PT-2970)

a-gvo:
@sdumitriu @buske Please review.


components/gene-panels/api/src/main/java/org/phenotips/panels/GeneCountsDto.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Set;
+
+/**
+ * A DTO containing a set of phenotypes associated with a specific gene, as well as the number of phenotypes.
+ *
+ * @version $Id$
+ * @since 1.3M5
+ */
+@Unstable("New API introduced in 1.3")
+public interface GeneCountsDto
```
buske: I'd suggest renaming this `PhenotypeSet` and removing its knowledge of the gene the phenotypes are associated with, since you always store that as the key of a Map to these objects.

components/gene-panels/api/src/main/java/org/phenotips/panels/GeneCountsDto.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return a set of HPO terms associated with a specific gene
+     */
+    Set<String> getHpoTerms();
+
+    /**
+     * Adds the HPO term to the set.
+     *
+     * @param hpoTerm the HPO term to be added
+     */
+    void addHpoTerm(final String hpoTerm);
```
buske: I'd suggest this be just `addTerm(termId)` and make the code as HPO-agnostic as possible.

components/gene-panels/api/src/main/java/org/phenotips/panels/GeneCountsDto.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return the total size of the HPO terms set
+     */
+    int getCount();
+
+    /**
+     * Gets the set of HPO terms associated with a specific gene.
+     *
+     * @return a set of HPO terms associated with a specific gene
+     */
+    Set<String> getHpoTerms();
```
buske: I'd suggest `getTerms()`

components/gene-panels/api/src/main/java/org/phenotips/panels/GenePanelsGeneCountsCreator.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+/**
+ * Provides access to gene panels.
+ *
+ * @version $Id$
+ * @since 1.3M5
+ */
+@Unstable("New API introduced in 1.3")
+public interface GenePanelsGeneCountsCreator
```
buske: I don't think `GenePanels` is a necessary prefix on classes like this. I'd suggest removing it.

components/gene-panels/api/src/main/java/org/phenotips/panels/GenePanelsGeneCountsCreator.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * Creates JSON representation of gene panels data.
+     *
+     * @return JSON representation fo gene panels data
```
buske: "of"

components/gene-panels/api/src/main/java/org/phenotips/panels/GenePanelsGeneCountsCreator.java
```diff
@@ -0,0 +1,49 @@
...
+     * @throws JsonProcessingException if JSON cannot be created
+     */
+    String createGeneCountsJson() throws JsonProcessingException;
+
+    /**
+     * Returns the provided list of present hpo IDs.
+     *
+     * @return the user provided list of HPO IDs
+     */
+    Set<String> getObservedSymptoms();
```
buske: I'd suggest `getPresentFeatures, to match https://github.com/phenotips/phenotips/blob/master/components/patient-data/api/src/main/java/org/phenotips/data/Feature.java

a-gvo:
@buske I've included the code review changes that you mentioned above, as well as what we've discussed later. For PT-2874, I didn't do a separate commit for the changes since it was changed quite a bit, and needs to be reviewed again. Sorry and thank you!

a-gvo:
Just a note to go with this pull request: currently pagination is implemented on the front-end. In a day or so this will be moved to the back end.

marta-:
@a-gvo can you fix the conflicts with master?

a-gvo:
@marta- Fixed

sashaandjic:
I can confirm that UI wise every subtask has been implemented. My only question now is the usability of the new section as it is manifested as an FWI section. Would it be useful to have ability to select individual genes from the list (like we have in the diagnosis suggestions), or perhaps selecting all of them? Also, some phenotypes come with the large set of genes (e.g. breast carcinoma) and it is nice that we offer all of them in multiple pages, however I am not sure if it is intuitive enough to the end user that there are e.g. 5 pages? The other thing is, do we need to present them in the certain order? One small thing: it would be nice to add a tooltip for the download button "Export gene counts in TSV format.".

marta-:
+1 for the tooltip to the download button.

I think displaying a message along the lines of "Showing results 1-10 of 5446" may help users figure out there's some pagination available.

I'm not a fan of how the pagination buttons prev & next have different sizes than first & last:
image

I'd use `fa-angle-* fa-lg fa-fw` instead of `fa-chevron-*` (with no `margin`) for prev and next. What do others think?

image

@sashaandjic : can you clarify "Do we need to add the source of our instant search (HPO)?"

sashaandjic:
@marta- for the diagnosis section we say "INSTANT OMIM SEARCH", and for genes "INSTANT GENE SEARCH". I though we may reword it to "INSTANT HPO SEARCH".

As for the pagination, we currently have "Page 1 of 8" in the right bottom corner.

sashaandjic:
The other inconsistency is that Gene Panels are displayed in their own section. With the current info available, it makes sense to move it to the Genotype Section, right after the LIST OF GENES.

sashaandjic:
Also moving the Phenotype description next to the "i" would save us some real estate. wdyt? 
screen shot 2017-01-05 at 3 20 21 pm

a-gvo:
@sashaandjic Gene Panels are in their own section because, from what I understood, the plan is to eventually add other gene-panel related functionality. 

Also, I'll look into sorting by phenotype :).

components/gene-panels/api/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+  <parent>
+    <groupId>org.phenotips</groupId>
+    <artifactId>gene-panels</artifactId>
+    <version>1.3-SNAPSHOT</version>
+  </parent>
+  <artifactId>gene-panels-api</artifactId>
+  <name>PhenoTips - Gene Panels - APIs</name>
+
+  <properties>
+    <coverage.instructionRatio>0.9</coverage.instructionRatio>
```
sdumitriu: This should always be entered with two decimals (as `0.90`), since the Jacoco plugin compares with the same precision specified here, and we want to know if it changes from 0.96 to 0.91.

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
```diff
@@ -113,6 +114,15 @@ public String getCitation()
     }
 
     @Override
+    protected Map<String, TermData> load(final String sourceUrl)
```
sdumitriu: This should instead use the upcoming vocabulary extension paradigm, once it's merged into master.

components/gene-panels/api/src/main/java/org/phenotips/panels/GenePanel.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.xwiki.stability.Unstable;
+
+import java.util.List;
+import java.util.Set;
+
+import org.json.JSONObject;
+
+/**
+ * An object containing methods for manipulating and displaying gene panels related data.
```
sdumitriu: This is uninformative, it describes technical things that can already be inferred (of course it's an object, and all objects contain methods for manipulating data), while it should say more about what a gene panel is. Something like this would be better:

> A Gene Panel links {@link #getPresentFeatures() a collection of observed symptoms} to {@link #getPhenotypesForGeneList() a list of genes} frequently associated with these symptoms, in descending order of relevance.

components/vocabularies/hpo/api/src/main/java/org/phenotips/vocabulary/internal/solr/HumanPhenotypeOntology.java
a-gvo: All code review changes addressed, with the exception of this one.

sashaandjic:
Retested. Tooltip added and genes are grouped by phenotype.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -4793,4 +4886,128 @@ input[type="checkbox"] + .LV_validation_message {
...
+}</code>
+    </property>
+    <property>
+      <contentType>CSS</contentType>
+    </property>
+    <property>
+      <name>Instant Gene Search</name>
+    </property>
+    <property>
+      <parse>1</parse>
```
sdumitriu: Should be `0`, there's no Velocity code inside.

components/gene-panels/api/src/test/java/org/phenotips/panels/internal/DefaultGenePanelFactoryImplTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+        Assert.assertEquals(GENE2, termsForGene.get(0).getGeneId());
+        Assert.assertEquals(GENE2, termsForGene.get(0).getGeneSymbol());
+        // The number of terms associated with "gene1" should be 1.
+        Assert.assertEquals(1, termsForGene.get(0).getCount());
+        // Check the term associated with "gene1" is the one that is expected.
+        Assert.assertEquals(presentTerms, termsForGene.get(0).getTerms());
+
+
+        Assert.assertEquals(GENE1, termsForGene.get(1).getGeneId());
```
sdumitriu: It's a coincidence, and tests fail on my computer. The result should always have a deterministic sorting.

PR #2167, opened 2016-11-25
PT-1787: Using two consecutive dashes -- in the patient's identifier field will display it striked out in view mode

sdumitriu:
Fixed as well as other fields affected by the same problem

@sashaandjic @allasm This affects almost all the patient record form, please test thoroughly.

### DIFF
--- a/components/base-war/src/main/webapp/templates/commentsinline.vm
+++ b/components/base-war/src/main/webapp/templates/commentsinline.vm
@@ -166,7 +166,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
     #if ($isAnnotation)
       <a href="#ID$comment.number"><blockquote class="annotatedText">$!escapetool.xml($commentSelection)</blockquote></a>
     #end
-    <div class="commentcontent#if($isAnnotation) annotationText#end">$commentDoc.getRenderedContentRestricted($comment.getProperty('comment').getValue(), $doc.getSyntax().toIdString())</div>
+    <div class="commentcontent#if($isAnnotation) annotationText#end">$escapetool.xml($comment.getValue('comment'))</div>
   </div>## xwikicomment
 #end
 ##
--- a/components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
@@ -72,7 +72,7 @@
   #if($xcontext.action == 'edit' || $xcontext.action == 'inline')
     {{html wiki=false clean=false}}&lt;input class="xwiki-free-multiselect-value" type="text" size="30" name="${prefix}${name}" id="${prefix}${name}_${counter}" value="$!escapetool.xml($v)"/&gt;{{/html}}##
   #else
-    $escapetool.xml($v)##
+    $!{services.rendering.escape($v, $doc.syntax)}##
   #end
 #end
 #if ($xcontext.action == 'edit' || $xcontext.action == 'inline')
--- a/components/widgets/ui/src/main/resources/PhenoTips/TabelarDataMacros.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/TabelarDataMacros.xml
@@ -69,7 +69,14 @@ $xwiki.ssx.use('PhenoTips.TabelarDataMacros')##
       #set ($disableLabels = true)
       #foreach ($o in $dataObjects)
         #set ($localCounter = 1)
-        #foreach($prop in $dataClass.properties)#if ($prop.name != $options.propertyName)|= ${prop.translatedPrettyName}: |#if ($localCounter == 1)=#end#set($localCounter = $mathtool.add($localCounter, 1)) $doc.display($prop.getName(), $displayMode, $o)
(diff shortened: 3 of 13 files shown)


components/base-war/src/main/webapp/templates/commentsinline.vm
```diff
@@ -166,7 +166,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
     #if ($isAnnotation)
       <a href="#ID$comment.number"><blockquote class="annotatedText">$!escapetool.xml($commentSelection)</blockquote></a>
     #end
-    <div class="commentcontent#if($isAnnotation) annotationText#end">$commentDoc.getRenderedContentRestricted($comment.getProperty('comment').getValue(), $doc.getSyntax().toIdString())</div>
+    <div class="commentcontent#if($isAnnotation) annotationText#end" style="white-space: pre-wrap;">$escapetool.xml($comment.getValue('comment'))</div>
```
marta-: It would be neater to put the css rule in the skin instead of a style attribute.

marta-:
The fix obviously captures much more than just dashes. It's probably a good idea to have a more generic issue to commit for (something like "unexpected display of values because of interpretation of wiki markup", but better phrased :) )

components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
```diff
@@ -72,7 +72,7 @@
   #if($xcontext.action == 'edit' || $xcontext.action == 'inline')
     {{html wiki=false clean=false}}&lt;input class="xwiki-free-multiselect-value" type="text" size="30" name="${prefix}${name}" id="${prefix}${name}_${counter}" value="$!escapetool.xml($v)"/&gt;{{/html}}##
   #else
-    $escapetool.xml($v)##
+    $!{services.rendering.escape($v, $doc.syntax)}##
```
allasm: Just curious what is the difference between `escapetool.xml` and `services.rendering.escape` and where should we use one  and when the other? E.g. in the previous file `escapetool.xml` was added, but here it is replaced

sashaandjic:
+1

components/base-war/src/main/webapp/templates/commentsinline.vm
sdumitriu: Done.

components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
allasm: Another question - why is .vm file on disk different from other velocity code?

And maybe we can create our own custom escape tool to avoid calling two different tools, which some developers may (which means "will")  forget to do?

components/widgets/ui/src/main/resources/PhenoTips/TabelarDataMacros.xml
```diff
@@ -69,7 +69,14 @@ $xwiki.ssx.use('PhenoTips.TabelarDataMacros')##
...
         #set ($localCounter = 1)
-        #foreach($prop in $dataClass.properties)#if ($prop.name != $options.propertyName)|= ${prop.translatedPrettyName}: |#if ($localCounter == 1)=#end#set($localCounter = $mathtool.add($localCounter, 1)) $doc.display($prop.getName(), $displayMode, $o)
+        #foreach($prop in $dataClass.properties)#if ($prop.name != $options.propertyName)|= ${prop.translatedPrettyName}: |#if ($localCounter == 1)=#end#set($localCounter = $mathtool.add($localCounter, 1))##
+          #if ("$!prop.getValue('contenttype')" == 'PureText' &amp;&amp; $xcontext.action != 'edit')
+            (((${doc.display($prop.name, $displayMode, $o).replaceFirst('^(\{\{html clean="false" wiki="false"}})?', '{{content syntax="plain/1.0"}}').replaceFirst('(\{\{/html}})?$', '{{/content}}')})))##
+          #elseif("$!prop.type" == 'StringClass' &amp;&amp; $xcontext.action != 'edit')
+            $!{services.rendering.escape($o.getValue($prop.name), $doc.syntax)}##
+          #else
+            ${doc.display($prop.getName(), $displayMode, $o).replaceAll('^(\{\{html clean="false" wiki="false"}})?&lt;p&gt;|&lt;/p&gt;(\{\{/html}})?$', '$1$2')}##
+          #end
```
allasm: Can you please clarify a bit what this change is about?

allasm:
Changes look OK (in the code) and work (when tested), but I feel like I may have easily missed something while doing code review

components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
sdumitriu: Files on disk don't support wiki markup, they just pass through Velocity and are copied to the output as-is. Content in wiki pages are passed through the rendering engine, and Velocity code in them must explicitly be placed inside `{{velocity}}` macros.

What do you mean by "custom escape tool"? `$escapetool` is the general escaping tool to be used, and it has useful methods for escaping different kinds of content (xml, javascript, json, url parameters, email headers). Ideally, it should also have a method for escaping wiki markup, but that's not possible for several reasons:

- The escape tool is defined in xwiki-commons, while the rendering engine is in xwiki-rendering, which is an upstream project depending on xwiki-commons, and the base project can't depend on the upstream, otherwise it would cause a circular dependency
- the rendering engine is a component, while Velocity tools are plain objects

In theory it is possible to define an "enhanced escapetool" and use it in PhenoTips, but I think it's too much effort for something that already works.

PR #2174, opened 2016-12-08
PT-2980: Global mode of inheritance and Global age of onset cannot be pushed / imported

sdumitriu:
Fixed

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GlobalQualifiersController.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GlobalQualifiersController.java
@@ -27,7 +27,6 @@ import org.phenotips.vocabulary.VocabularyTerm;
 import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.bridge.DocumentModelBridge;
 import org.xwiki.component.annotation.Component;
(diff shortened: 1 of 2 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GlobalQualifiersController.java
```diff
@@ -112,31 +119,31 @@
...
-                    ((DBStringListProperty) field).setList(termsToXWikiFormat(terms));
+            PropertyClass xpropertyClass = (PropertyClass) xclass.get(propertyName);
+            if (xpropertyClass != null) {
+                PropertyInterface xproperty = xpropertyClass.newProperty();
+                if (xproperty instanceof BaseStringProperty) {
+                    // there should be only one term present; just taking the head of the list
+                    dataHolder.set(propertyName, terms.isEmpty() ? null : termsToXWikiFormat(terms).get(0),
+                        this.xcontextProvider.get());
+                } else if (xproperty instanceof ListProperty) {
+                    dataHolder.set(propertyName, termsToXWikiFormat(terms), this.xcontextProvider.get());
```
veronikaslc: `this.xcontextProvider.get()` mentioned 4 times, may be it looks better to `XWikiContext context = this.xcontextProvider.get();` and then use `context` all over the lines

PR #2177, opened 2016-12-12
PT-2984: Fixed tabbing out of gene field if the correct gene is entered.

a-gvo:
@sashaandjic @marta- please review.


sashaandjic:
Now I can tab out of the gene field, but I have to press <tab> two times if I typed the full name. For partial gene name I always needed just one <tab> to leave the field and get the full gene name appear. Should this be applied on top of PR PT-2012?

sashaandjic:
+1

PR #2178, opened 2016-12-13
PT-2974 and PT-2971

allasm:


allasm:
@sashaandjic, please test

PR #2179, opened 2016-12-14
PT-2987: "Life status" field in the patient form

veronikaslc:


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -430,6 +430,40 @@ document.observe('xwiki:dom:loaded', function()
...
+        date_of_death_block.hide();
+        date_of_death.value = "";
+        death_date_entered.value="{}"
+        date_of_birth.value = "";
+        birth_date_entered.value="{}"
+     }
+  }
+
+  var life_status = $$("[id$='life_status']")[0];
+  displayDates(life_status.value);
```
marta-: This will throw an error if life status is disabled in the patient form. Always check if the element exists before reading any data from it.

veronikaslc:
@allasm please, review

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/LifeStatusController.java
```diff
@@ -58,9 +58,7 @@
 
     private static final String DECEASED = "deceased";
 
-    private static final String UNBORN = "unborn";
```
allasm: I recall we explicitly decided to add this on in the past (I think the logic was that a proband may be an unborn child with some detected phenotypes or genotypes, and that when parents come for help that would be the patient/proband)

Lets confirm with @marta- and Felicia that we don't want it any more (even if it was not supported in the UI before maybe it is a good time to add it)

allasm: ...the reason we never mentioned it is probably because everyone forgot we have already added it

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetMacros.xml
```diff
@@ -199,6 +199,45 @@
...
+#set ($format = $services.recordConfiguration.activeConfiguration.dateOfBirthFormat)
+#set ($definedFormat = "$!object.xWikiClass.get($name).getProperty('dateFormat').value")
+#if ($xcontext.action == 'edit')
+  #if ($value)
+    #set ($value = $services.jodatime.getMutableDateTime($value.time))
+    #set ($discard = $value.setZoneRetainFields($services.jodatime.UTCTimezone))
+  #end
+  &lt;input type='text' id='${prefix}${name}' title="$!escapetool.xml($format)" name='${prefix}${name}' class='hidden fuzzy-date' value="$!services.jodatime.getDateTimeFormatterForPattern($definedFormat).print($value)" alt="$!services.jodatime.ISODateTimeFormatter.print($value)"/&gt;&lt;input type='hidden' name='${prefix}${name}_entered' id='${prefix}${name}_entered' value="$!{escapetool.xml($object.getProperty("${name}_entered").value)}"/&gt;
+#else
+#set ($enteredValue = $object.getProperty("${name}_entered").value)
```
allasm: where does the `$name` come from? Would it be better to add it as an explicit macro parameter?

allasm: I think I figured it out, as it is this macro relies on being embedded as one of `customDisplay`-s of a property, in which case $name is implicitly defined. If I'm right, it is much better to change it to be an explicit parameter and/or rename the macro to `displayPropertyAsDate()` or something like that, since it looks like all of this code relies on $name being the property name, this macro cant be used to display a date that you have in say a variable `$date`, so the name may be confusing for those who don't know the back story

veronikaslc: @allasm macro renamed

components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R71501PhenoTips2991DataMigration.java
```diff
@@ -0,0 +1,187 @@
...
+                newProperty.setName(DATA_NAME);
+                newProperty.setValue(getLifeStatus(data));
+                data.addField(DATA_NAME, newProperty);
+
+                data.removeField(DATE_OF_DEATH_UNKNOWN);
+
+                patientXDocument.setComment(this.getDescription());
+                patientXDocument.setMinorEdit(true);
+            } catch (Exception e) {
+                this.logger.warn("Failed to update a global mode of inheritance property: {}", e.getMessage());
```
marta-: wrong warning message

PR #2184, opened 2016-12-21
PT-2783 diagnosis-redesign

tt6746690:
+ Redesign diagnosis suggestions using the PhenoGrid widget developed by Monarch Initiative.
+ The suggestions should be visible in both view mode and edit mode (right now they only appear in edit mode).
+ The new widget should still allow for turning on and off phenotypes to refine suggestions (just like the current one)
+ Users should be able to see all phenotypes associated with a disorder (just like the current one).


AlexLapena:
The text here looks a bit too close to the chart, it appears as if the box covers up the text. 

diag-redesign

sdumitriu:
Code moved to this repository, new PR #2197

PR #2189, opened 2017-01-12
PT-3015: After removing an A&W node, the A&W label stays in the pedigree; PT-3032: A&W symbol and adoption brackets overlap for females

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
```diff
@@ -1100,7 +1102,7 @@ define([
          */
         getAllGraphics: function($super) {
             //console.log("Node " + this.getNode().getID() + " getAllGraphics");
-            return $super().push(this.getHoverBox().getBackElements(), this.getLabels(), this._linkArea, this.getCarrierGraphics(), this.getEvaluationGraphics(), this.getHoverBox().getFrontElements());
+            return $super().push(this.getHoverBox().getBackElements(), this.getLabels(), this._awLabel, this._linkArea, this.getCarrierGraphics(), this.getEvaluationGraphics(), this.getHoverBox().getFrontElements());
```
allasm: This works here but is not right, as may cause other bugs. Instead, `this.getLabels()` should include `this._awLabel`. Other code may also expects all labels to be set in `getLabels()` (e.g. code to move labels under certain conditions)

allasm: This works here but is not right, as may cause other bugs. Instead, `this.getLabels()` should include `this._awLabel`. Other code may also expects all labels to be set in `getLabels()` (e.g. code to move labels under certain conditions)

veronikaslc:
@allasm added `this._awLabel` to `this.getLabels()`

veronikaslc:
Appeared that original fix was correct, so reverted it back

marta-:
@sashaandjic please test.

sashaandjic:
+1

PR #2190, opened 2017-01-13
Fixing regressions introduced by PT-2932 (PT-3021, PT-3022, PT-3031)

marta-:
In [this commit for PT-2932](https://github.com/phenotips/phenotips/commit/b791fcb56dced61ad895b1bb9a7b8396acb17726), the generation of the selection tools was moved from JavaScript to Velocity, but now the selection tools are generated ONLY in the push dialog, not the Export Excel/JSON dialog. However, [the JavaScript still tried to attach behaviors to the (non-existing) tools](https://github.com/phenotips/phenotips/commit/b791fcb56dced61ad895b1bb9a7b8396acb17726#diff-beef42dea801bc57f407ac4f98c830d4R317) which produced an error and prevented the rest of the code from running.

### DIFF
--- a/components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
+++ b/components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
@@ -229,8 +229,8 @@
  )))##section filters
 #end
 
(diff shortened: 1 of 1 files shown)


veronikaslc:
tested

allasm:
also tested

sashaandjic:
I was not able see the flashing problem on my instance. PT-3021, PT-3022 and PT-3031 tested and working as expected.

PR #2192, opened 2017-01-18
[misc] Committed the Ethnicity parser

sdumitriu:


components/vocabularies/ethnicity/parser/src/main/java/org/phenotips/vocabulary/internal/solr/EthnicityParser.java
```diff
@@ -0,0 +1,93 @@
...
+            Document doc = Jsoup.parse(listFile, "UTF-8");
+            Elements rows = doc.select("tr");
+            PrintWriter output = new PrintWriter(outputName);
+            PrintWriter outputXml = new PrintWriter(outputNameXml);
+            outputXml.println("<add>");
+            for (Element row : rows) {
+                String ethnicity = row.select("a").get(0).text();
+                int number = 1000;
+                try {
+                    Matcher matcher = Pattern.compile("^<?\\s*([0-9,]+)([^0-9]+)?([0-9,]+)?")
```
marta-: Please comment this block of code to explain the regex, why you use the 3rd and the 1st matched group, etc.

components/vocabularies/ethnicity/parser/src/main/java/org/phenotips/vocabulary/internal/solr/EthnicityParser.java
```diff
@@ -0,0 +1,93 @@
...
+
+        try {
+            Document doc = Jsoup.parse(listFile, "UTF-8");
+            Elements rows = doc.select("tr");
+            PrintWriter output = new PrintWriter(outputName);
+            PrintWriter outputXml = new PrintWriter(outputNameXml);
+            outputXml.println("<add>");
+            for (Element row : rows) {
+                String ethnicity = row.select("a").get(0).text();
+                int number = 1000;
```
marta-: Please comment to explain this number.

sdumitriu: @marta- Fixed, better now?

marta-: yes, thanks.

sdumitriu:
@sashaandjic There are no UI changes, but the suggestions should be somewhat better, for example spelling mistakes should be corrected.

sdumitriu:
@sashaandjic Actually that comment is for #2196, this is code that's not used in any way at runtime.

sashaandjic:
@sdumitriu Is there any possibility to have suggestions sorted alphabetically? They appear as sorted, however if you look past first character they are not.

sdumitriu:
That's not a good idea. They are sorted in order of population size, since it's more likely that the user wants to say "Russian" than "Ra Glai".

sashaandjic:
Ah, that does make sense.

PR #2195, opened 2017-01-21
[misc] Create controllers for features and disorders

veronikaslc:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/DisordersController.java
```diff
@@ -0,0 +1,228 @@
...
+        } catch (Exception e) {
+            this.logger.error("Failed to access patient data for [{}]: {}", patient.getDocument(), e.getMessage());
+        }
+        return null;
+    }
+
+    @Override
+    public void writeJSON(Patient patient, JSONObject json, Collection<String> selectedFieldNames)
+    {
+        if (selectedFieldNames != null && !selectedFieldNames.contains(DISORDER_PROPERTIES)) {
```
sdumitriu: I don't think this works, should be `contains(DISORDER_PROPERTIES_OMIMID)`.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/DisordersController.java
```diff
@@ -0,0 +1,228 @@
...
+        } catch (Exception e) {
+            this.logger.error("Could not load disorders from JSON: [{}]", e.getMessage(), e);
+            return null;
+        }
+    }
+
+    @Override
+    public void save(Patient patient, DocumentModelBridge doc)
+    {
+        PatientData<Disorder> features = patient.getData(this.getName());
```
sdumitriu: Should be called `disorders`, not `features`.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/DisordersController.java
veronikaslc: Then why we have it like that here https://github.com/phenotips/phenotips/blob/master/components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java#L355 ?

veronikaslc:
@sdumitriu comments addressed

PR #2196, opened 2017-01-22
PT-3043: Improve ethnicities search

sdumitriu:


sashaandjic:
Much better!

sdumitriu:
@sashaandjic can you check if PT-1189 is also fixed?

sashaandjic:
PT-1189 seemed better prior to this change.

PR #2198, opened 2017-01-25
Entities-NIAID: Added new repo + distrib config to entities pom

sebialex:
@sdumitriu please review when you have time, thanks


sdumitriu:
This isn't needed, you can manually speccify the deployment server on the command line. I'd rather not put this reference upstream unless you really want it.

sebialex:
Ok, it was mostly for convenience, I don't want to remember some config that I use only once every few months. And I figured that this branch will never be merged into anything else and eventually will be deleted.

But that's ok, I'll just copy the config from another repo when needed.

PR #2201, opened 2017-02-07
PT-3059: Variants not properly imported when doing 1.3M1 to 1.3M5 JSON import

veronikaslc:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -382,11 +396,23 @@ public void writeJSON(Patient patient, JSONObject json, Collection<String> selec
         return null;
     }
 
+    /**
+     * Supports both 1.3-m5 and older 1.3-xx format. 1.3-m5 and newer variant JSON format: {"gene": ENSEMBL_Id [, ...] }
+     * 1.3-old format: {"gene": HGNC_Symbol [, ...] }
```
sdumitriu: Should be `"genesymbol"`.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractSolrVocabulary.java
```diff
@@ -141,7 +141,9 @@ public VocabularyTerm getTerm(String id)
 
         Set<VocabularyTerm> result = new LinkedHashSet<>();
         for (String id : ids) {
-            result.add(rawResult.get(id));
+            if (rawResult.containsKey(id) && rawResult.get(id) != null) {
```
sdumitriu: This is for another issue, please move out.

PR #2210, opened 2017-02-14
PT-2963: Make language and look of the Push modal more user friendly

veronikaslc:


components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -1,16 +1,12 @@
...
+phenotips.patientSharing.push.dialog.remoteGroup.label=Would you like to add this patient to one of __serverID__ groups?
+phenotips.patientSharing.push.dialog.remoteGroup.hint=If a group is selected, the group will be marked as the owner of the patient record and you will be marked as a collaborator. If no group is selected you will be marked as the owner of the patient record..
 phenotips.patientSharing.push.dialog.preview.label=Patient data preview
 phenotips.patientSharing.push.dialog.preview.hint=Please review the data that will be sent to the server.
 phenotips.patientSharing.push.dialog.confirmation.label=I agree to push selected patient data to
-phenotips.patientSharing.push.dialog.title.multiplePatients=Pushing patient data to a remote PhenoTips instance\u2026
-phenotips.patientSharing.push.dialog.title.singlePatient=Pushing __patientID__ to a remote database\u2026
+phenotips.patientSharing.push.dialog.title.multiplePatients=Pushing patient data to __serverID__\u2026
+phenotips.patientSharing.push.dialog.title.singlePatient=Pushing __patientID__ to __serverID__\u2026
```
a-gvo: Do these modified strings need to be removed from property files for other languages?

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -1,16 +1,12 @@
...
-phenotips.patientSharing.push.dialog.authentication.hint=Please choose the user name to be used to authenticate on the server and send the data; this user will be marked as the reporter on the remote server.
-phenotips.patientSharing.push.dialog.remoteGroup.label=Remote PhenoTips group
-phenotips.patientSharing.push.dialog.remoteGroup.hint=The PhenoTips group you would like to share this patient with on the remote server. If a group is selected, the group would be the owner and the given remote user will be a collaborator. If no group is selected, the remote user will be the only owner.
+phenotips.patientSharing.push.dialog.authentication.label=Log in to __serverID__
+phenotips.patientSharing.push.dialog.authentication.hint=Please log in to __serverID__ using your __serverID__ username and password.
+phenotips.patientSharing.push.dialog.remoteGroup.label=Would you like to add this patient to one of __serverID__ groups?
+phenotips.patientSharing.push.dialog.remoteGroup.hint=If a group is selected, the group will be marked as the owner of the patient record and you will be marked as a collaborator. If no group is selected you will be marked as the owner of the patient record..
```
a-gvo: There's an extra period at the end.

a-gvo:
@veronikaslc Not sure if I'm doing something wrong, but no dialogue comes up when I try to push.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources_ar.properties
```diff
@@ -1,8 +1,4 @@
...
-phenotips.patientSharing.push.dialog.serverSelection.hint.multiplePatients=\u062e\u0627\u062f\u0645 \u0641\u064a\u0646\u0648\u062a\u064a\u0628\u0633 \u0627\u0644\u0628\u0639\u064a\u062f \u062d\u064a\u062b \u0633\u064a\u062a\u0645 \u0625\u0631\u0633\u0627\u0644 \u0643\u0627\u0641\u0629 \u0627\u0644\u0628\u064a\u0627\u0646\u0627\u062a.
-phenotips.patientSharing.push.dialog.serverSelection.label.singlePatient=\u0625\u0644\u0649 \u0623\u064a\u0646 \u064a\u062a\u0645 \u0625\u0631\u0633\u0627\u0644 \u0645\u0644\u0641 \u062a\u0639\u0631\u064a\u0641 \u0647\u0630\u0627 \u0627\u0644\u0645\u0631\u064a\u0636\u061f
-phenotips.patientSharing.push.dialog.serverSelection.hint.singlePatient=\u062e\u0627\u062f\u0645 \u0641\u064a\u0646\u0648\u062a\u064a\u0628\u0633 \u0627\u0644\u0628\u0639\u064a\u062f \u062d\u064a\u062b \u0633\u064a\u062a\u0645 \u0625\u0631\u0633\u0627\u0644 \u0628\u064a\u0627\u0646\u0627\u062a \u0647\u0630\u0627 \u0627\u0644\u0645\u0631\u064a\u0636.
```
marta-: Just FYI, you don't have to delete deprecated translations from other languages anymore. Our Crowdin integration removes them automatically and syncs all the keys with the default language.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
a-gvo: And @marta- answered my question below :).

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -49,19 +45,17 @@ phenotips.PushPatient.male=Male
...
+phenotips.PushPatient.patientUploadedHours=This patient was uploaded to __serverID__ __hoursN__ hour(s) ago and was assigned patient ID __patientID__
 phenotips.PushPatient.patientUploadedUpdate=If data in the record has changed, push to __serverID__ again to update the record there.
 phenotips.PushPatient.updateRemotePatient=Update remote patient
 phenotips.PushPatient.createRemotePatient=Create new remote patient
 phenotips.PushPatient.loadingServers=Loading server list\u2026
 phenotips.PushPatient.serverDesc=Selected server does not allow updating existing patients, so can only push again to a new patient record.
 phenotips.PushPatient.successfulUpdate=Updated patient successfully.
-phenotips.PushPatient.successfulPush=Pushed patient successfully, ID of the new patient on the remote server is
-phenotips.PushPatient.pushFailedUpdatesDisabled=Unable to update this patient - updates are disbaled on the remote server
+phenotips.PushPatient.successfulPush=Pushed patient successfully, ID of the new patient on the remote server is __patientID__
```
marta-: Let's change this to something like `The patient was successfully submitted and its identifier on __serverName__ is __patientID__.`

Note that apparently "remote server" doesn't mean anything to our users, so we must be specific about which "server" we're talking about. Apologies about the incomplete description of the issue, but hopefully during review we can catch and fix at least some of the "remote server" occurrences that were missed.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -49,19 +45,17 @@ phenotips.PushPatient.male=Male
...
+phenotips.PushPatient.pushFailedUpdatesDisabled=Unable to update this patient - updates are disabled on the remote server
```
marta-: Since we're cleaning up translations, this should probably read `Updating this patient failed because updates are disabled on __serverName__.`

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -193,19 +193,15 @@
...
-      var serverDesc  = this._pushManyPatients ? "$services.localization.render('phenotips.patientSharing.push.dialog.serverSelection.hint.multiplePatients')":
-                                                 "$services.localization.render('phenotips.patientSharing.push.dialog.serverSelection.hint.singlePatient')";
       this._container = new Element('form', {'id' : 'push-patients-ui', 'class' : 'xform', 'method' : 'post', 'action' : this._serviceURL});
-      this._container.insert(generateSection('server-selection', serverTitle, serverDesc))
-                     .insert(generateSection('user-selection', "$services.localization.render('phenotips.patientSharing.push.dialog.authentication.label')","$services.localization.render('phenotips.patientSharing.push.dialog.authentication.hint')"))
+      this._container.insert(generateSection('user-selection', "$services.localization.render('phenotips.patientSharing.push.dialog.authentication.label')","$services.localization.render('phenotips.patientSharing.push.dialog.authentication.hint')"))
```
marta-: Would be nice to start using `$!escapetool.javascript(...)` to escape all translated text when we modify js code, even if it wasn't done originally.

marta-:
Another suggested (not requested) change (still awaiting team consensus) is that it would be nice to start using the standard way of specifying parameters in strings, which is `{n}`, instead of what we usually do in js files `__namedParameter__`, both for consistency and to avoid confusing our translators.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
veronikaslc: Here are other occurrences:
`phenotips.patientSharing.push.dialog.consents.warning=Consents that need to be granted before pushing could not be found. This might prevent you from pushing this patient to a remote server`
`phenotips.PushPatient.serverAndUserFirstMessage=Please login to the remote server first`
`phenotips.PushPatient.oneServerMessage=There is only one configured remote server: `
`phenotips.PushPatient.pushFailedUpdatesDisabled=Unable to update this patient - updates are disabled on the remote server`
`phenotips.PushPatient.pushFailedInvalidGuid=Unable to update this patient - stored remote GUID is incorrect. Maybe the patient was deleted on the remote server`

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources_ar.properties
veronikaslc: awesome :)

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
marta-: Thanks @veronikaslc . My suggestions:
```
phenotips.patientSharing.push.dialog.consents.warning=Consents that need to be granted before pushing could not be found. This might prevent you from submitting this patient.
phenotips.PushPatient.serverAndUserFirstMessage=You must be logged in to see these options.
phenotips.PushPatient.pushFailedInvalidGuid=Updating failed because the corresponding patient record to update could not be found on {0}. It is possible that the record was deleted from {0}.
```
I'm not sure about this last one, maybe Felicia should chime in.

I don't think this next one is actually used anywhere anymore, I can't find it:
```
phenotips.PushPatient.oneServerMessage=There is only one configured remote server:
```

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
veronikaslc: commit amended to exclude unnecessary files

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
veronikaslc: its used here https://github.com/phenotips/phenotips/blob/master/components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml#L829

veronikaslc:
@a-gvo @marta- commit amended with codereview changes

sashaandjic:
Found two problems:

1. See the highlighted text
screen shot 2017-02-17 at 5 32 16 pm

2. See the highlighted text ( it says groups - it should say group) 
screen shot 2017-02-17 at 5 33 46 pm

sashaandjic:
Problem #1 fixed. Push dialog is looking really good. Problem #2 (group vs groups) isn't a big issue, but would be nice to have corrected.

screen shot 2017-02-17 at 11 33 14 pm

sashaandjic:
One more thing: It should say "Keep existing group" not "groups".

sashaandjic:
2 small issues found:
- replace None with username
- remove phenotips.patientSharing.push.dialog.remoteGroup.hint

screen shot 2017-02-21 at 8 12 20 pm

sashaandjic:
Another problem discovered, once push was successfully completed, the message at the bottom is badly formatted :
screen shot 2017-02-22 at 1 52 04 pm

One potential issue that can confuse the end user is the highlighted message in the screenshot below. Such message should be displayed (if needed) when the push dialog is initially open. Upon successful push, the message should disappear as it only brings confusion. 
screen shot 2017-02-22 at 1 52 28 pm

sashaandjic:
No more complaints. It looks fantastic!

sashaandjic:
Problem found: after switching to another remote server, some text in the dialog is still referencing to the  server formerly used. In the example below I first chose Tim Hortons to push the data; next time when I wanted to push to Tarantula, Tim Hortons' name was showing in 2 places.
screen shot 2017-02-23 at 9 04 51 am

sashaandjic:
@veronikaslc reminders: If user isn't logged on the remote server, display text stating "Please log in to PhenomeCentral..."

veronikaslc:
IMHO `I agree to push selected patient data to ...`  doesn't make much sense because nothing is selected yet. Either it needs to be placed in the next selection data dialog or rephrased.

marta-:
@veronikaslc Good point. I vote for moving it to the next dialog.

marta-:
What about consents?

veronikaslc:
I vote for rephrasing. `I agree to push patient [patient_id] to [remote_server]`

sashaandjic:
Sounds good to me.

veronikaslc:
or more generic ' agree to push patient data to [remote_server]' Thus it fits for both single and multiple patients.

sashaandjic:
+1 for 'agree to push patient data to..'

veronikaslc:
changes committed

sashaandjic:
How much work is to move all of them to the next dialog given that both consent and "I agree.." checkboxes have logic that needs to be moved and tested too?

veronikaslc:
Next dialog has a lot to display already, I'd rather leave it in first one.

sashaandjic:
One thing bothers me: if I didn't select mandatory consent it still lets me go select fields and then tels me "On you have no consent". We should prevent opening of the second dialog if mandatory consents are not granted.

sashaandjic:
I tested your latest changes. It looks really good!

veronikaslc:
@sashaandjic @marta- Can we have consent assigned to specific data? If so, then it make sense to check consent after data selected.

sashaandjic:
It is the other way around. Having no mandatory consent, means no data to show. Other consents may be required to show specific sections. I think it is ok to leave them on the first dialog. The only thing that I would change is this: if mandatory consents are not checked and user wants to proceed, display a warning and prevent users from launching the other dialog. Andrey already implemented the whole thing, but it happens a bit too late, as we let user select fields and then raise an error. There is also RETRY button that is not functionig very well (e.g. if I check the missing consent, the RETRY button disappears and I have to select the fields again).

veronikaslc:
We do not have a separate API for checking required consent. So there is no way now to check required consent before push happens. Implementing this will require additional time.

sashaandjic:
ok.

components/patient-data-sharing/push-api/src/main/resources/ApplicationResources.properties
```diff
@@ -78,8 +66,8 @@ phenotips.PushPatient.pushPhenomeCentral=Push to PhenomeCentral
 phenotips.PushPatient.pushToAnotherServer=Push to {0}
 phenotips.PushPatient.confirmationText=Save before exporting?
 phenotips.PushPatient.button.cancel=Cancel
-phenotips.PushPatient.button.singlePatient.push=Next: Select data fields to be pushed
-phenotips.PushPatient.button.manyPatients.push=Next: Select patients and data fields to be pushed
+phenotips.PushPatient.button.singlePatient.push=Next
+phenotips.PushPatient.button.manyPatients.push=Next
```
sdumitriu: These could be squashed into a single key.

sdumitriu:
@veronikaslc @sashaandjic Is this ready for merging?

sashaandjic:
@sdumitriu Yes, as far as I am concerned.

PR #2211, opened 2017-02-14
PT-3070: 'Assign patient to a family' modal isn't completely translated

veronikaslc:


components/family-studies/ui/src/main/resources/PhenoTips/FamilySelector.xml
```diff
@@ -144,15 +144,15 @@
             this.createFamily.observe('click', this._onOptionSelected.bind(this) );
             this.assignFamily.observe('click', this._onOptionSelected.bind(this) );
 
-            createLabel.insert(this.createFamily).insert("Create a new family");
+            createLabel.insert(this.createFamily).insert("$services.localization.render('phenotips.UIXField.family_membership.create')");
```
marta-: Please always use `"$!escapetool.javascript($services.localization.render('...'))"` to prevent any special characters that may occur in translated strings from breaking the whole script.

veronikaslc: should I also add it to other translations in javascript extensions?

marta-: Whenever you have to modify a script and notice that doesn't have escaping, definitely yes. If you want to actually go through all the scripts and add escaping now, that would be awesome, but I can imagine it would take a while.

sashaandjic:
I still cannot see the spanish translation. @veronikaslc said that we are missing actual translations enev though the support for them is there.

marta-:
@sashaandjic Sorry, my mistake to have assigned this to you to test when there aren't any Spanish translations yet, so the text will default to English.

PR #2212, opened 2017-02-15
PT-3079: Unable to create/update patient record with phenotypes without specifying nonstandard_features[] through REST API

veronikaslc:


marta-:
I didn't test, but that sure looks like it will do the trick :)

PR #2213, opened 2017-02-15
PT-3013: Pedigree editor form improvements

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
```diff
@@ -1354,17 +1354,21 @@ define([
 
                 var suggestContainer = container.down('div.patient-newlink-container');
                 var suggestInput     = container.down('input[type=text].suggest-patients');
+                suggestInput.placeholder = " type patient name or identifier";
```
marta-: Why does it start with a space and small `t`?

veronikaslc: imho looks better that way

marta-: @veronikaslc If a space looks better for the placeholder, then it probably also looks better when there's a value entered, in which case the effect should be obtained via styling (padding left & right) instead.

I don't have a strong opinion regarding small vs capital first letter, as link as it's consistent across placeholders on the same form.

veronikaslc: I see, it's actually looks well without that space, so I'll just remove it
Regarding placeholders, in patient form we have it with both capital and small first letters, I personally prefer small, but it's super minor.

sashaandjic:
+1

PR #2214, opened 2017-02-15
PT-3078: Changed the label of the "Download" button of the Gene Panel…

a-gvo:
…s section to "Download as TSV".


marta-:
@a-gvo I've pushed some changes to this branch, as per the PT-3078 comment thread (discussion that happened after your initial commit).

marta-:
@veronikaslc please re-review

sashaandjic:
One suggestion: when hovering over ? the tooltip "How to enter this information" appears. It would be better to say something like "What is this?".

marta-:
@sashaandjic i agree. The default widget has this tooltip and no api to customize it. We could change the default for the entire form i guess if it makes sense.

sashaandjic:
ok

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -4993,8 +4993,8 @@ input[type="checkbox"] + .LV_validation_message {
   font-size: .85em;
   margin: 0 auto;
 }
-#genePanelsExport {
-  font-size: .6em;
+#genePanelsExport, .gene-panels-search-box h4 .xHelpButton, .gene-panels-search-box h4 .xTooltip {
+  font-size: .65em;
```
sdumitriu: Magic value...

PR #2215, opened 2017-02-15
PT-3077: Renamed "Gene Panels" section to "Diagnostic Suggestions".

a-gvo:


marta-:
Thanks @a-gvo , but I'm closing this without merging because the issue description was incomplete. I'm also closing PT-3077 in favor of PT-3075.

Basically, if we're going to use this generic section title, it makes sense to also move the OMIM suggestions in the "Diagnostic Suggestions" section, as discussed in the Monday standup. And since this section will present more than gene suggestions, then the internal identifiers, extension points and translation keys need to be updated to say something other than `genePanels` for better semantics.

PR #2216, opened 2017-02-15
PT-3072: Create a PhenoTipsGene class to hold gene info; PT-3082: Excel export list EnsembleIDs in "Gene name" and "Gene symbol" columns

veronikaslc:


sashaandjic:
Tested: Gene IDs and gene symbols are now present in Excel exports. I also noticed that we now export phenotype metaIDs as well!

allasm:
Based on the dev. meeting discussion we should make carrier genes editable in the pedigree (like candidate/solved but unlike negative which are read only)

components/export/api/src/main/resources/ApplicationResources.properties
```diff
@@ -48,7 +49,7 @@ phenotips.export.excel.label.phenotype.meta.name=Meta
 phenotips.export.excel.label.phenotype.meta.id=ID
 
 phenotips.export.excel.label.genotype=Genotype
-phenotips.export.excel.label.genotype.name=Gene Name
+phenotips.export.excel.label.genotype.name=Gene ID
```
veronikaslc: @allasm It is detailed in the issue description https://phenotips.atlassian.net/browse/PT-3082

components/export/ui/src/main/resources/PhenoTips/ExportJSON.xml
```diff
@@ -41,7 +41,7 @@
 
 {{velocity wiki="true"}}{{{
 #set ($codeProperties = ['phenotype', 'negative_phenotype', 'prenatal_phenotype', 'omim_id'])
-#set ($nonPatientClassProperties = ['doc.name', 'referrer', 'author', 'creationDate', 'date', 'genes', 'genes_status', 'genes_strategy', 'genes_comments', "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity", "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence", "variants_segregation", "variants_sanger"])
+#set ($nonPatientClassProperties = ['doc.name', 'referrer', 'author', 'creationDate', 'date', 'genes', 'genes_symbol', 'genes_status', 'genes_strategy', 'genes_comments', "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity", "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence", "variants_segregation", "variants_sanger"])
```
veronikaslc: gene ID is included by default, same as varient cdna field

components/patient-data/api/src/main/java/org/phenotips/data/Gene.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    String getId();
+
+    /**
+     * Return gene symbol.
+     *
+     * @return symbol
+     */
+    @Override
+    String getName();
```
veronikaslc: we use `name` for gene symbol everywhere in UI, for example `Phenotips.GeneClass.geneNamePlaceholder=Enter a gene name`. @marta- should we rename 'name' with `symbol` everywhere in UI and in the code?

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsGene.java
```diff
@@ -0,0 +1,324 @@
...
+        }
+    }
+
+    /**
+     * Sets gene Ensembl ID and status via HGNC Vocabulary term.
+     *
+     * @param geneName either gene ensembl id or gene symbol
+     */
+    @SuppressWarnings("unchecked")
+    public void setNames(String geneName)
```
veronikaslc: here we are setting both ID and name. @marta- which method name is better?

veronikaslc:
@marta- what would be the use of reading property names directly from the XWiki Document? We do not treat them uniformly anywhere.

veronikaslc:
2 tests still failing, to test run -Pquick

components/patient-data/api/src/main/java/org/phenotips/data/Gene.java
allasm: we may refer to those as "names" in UI, but in the code `name` is a very generic overused name (duh). I see you rename some methods which had a `symbol` in their name so at least now it is more consistent, but IMHO it woudl be easier to maintian/understand if we use `symbol` instead everywhere

veronikaslc:
rebased

sashaandjic:
Should this one be retested?

veronikaslc:
@sashaandjic later on, I just rebased it, need to retest and fix json tests to stop random failing

PR #2217, opened 2017-02-17
PT-3084: create 'carrier' gene status option (Genotype info>list of genes)

veronikaslc:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
```diff
@@ -116,7 +118,7 @@
     private static final String JSON_REJECTEDGENES_KEY = "rejectedGenes";
 
     private static final List<String> STATUS_VALUES = Arrays.asList(INTERNAL_CANDIDATE_VALUE, INTERNAL_REJECTED_VALUE,
-        INTERNAL_SOLVED_VALUE);
+        INTERNAL_SOLVED_VALUE, INTERNAL_CARRIER_VALUE);
```
marta-: This comment has nothing to do with the 'carrier' issue, but hardcoding property values in the controller is wrong and, as we can see here, difficult to maintain. The value should be loaded from the XClass.

components/patient-data/indexing/pom.xml
```diff
@@ -29,7 +29,7 @@
   <name>PhenoTips - Patient data - Solr indexing</name>
 
   <properties>
-    <coverage.instructionRatio>0.96</coverage.instructionRatio>
+    <coverage.instructionRatio>0.94</coverage.instructionRatio>
```
marta-: :(

marta-:
@veronikaslc I haven't tested yet, but as far as i understand from the code it doesn't look like carriers are represented in the pedigree according to the standard. They should be dots in the center of the individual's shape (or near the center, if an individual is a carrier for more than one gene), and the color of the dot for Gene X should always be the same as the color associated to confirmed causative Gene X (if this confirmed gene exists in the pedigree).

veronikaslc:
@marta- didn't know that is needed, will do. We probably have to specify expected behaviour in more details.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
veronikaslc: That's why I have a PR for GeneClass :)

marta-: But [the values are still hardcoded in that PR](https://github.com/phenotips/phenotips/pull/2216/files#diff-665baeaa42f5fff78d584ea090d111e2R61)...

allasm:
@marta - I think we can implement pedigree graphic for gene carriers in v2. Currently the carrier "dot" graphic is overloaded, as it may refer to either genes, disorders or cancers. And since all of those will be modified in PhenoTips (cancers added, carrier status added for disorders, as far as I know?) it probably makes more sense to modify pedigree once all those are done.

Also if you mean to add dots in the middle of gene section (instead of one dot in the middle) that should probably be a separate task and done after the main "create 'carrier' gene status" task is competed

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/GeneListController.java
veronikaslc: It seems more logically to do it in in the PR for GeneClass.

components/patient-data/indexing/pom.xml
veronikaslc: added test, coverage restored

marta-:
@veronikaslc @allasm okay

components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
```diff
@@ -139,6 +139,13 @@ define([
                 'function' : 'setRejectedGenes' // not needed: read only for now
             },
             {
+                'name' : 'carrier_genes',
+                'label' : 'Genotype: carrier genes',
+                'type' : 'gene-picker',
+                'tab': 'Clinical',
+                'function' : 'setCarrierGenes' // not needed: read only for now
```
allasm: wrong comment, it is not read-only?

components/pedigree/resources/src/main/resources/pedigree/view/carrierGeneLegend.js
```diff
@@ -0,0 +1,20 @@
...
+ * @class CarrierGeneLegend
+ * @constructor
+ */
+ define(["pedigree/view/geneLegend"], function(GeneLegend){
+    var CarrierGeneLegend = Class.create( GeneLegend, {
+
+        initialize: function($super) {
+            $super('Carrier Genes', 'genes',
+                   "carrier",
+                   [], // these are never displayed in a legend so don't need colours
```
allasm: Are you sure we don't want to display them in a legend/on the node with colors?

components/pedigree/resources/src/main/resources/pedigree/view/person.js
```diff
@@ -998,7 +1008,12 @@ define([
             return this._getGeneArray("rejected");
         },
 
-        // returns null if geneID is not presemnt, or gene status if it is
+        // used by controller in conjuntion with setCandidateGenes
```
allasm: wrong copy-paste comment

allasm:
Based on the dev. meeting discussion we should make carrier genes editable in the pedigree (like candidate/solved but unlike negative which are read only)

components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
veronikaslc: it is read-only

components/pedigree/resources/src/main/resources/pedigree/view/carrierGeneLegend.js
veronikaslc: in future

components/pedigree/resources/src/main/resources/pedigree/view/person.js
veronikaslc: we have that comment for `getCausalGenes()` and `getRejectedGenes()`, if it is valid there why it is not valid for `getCarrierGenes()`?

veronikaslc:
@marta- Did not we decide to make carrier genes editable it in future, not this PR?

veronikaslc:
@allasm made a commit with codereview comments

sashaandjic:
- carrier genes are not showing up in the legend and they are not coloured

veronikaslc:
@sashaandjic comments addressed

sashaandjic:
Tested! No problems found.

sdumitriu:
@veronikaslc @marta- @allasm Is this ready to be re-reviewed?

veronikaslc:
Yes it is ready.

PR #2218, opened 2017-02-17
PT-2843: On the Family page, the family identifier label and the corr…

marta-:
…esponding field should be closer

Quick fix to stack the label and the value vertically. This layout fits better with the current family form. Should be revisited once we redo the whole PhenoTips UI.

### DIFF
--- a/components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
+++ b/components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
@@ -112,20 +112,11 @@
 
 {{velocity}}
 #if($xcontext.action == 'edit' || $doc.display('external_id') != '')
-  #if ($xcontext.action != 'edit')
-    (((#__form_label('external_id')
-       #__form_field('external_id'))))
-  #else
-    (% class="fieldset external_id" %)(((
-      (% class="half-width external_id" %)(((
-        #__form_label('external_id' 'section')
-      )))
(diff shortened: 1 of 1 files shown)


sashaandjic:
It would be nice if in the view mode the warning is also displayed below its label.
Edit mode:
screen shot 2017-02-17 at 3 37 21 pm
View mode:
screen shot 2017-02-17 at 3 37 33 pm

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
```diff
@@ -112,20 +112,11 @@
...
-      )))
-      (% class="half-width external_id" %)(((
-      #__form_field('external_id')
-      )))
-      (% class="clear" %)((()))
-    )))
-  #end
+  (% class="fieldset external_id" %)(((
+    #__form_label('external_id' 'section')
+
```
sdumitriu: Are you sure this blank line is needed? In view mode, it causes the layout to be inconsistent, the value is on a separate line.

marta-:
@sdumitriu that's intentional,  see discussion thread on the Jira issue.

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
marta-: @sdumitriu that's intentional, see discussion thread on the Jira issue.

PR #2220, opened 2017-02-18
PT-3075: Reorganize gene panels/matching disorders in OMIM sections

marta-:


sashaandjic:
Tested, but don't like it. It feels quite disconnected. It would help if diagnoses and genes are flipped within the new section, so they are closer to their corresponding sections. I still think that adding something similar to "You may want to investigate" dropdown in Genes and Diagnosis sections would be better solution.
screen shot 2017-02-18 at 12 16 03 pm

PR #2226, opened 2017-02-24
[misc] Disable next button in push dialog if not all mandatory consents checked

veronikaslc:


sashaandjic:
Tested: NEXT button is enabled only when all required consents are checked. 
Problem: I just noticed that push fails in the following scenario. My patient is entered before I introduced mandatory consents, and as a result I can no longer see my patient data. However, I am able to export such patient. Push option is also available, but in the very last step it fails due to the missing consents. We should disable export and push options for patients without mandatory consents. Not sure if this scenario can happen in real life...

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
```diff
@@ -409,6 +409,14 @@
           }).get();
         },
 
+        allReqiredChecked: function() {
+          if (!this.classVariables.consentsBlock) {
+            return true;
+          }
+          var requiredUnchecked = this.classVariables.consentsBlock.find("li[class*='required'] input[type=checkbox]:not(:checked)");
```
sdumitriu: Why `[class*='required']`? Wouldn't `li.required` be what you actually need?

sashaandjic:
Push is failing.

sdumitriu:
Push works for me, but it succeeds even without mandatory consents checked.

sashaandjic:
That's strange. I can select the fields to be pushed, but once I try to push I get this:
screen shot 2017-02-26 at 1 33 07 pm

sashaandjic:
I retested it again with a clean install. Everything works as expected. Two things to consider fixing in the future:
- if I don't have required consent to see the patient data, I should not be able to push or export such record
- scenario: I'm trying to update patient record on a remote server by pushing it for the second time, however the patient record on the remote server no longer belong to me. The error message needs to be more explanatory. Right now it states 
image

sashaandjic:
Not sure if we need to add "(required)" to mandatory consents. This is how it is going to look when pushing on PhenomeCentral: 
screen shot 2017-02-26 at 6 05 33 pm

sdumitriu:
@sashaandjic Can you try pushing to the actual PhenomeCentral (or a 1.3M1 instance), perhaps that version allows pushing even without mandatory consents.

sashaandjic:
I did. Please see the latest screenshot where I pushed the data to the preprod server. I made yet another  clean install and everything looks fine. What puzzles me is that push fails on one out 3 local instances on my machine does. Each instance comes from a different build.

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsent.xml
veronikaslc: it is

sdumitriu:
Re-tested, "Next" is correctly disabled until all mandatory consents are checked.

@veronikaslc Can you address my codereview comments?

veronikaslc:
on it right now

veronikaslc:
@sdumitriu comments addressed

PR #2228, opened 2017-02-26
Family search should return at most the requested number of unique results

marta-:
Plus some variable naming and documentation fixes.

### DIFF
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/FamilySearchResult.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/FamilySearchResult.java
@@ -21,6 +21,8 @@ import org.phenotips.data.Patient;
 import org.phenotips.data.PatientData;
 import org.phenotips.studies.family.Family;
 
+import java.util.Objects;
+
 import org.apache.commons.lang3.StringUtils;
 
 /**
@@ -69,6 +71,21 @@ public class FamilySearchResult
         this(family, requiredPermissions);
         addPatientDescription(patient, usePatientName);
 
+    }
+
+    @Override
+    public boolean equals(Object other)
+    {
+        if (!(other instanceof FamilySearchResult)) {
+            return false;
+        }
+        return (this.id == ((FamilySearchResult) other).getId());
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(this.id);
     }
 
     private void setBasicDescription()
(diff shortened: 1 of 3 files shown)


marta-:
@allasm please review this first thing on Monday. Thanks!

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/FamilySearchResult.java
```diff
@@ -71,6 +73,21 @@ public FamilySearchResult(Patient patient, boolean usePatientName, Family family
 
     }
 
+    @Override
+    public boolean equals(Object other)
```
veronikaslc: I can't find where exactly we use it. If we don't need it, why we have it?

veronikaslc:
I don't see how/where exactly we avoid duplications in this PR or in the code in general.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/PhenotipsFamilyExport.java
```diff
@@ -213,12 +214,12 @@ private void queryFamilies(String input, String requiredPermissions, int results
...
+            results.add(new FamilySearchResult(family, requiredPermissions));
         }
     }
 
     private void queryPatients(String input, String requiredPermissions, int resultsLimit,
-        List<FamilySearchResult> resultsList)
+        Set<FamilySearchResult> results)
     {
         StringBuilder querySb = new StringBuilder();
         querySb.append("from doc.object(PhenoTips.PatientClass) as patient, ");
```
veronikaslc: More of a side question, but why do we need next line?

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/FamilySearchResult.java
marta-: You always need to override equals & hashCode for any class that is used in a Set or as key in a Map.

marta-:
@veronikaslc :
> I don't see how/where exactly we avoid duplications in this PR or in the code in general.
By using a Set instead of a List.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/export/PhenotipsFamilyExport.java
marta-: Right, might be redundant. Doesn't break anything ATM tho.

veronikaslc: We can potentially just search for patients whose families are not included in the results from family search.

marta-: We could, there are many ways of implementing this. This one seemed to me the one easiest to maintain.

PR #2229, opened 2017-02-26
PT-3101: Add support for searching for families in the header search box

marta-:
Configured the FamilySearch service as one of the sources for the search box suggestions.


marta-:
Works best with changes from PR #2228 .

PR #2232, opened 2017-03-01
PT-2545: Reindexing OMIM fails

veronikaslc:


veronikaslc:
@sdumitriu This seems to be working for me. Not entirely sure yet though.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -155,8 +248,8 @@ private SolrQuery addGlobalQueryParameters(SolrQuery query)
     private SolrQuery addFieldQueryParameters(SolrQuery query)
     {
         query.set(DisMaxParams.PF, "name^40 nameSpell^70 synonym^15 synonymSpell^25 text^3 textSpell^5");
-        query.set(DisMaxParams.QF,
-            "name^10 nameSpell^18 nameStub^5 synonym^6 synonymSpell^10 synonymStub^3 text^1 textSpell^2 textStub^0.5");
+        query.set(DisMaxParams.QF, "name^10 short_name^5 included_name^5 nameSpell^18 nameStub^5 "
```
sdumitriu: The `Stub` and `Spell` versions of `short_name` and `included_name` should also be included.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -243,4 +338,183 @@ private int clear()
...
+
+    private void parseOmimData()
+    {
+        try {
+            Reader in =
+                new InputStreamReader(new URL(OMIM_SOURCE_URL).openConnection().getInputStream(),
+                    Charset.forName(ENCODING));
+            for (CSVRecord row : CSVFormat.TDF.parse(in)) {
+                // ignore the comments, moved or removed entries
+                if (row.get(0).startsWith("#") || ("Caret").equals(row.get(0))
```
sdumitriu: Instead of manually checking for `#` each time, you can just use `CSVFormat.TDF.withCommentMarker('#')` to ignore comment lines automatically.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -243,4 +338,183 @@ private int clear()
...
+                    || ("REMOVED FROM DATABASE").equals(row.get(2))) {
```
sdumitriu: Every `REMOVED FROM DATABASE` entry also starts with `Caret`, so this check is not needed.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
sdumitriu: Caret entries shouldn't be ignored, but indexed as is.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
sdumitriu: `addDynamicQueryParameters` should also be updated, to:

- not reference the `keywords` field, which is no longer being populated
- update the `setFilterQueries` call to filter on `type` instead of the name prefix, which no longer works
- update `BQ` to also search in `short_name` and `synonymSpell`

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -243,4 +338,183 @@ private int clear()
...
+
+                this.crtTerm = new SolrInputDocument();
+                // set id
+                setCrtTerm(ID_FIELD, row.get(1));
+
+                // set symbol
+                setCrtTerm(SYMBOL_FIELD, SYMBOLS.get(row.get(0)));
+                // set type (multivalued)
+                for (String type : TYPES.get(row.get(0))) {
+                    setCrtTerm(TYPE_FIELD, type);
```
sdumitriu: `setCrtTerm` sets a value, doesn't append, so only the last set value will actually be taken into account. `crtTerm.addField` should be used instead of `crtTerm.setField`.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -243,4 +338,183 @@ private int clear()
...
+                // set short name
+                String shortNameString = StringUtils.substringAfter(row.get(2), TITLE_SEPARATOR).trim();
+                String[] shortNames = StringUtils.split(shortNameString, TITLE_SEPARATOR);
+                for (String shortName : shortNames) {
+                    setCrtTerm(SHORT_NAME_FIELD, shortName.trim());
+                }
+
+                // set synonyms
+                if (StringUtils.isNotBlank(row.get(3))) {
+                    String[] synonyms = StringUtils.split(row.get(3), SYNONYM_SEPARATOR);
```
sdumitriu: The checkstyle error is easy to fix if these two if-for branches would be replaced by a `setListFieldValue` method.

components/vocabularies/omim/api/src/main/java/org/phenotips/vocabulary/internal/solr/MendelianInheritanceInMan.java
```diff
@@ -209,14 +302,15 @@ public String getVersion()
     public synchronized int reindex(String sourceURL)
     {
         try {
-            Collection<SolrInputDocument> data = new OmimSourceParser(this.hpo, sourceURL).getData();
-            if (data.isEmpty()) {
+            index();
```
sdumitriu: `reindex` is supposed to start with a `clear`, otherwise stale terms will be left in the index.

sdumitriu:
@veronikaslc I'll commit fixes to the comments above shortly.

components/vocabularies/omim/api/src/main/resources/omim/conf/managed-schema.xml
```diff
@@ -143,17 +143,24 @@
   <fields>
     <field name="version" type="string" indexed="true" stored="true" omitNorms="true"/>
     <field name="id" type="string" indexed="true" stored="true" required="true" omitNorms="true"/>
+    <field name="symbol" type="string" indexed="true" stored="true" omitNorms="true"/>
+    <field name="type" type="string" indexed="true" stored="false" omitNorms="true"/>
```
sdumitriu: This should be multivalued.

PR #2236, opened 2017-03-06
Make generic Read More button specific

katpavlova:
Change "Read More" button for browsing Active Groups to "Browse Groups"


marta-:
@katpavlova for future reference,when fixing a specific jira issue (not making a generic cleanup fix), our conventions are as follows:
* the branch name should be the issue id, in this case `PT-3116`
* each commit to the branch should make reference to the issue id, e.g. `PT-3116: Correct Active Groups' "Read More" string` (or, if the commit only fixes part of the issue and there's no subtask for that part, `PT-####: <What the commit does>`
* the pull request title should make reference to the issue id and contain its summary.

All this helps track the development in JIRA, which picks up issue IDs and displays the branches, commits and PRs in the development box on the issue page, allowing us to see what commit(s) fixed each issue. See for example https://phenotips.atlassian.net/browse/PT-3033 .

katpavlova:
@marta- You're right, that's in the onboarding doc, I'm so sorry for forgetting. Thanks for pointing that out. :)

PR #2241, opened 2017-03-10
PT-3061: Annotated ORPHANET disorders with HPO symptoms

a-gvo:


components/vocabularies/hpo/annotation-phenotypes/src/main/java/org/phenotips/vocabulary/annotation/internal/DiseaseNotSymptomAnnotationExtension.java
```diff
@@ -50,10 +55,12 @@
 
     private static final String NOT_SYMPTOM = "not_symptom";
 
+    private static final Collection<String> TARGET_VOCABULARIES = createTargetVocabularyLabels();
```
veronikaslc: can we shorten this to  `private static final Collection<String> TARGET_VOCABULARIES = Collections.singletonList(ORPHANET_LABEL);` ?

components/vocabularies/hpo/annotation-phenotypes/src/main/java/org/phenotips/vocabulary/annotation/internal/DiseaseSymptomAnnotationExtension.java
```diff
@@ -50,10 +55,12 @@
 
     private static final String SYMPTOM = "symptom";
 
+    private static final Collection<String> TARGET_VOCABULARIES = createTargetVocabularyLabels();
```
veronikaslc: same here, if we can

components/vocabularies/hpo/annotation-phenotypes/src/main/java/org/phenotips/vocabulary/annotation/internal/DiseaseNotSymptomAnnotationExtension.java
veronikaslc: or even just return `Collections.singletonList(ORPHANET_LABEL)` in `getTargetVocabularyIds()`?

PR #2245, opened 2017-03-16
PT-2536: Assign patient to the family modal improvement

marta-:
.When "Add to an existing family" button is chosen, move the focus to Family name field.


sashaandjic:
OK. It would be perfect if there is more room above first selection option "Create a new family".

allasm:
I just added two small margins to make it look better on some screens (e.g. mine) and squashed different layout changes into one commit + cleaned commit message (removed leading dot)

PR #2246, opened 2017-03-17
PT-3129 and PT-3132

allasm:
(discussing now if PT-3130 should also be included)

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
@@ -64,11 +64,11 @@ define([
         horizontalPersonSeparationDist: 10,
         horizontalTwinSeparationDist:    8,
         horizontalRelSeparationDist:     6,
-        yDistanceNodeToChildhub:        20,
+        yDistanceNodeToChildhub:      21.6,
         yDistanceChildhubToNode:        14,
         yExtraPerHorizontalLine:         4,
         yAttachPortHeight:             1.5,
-        yCommentLineHeight:            2.9,
+        yCommentLineHeight:            2.8,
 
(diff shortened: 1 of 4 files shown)


components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
```diff
@@ -64,11 +64,11 @@ define([
         horizontalPersonSeparationDist: 10,
         horizontalTwinSeparationDist:    8,
         horizontalRelSeparationDist:     6,
-        yDistanceNodeToChildhub:        20,
+        yDistanceNodeToChildhub:      21.6,
```
marta-: so many magic numbers...

sashaandjic:
PT-3132 ok
before
after

PT-2129 I don't see any difference

marta-:
@sashaandjic it's a typo. @allasm meant PT-3129, not PT-2129.

components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
allasm: those magic numbers set the spacing between nodes in pedigree, for best visual results, so they are kind of magic. We have size of node graphic in pixels defined elsewhere, but that is independent, since this number sets the spacing between nodes regardless of how big each node is. I guess there can be a more uniform way to define things, but with any system various aspects of pedigree spacing should be set somewhere using some numbers humans visually like, so these are it. And since internal pedigree has many parts, spacing between all those parts should be set. 

P.S. The one magic number which is not independent is `yCommentLineHeight` which depends on comment font size. It was added last, and in theory worth being computed based on the font size somehow (plus line spacing etc - so at the end of the day one magic number seems easier to tweak, when tweaking)

PR #2250, opened 2017-03-20
PT-2703: Increase checkbox opacity for inactive administration select…

marta-:
…ions


components/vocabularies/hpo/ui/src/main/resources/PhenoTips/FormDesigner.xml
```diff
@@ -1099,7 +1099,10 @@ document.observe("xwiki:dom:loaded", function(){
 }
 
 .designer-panel.sections .disabled {
-  opacity: .3;
+  opacity: .8;
```
sdumitriu: I find `.6` better, WDYT?

marta-: Don't know what to say. The initial complaint was that the whole thing looks disabled and inactive, so we're trying to avoid that. Shall we compromise? .7? :)

PR #2252, opened 2017-03-20
PT-3143: Advanced users are unable to run option Export XAR on patien…

marta-:
…t record form

Quick fix: don't show the Export XAR option to non-admins


components/skin/ui/src/main/resources/PhenoTips/ContentTopMenu.xml
```diff
@@ -891,7 +891,9 @@ $xwiki.ssx.use('PhenoTips.JSONImportDialog')##
 ** (% class="empty" %)----
 ** [[$services.localization.render('phenotips.contentTopMenu.moreActionsMenu.exportPDF')&gt;&gt;path:$doc.getURL('export', "format=pdf&amp;pdfcover=0&amp;pdftoc=0&amp;pdftemplate=${pdfTemplate}")|| class="export-link"]]
 #if($isAdvancedUser)
+#if ($hasAdmin)
```
sdumitriu: I wonder if the user has to be both an admin and an advanced user to get this entry in the menu. In other words, should the `if` above be moved after this `if`?

PR #2255, opened 2017-03-20
PT-3142, PT-3145, PT-3147 - minor pedigree bugs

allasm:


allasm:
@sashaandjic  - please test, even if Sergiu already merges this into master

PT-3142 is easy and the issue explains itself

PT-3145 has a sample JSON which could not beim ported in 1.3rc1 (but can be imported after this fix)

PT-3147 is tricky, I can not find a user-visible problem caused by the bug being fixed, but there was a bug as an undefined variable was referred and a (useful :) chunk of code was skipped due to exception - in _somce_ cases. So to test need to create various pedigrees and remove various bits and pieces of them, and make sure they look OK and - more importantly - behave ok after removal, e.g. create a complex pedigree with lines crossing other lines, and multi-generational edges, remove some noded, and keep adding more and make sure everything is ok

sashaandjic:
i'm on it

sashaandjic:
no problems found

PR #2262, opened 2017-03-24
PT-3167: Correct unhelpful update failure message

katpavlova:
"Unable to perform the update" was redundant as an error message, since it would always be paired with the "Failed to update" string in the UI. Since phenotips.patientAccessRightsManagement.updateError is the error used when an unknown update error occurs, the string was updated to reflect that for easier error tracking.


sdumitriu:
Please use the affected issue number in the commit message.

Commit manually amended and pushed into master.

PR #2264, opened 2017-03-27
PT-3168: Remote server name missing in error mesage

marta-:


sashaandjic:
tested. OK.
screen shot 2017-03-27 at 3 21 39 pm

PR #2266, opened 2017-03-28
PT-1366: The vocabulary reindexing operation should not leave the system in a bad state at any moment

veronikaslc:


veronikaslc:
the only problem is to delete one file in temp directory `java.io.IOException: Unable to delete file: ...\PhenoTips\solr\hpo_temp\data\spellchecker\_0.cfs`
 
windows displays `File In Use` error `... the file is open in Java(TM) Platform SE binary`
  
other than that reindexing should work

allasm:
what happens if that `File IN Use` error happens? A junk file is left or re-indexing still fails?

veronikaslc:
Nothing happens, directory is cleaned after successful reindexing, so just one file is left.

sashaandjic:
tested

veronikaslc:
rebased

veronikaslc:
@sdumitriu if `reindex()` is identical for all three, can I move it to `AbstractSolrVocabulary`?

PR #2268, opened 2017-03-28
[misc] Minor ui improvement for smaller resolutions

marta-:
Allows this element not to slide on the next line because of the larger permissions element.

### DIFF
--- a/components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
+++ b/components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
@@ -551,7 +551,7 @@
   vertical-align: middle;
 }
 .study-assignment-config {
-  width: 48%;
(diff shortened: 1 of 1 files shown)


components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
```diff
@@ -551,7 +551,7 @@
   vertical-align: middle;
 }
 .study-assignment-config {
-  width: 48%;
+  max-width: 48%;
```
sdumitriu: I think this can be removed completely. Initially permissions had half and studies had half, so that anything coming after them would be forced to be on the next line, but it's simpler to just use `clear:both` on the following item.

marta-: k.

PR #2270, opened 2017-03-28
PT-3162: Empty remote server entries should be ignored in the 'Push to' menu

veronikaslc:


components/patient-data-sharing/push-api/src/main/java/org/phenotips/data/push/internal/DefaultPushPatientService.java
```diff
@@ -156,9 +157,14 @@ private XWikiContext getXContext()
...
                     serverConfiguration.getStringValue(DefaultPushPatientData.PUSH_SERVER_CONFIG_ID_PROPERTY_NAME));
-                PushServerInfo info = new DefaultPushServerInfo(
-                    serverConfiguration.getStringValue(DefaultPushPatientData.PUSH_SERVER_CONFIG_ID_PROPERTY_NAME),
-                    serverConfiguration.getStringValue(DefaultPushPatientData.PUSH_SERVER_CONFIG_URL_PROPERTY_NAME),
+                String name =
+                    serverConfiguration.getStringValue(DefaultPushPatientData.PUSH_SERVER_CONFIG_ID_PROPERTY_NAME);
+                String url =
+                    serverConfiguration.getStringValue(DefaultPushPatientData.PUSH_SERVER_CONFIG_URL_PROPERTY_NAME);
+                if (StringUtils.isEmpty(name) || StringUtils.isEmpty(url)) {
+                    continue;
```
marta-: I would suggest using `isBlank` instead.

PR #2271, opened 2017-03-29
PT-3083: Add link to request PC account on push to PC screen (if user has no PC account)

veronikaslc:


pom.xml
```diff
@@ -547,6 +547,12 @@
               <method>boolean isDisabled()</method>
               <justification>New methods in support of disabled visibilities.</justification>
             </difference>
+            <difference>
+              <className>org/phenotips/data/push/PushServerInfo</className>
+              <differenceType>7012</differenceType>
+              <method>java.lang.String getServerRegistrationURL()</method>
+              <justification>New methods to erxpose push server registration URL.</justification>
```
marta-: typo

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -773,6 +775,8 @@
         var groupSelection = this._remoteGroupLabel.replace("__serverID__", this._selectedServer);
         $$('#group-selection h2')[0].innerHTML = groupSelection;
 
+        if (serverData.registrationURL != "") {$('user-selection').insert(new Element('p', {'class' : 'intro xHint'}).insert("$!escapetool.javascript($services.localization.render('phenotips.PushPatient.remoteRegistration'))".replace("__serverID__", this._selectedServer).replace("__registrationUrl__", "&lt;a target=\"_blank\" href=\""+serverData.registrationURL+"\"&gt;"+serverData.registrationURL+"&lt;/a&gt;")));}
```
marta-: This appears to keep adding new elements for each server you attempt to push to.

To reproduce the issue:
- add 2 more push targets, one with a registration url and one without
- create a patient
- click on Push to PhenomeCentral: all looks fine
- close the dialog and click on the push target WITHOUT registration URL: the message for signing up to PhenomeCentral is still displayed
- close the dialog and click on the other push target WITH registration URL: both the message for signing up to PhenomeCentral and the message for signing up to this push target are displayed

marta-: Additionally, the line about signing up should disappear once the user is logged in.

veronikaslc:
fixed, rebased on master

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -500,6 +485,7 @@
           if (this._lastLoginResponse.status == "success") {
             console.log("--&gt; login successful!");
             $$('#user-selection p')[0].hide();
+            $$('#user-selection p')[1].hide();
```
sdumitriu: `[0]` and `[1]` looks very brittle. You should add some semantic classes.

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -769,10 +754,17 @@
...
         $$('#user-selection h2')[0].innerHTML = userSelection;
         var userSelectionHint = this._authenticationHint.replace(/__serverID__/g, this._selectedServer);
-        $$('#user-selection p')[0].innerHTML = userSelectionHint;
+        $('user-selection-intro').innerHTML = userSelectionHint;
+        $('user-selection-intro').show();
         var groupSelection = this._remoteGroupLabel.replace("__serverID__", this._selectedServer);
         $$('#group-selection h2')[0].innerHTML = groupSelection;
-
+        if (serverData.registrationURL != "" &amp;&amp; !this._lastApprovedUser) {
+           var remoteRegistration = this._remoteRegistrationHint.replace("__serverID__", this._selectedServer).replace("__registrationUrl__", "&lt;a target=\"_blank\" href=\""+serverData.registrationURL+"\"&gt;"+serverData.registrationURL+"&lt;/a&gt;");
```
marta-: The remoteURL is being altered a few lines above in the `_populateServerList` function, to prepend `http://` (using the `_addHTTP`) function. I would do something similar for the registration url (if the admin entrers one url without its protocol, they're very likely to do the same for the other).

However, I would account for the fact that in some cases the signing up can be done by contacting an email address ,
 since not all PhenoTips installations that receive contributions will have registration open, in which case the prefix should be `mailto:` instead.

Moreover, imo this cleaning up of the url should be done serverside in the service, rather than in javascript. 

@sdumitriu @veronikaslc thoughts?

PR #2272, opened 2017-03-30
PT-3179: Sending a JSON to a non-existing patient using the /patients/eid/ REST endpoint should create a new patient record with that external identifier

a-gvo:


components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java
```diff
@@ -201,4 +211,60 @@ private Response checkForMultipleRecords(Patient patient, String eid)
...
+        try {
+            if (StringUtils.isBlank(json)) {
+                this.logger.error("Provided patient json: {} is invalid.", json);
+                return Response.status(Status.BAD_REQUEST).build();
+            }
+            final JSONObject patientJson = new JSONObject(json);
+            final User currentUser = this.users.getCurrentUser();
+            if (!this.access.hasAccess(Right.EDIT, currentUser == null ? null : currentUser.getProfileDocument(),
+                this.currentResolver.resolve(Patient.DEFAULT_DATA_SPACE, EntityType.SPACE))) {
+                this.logger.error("Edit access denied to user [{}].", this.users.getCurrentUser());
```
sdumitriu: You already have `currentUser`, no need to invoke the manager again.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java
```diff
@@ -201,4 +211,60 @@ private Response checkForMultipleRecords(Patient patient, String eid)
...
+                return Response.status(Status.FORBIDDEN).build();
+            }
+            final Patient patient = this.repository.create();
+            // Since we're creating a new patient, if eid is not provided in patientJson, use the eid that was passed in
+            if (StringUtils.isBlank(patientJson.optString(EID_LABEL))) {
```
sdumitriu: For an existing patient, sending a blank eid will actually blank it. I think we should allow the same here. Replace with `json.has`.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImplTest.java
```diff
@@ -148,16 +156,21 @@ public void setUp() throws ComponentLookupException, URISyntaxException
         Provider<XWikiContext> provider = this.mocker.getInstance(XWikiContext.TYPE_PROVIDER);
         this.context = provider.get();
         ReflectionUtils.setFieldValue(this.component, "uriInfo", this.uriInfo);
+        ReflectionUtils.setFieldValue(this.component, "currentResolver", this.resolver);
```
sdumitriu: This isn't needed, since it's an injected dependency, you can get the automocked instance with `mocker.getInstance`.

a-gvo: Fixed.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java
a-gvo: Fixed.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java
a-gvo: Fixed.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByExternalIdResourceImpl.java
```diff
@@ -201,4 +211,60 @@ private Response checkForMultipleRecords(Patient patient, String eid)
...
+                this.currentResolver.resolve(Patient.DEFAULT_DATA_SPACE, EntityType.SPACE))) {
+                this.logger.error("Edit access denied to user [{}].", currentUser);
+                return Response.status(Status.FORBIDDEN).build();
+            }
+            final Patient patient = this.repository.create();
+            // Since we're creating a new patient, if eid is not provided in patientJson, use the eid that was passed in
+            if (!patientJson.has(EID_LABEL)) {
+                patientJson.put(EID_LABEL, eid);
+            }
+            patient.updateFromJSON(patientJson);
```
sdumitriu: One thing different is that an internal ID conflict doesn't throw exception here.

PR #2274, opened 2017-04-07
[misc] Fixing sorting phenotypes by categories in helper class

veronikaslc:
A fix for sorting phenotype list by predefined categories obtained from configuration. Currently not used in codebase.

### DIFF
--- a/components/export/api/pom.xml
+++ b/components/export/api/pom.xml
@@ -29,7 +29,7 @@
   <name>PhenoTips - Data export - Spreadsheet Export APIs</name>
 
   <properties>
-    <coverage.instructionRatio>0.18</coverage.instructionRatio>
+    <coverage.instructionRatio>0.19</coverage.instructionRatio>
     <checkstyle.suppressions.location>${basedir}/src/checkstyle/checkstyle-suppressions.xml</checkstyle.suppressions.location>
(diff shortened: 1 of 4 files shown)


marta-:
@veronikaslc you should always explain what problem you are actually fixing. Please give example of input, what was the expected result, what was the actual result.

veronikaslc:
@marta- sure, my bad.
This PR fixes an exception in `featureSetUp()`  method  that reads the phenotype categories configuration ("phenotypeMapping") from `ScriptService` class. Method gets a list of all categories, and maps each category's title to a list of HPO ids which represent it. The map is used for sorting any list of phenotypes by categories.

veronikaslc:
rebased

veronikaslc:
rebased

components/export/api/pom.xml
```diff
@@ -111,6 +111,11 @@
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>annotations</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
```
sdumitriu: This is not used.

PR #2275, opened 2017-04-07
PT-3083: Add link to request PC account on push to PC screen (if user has no PC account)

veronikaslc:
Typo fix


sashaandjic:
tested
screen shot 2017-04-07 at 11 40 44 pm

PR #2277, opened 2017-04-07
PT-3185: Consolidate family-studies translations into one file

katpavlova:
Consolidate existing family-studies translations (into ApplicationResources.properties files) and delete all versions of FamilyTranslations.xml


katpavlova:
Translation conflicts have been resolved and re-consolidated, should be safe to overwrite the two conflicting files in master.

PR #2280, opened 2017-04-11
PT-3192: Phenotypes no longer fail to display for patients with non-s…

a-gvo:
…tandard phenotypes in non-standard categories.


sashaandjic:
tested

PR #2281, opened 2017-04-12
PT-3193: Fixed bug where MATCHING DISORDERS IN OMIM were not displaye…

a-gvo:
…d when browser back button was clicked.


sashaandjic:
tested

PR #2282, opened 2017-04-12
PT-2934 and PT-3181

allasm:


allasm:
@sashaandjic please test (a complete test also makes sure regular operations are not hurt, e.g. loading a patient with genes for PT-3181 or entering stuff in various text fields for PT-2934 - of course I tried it as well, but just in case)

sashaandjic:
tested. both defects resolved.

components/pedigree/resources/src/main/resources/pedigree/view/person.js
```diff
@@ -1579,6 +1579,8 @@ define([
                     //          {id: 'ENSG00000160688', gene: 'FLAD1', status: 'solved'},
                     //          {id: 'custom', 'gene': 'custom', status: 'candidate'} ]
                     this._setGenes(info.genes);
+                } else {
+                    this._setGenes([]);
```
marta-: so much duplicated code in `assignProperties` ....

veronikaslc: on the other hand it's so clean & easy to find a spot that needs a change

marta-: However, more generic code (that didn't just go through each and every one of many fields to do the same thing, but somehow triggered the update on all) would probably not have had the bug in the first place.

allasm: agree, probably it would be better. Just as many things this started form 3 or 4 fields which were different enough that coming up with a shared code was more pain than do it this way, and then it grew and has multiple similar entries. We should probably make a PR to refactor data/data field handling in pedigree, but not for this PR/not for 1.3

veronikaslc: I can think of something like making an array map with properties and iterate over an array with generic function:

`var properties = {
        "adoptedStatus"   : {
           default   : "",
           get : this.getAdopted(),
           set : setAdopted()      
        }, ...`

and then call `for (var property in properties) {
                           if(info.hasOwnProperty(property)) {
                               if (this.get(properties[property]) != info[property]) {
                                   this.set(info[property]);
                               }
                           } else {
                               this.set(properties[property].default);
                           }`

marta-: @allasm yes, obviously not for 1.3.

marta-: @veronikaslc we shouldn't have to specify the getters/setters in the map, they should all be standard like in Java: `get<CapitalizedFieldName>`. Also, since some setters trigger a possibly costly ui update, the check whether the new value is different from the old one makes more sense in the setter imo, so that all other code can just call the setter without worrying. Just my 2cents.

PR #2287, opened 2017-04-24
[misc][ux] Display cases before members on Study pages

marta-:


sdumitriu:
Should the same thing be done for workgroups?

marta-:
@sdumitriu I'd say no. For groups, members are more important and contributons are somewhat secondary.

PR #2290, opened 2017-04-25
PT-3199: 'Present' column in exported excel is blank for some cases (see description in the comment)

veronikaslc:
'Present' column in exported excel is blank for patients who have only negative phenotypes, only negative prenatal phenotypes, only positive phenotypes, only positive prenatal phenotypes, only positive phenotypes and positive prenatal phenotypes, only negative phenotypes and negative prenatal phenotypes

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -129,7 +129,7 @@ public class DataToCellConverter
 
         DataSection section = new DataSection();
         int hX = 0;
-        if (present.contains("positive") && present.contains("negative")) {
+        if (present.contains("positive") || present.contains("negative")) {
             DataCell cell =
                 new DataCell(this.translationManager.translate("phenotips.export.excel.label.phenotype.present"),
                     hX, 1, StyleOption.HEADER);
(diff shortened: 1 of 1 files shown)


components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -179,27 +179,26 @@ public DataSection phenotypeBody(Patient patient) throws Exception
...
-                lastStatus = feature.isPresent();
+            if (present.contains("positive") || present.contains("negative")) {
                 lastSection = "";
-                DataCell cell = new DataCell(lastStatus
+                DataCell cell = new DataCell(feature.isPresent()
                     ? this.translationManager.translate("yes")
                     : this.translationManager.translate("no"),
                     x, y);
-                if (!lastStatus) {
+                if (bothTypes && lastStatus != null && lastStatus != feature.isPresent()) {
```
allasm: you can get rid of the `bothTypes` variable completely, since if there are no "both types" the check `lastStatus != null && lastStatus != feature.isPresent()` will never be true. Same in the other method below

veronikaslc: I left it there intentionally not wiast time on more checks if not `bothTypes`

allasm: well, it does waste codereviwer time ;), but also evaluating it and storing also takes time, so IMHO even if we cared about nanoseconds it would be about the same, but with less code (which is always good). Not critical in any case

allasm: ...for that matter, `present.contains("positive")` is the slow part, not `feature.isPresent()` - so caching that and reusing would improve performance much more

veronikaslc:
fixed by https://github.com/phenotips/phenotips/commit/b7df69f77e0e3c086325eb1ac2473ee44e9b7180

sdumitriu:
Since the changes in #2312 touched the same code, the fix was ported into that pull request.

PR #2293, opened 2017-04-28
PT-1403: Replace all usages of the deprecated ScriptServices with the new ontology access services

veronikaslc:
Deprecated services: OmimService, OmimInfoService, OmimPredictService, GeneInfoService, GeneNameService, PhenotypeInfoService


veronikaslc:
addressed codereview comments

veronikaslc:
I am going to rebase this branch on top of latest master since alot has changed lately.

veronikaslc:
codereview comments are addressed

veronikaslc:
strange.. was fixing exactly those bugs last night, my running instance works well, let me double-check if I pushed all changes

veronikaslc:
I have just recompiled the branch. There is a strange thing. The source has all my latest changes, the all got pushed to git. The build of that source code somehow has old code. If I import `phenotips-widgets-ui.xar` separately (from the build that I just did) - it updates UI with my latest changes and everything works. The question is - why build doesn't have the updated code?

allasm:
I just built the branch and ran a local instance. All the code (as far as I can tell) which is supposed to be there was there, and all the bugs @sdumitriu mentioned above don't seem to be there

veronikaslc:

components/vocabularies/rest/src/main/java/org/phenotips/vocabularies/rest/internal/DefaultVocabularyTermSuggestionsResource.java
```diff
@@ -79,18 +75,17 @@ public VocabularyTerms suggest(String vocabularyId, String input, @DefaultValue(
...
+        JSONObject rep = new JSONObject();
+        JSONArray trms = new JSONArray();
         for (VocabularyTerm term : termSuggestions) {
-            VocabularyTermSummary termRep = this.objectFactory.createVocabularyTermRepresentation(term);
-            termRep.withLinks(this.autolinker.get().forSecondaryResource(VocabularyTermResource.class, this.uriInfo)
-                .withActionableResources(VocabularyResource.class)
-                .withExtraParameters("vocabulary-id", vocabularyId)
-                .withExtraParameters("term-id", term.getId())
+            JSONObject trm = term.toJSON();
+            trm.put("links", this.autolinker.get().forSecondaryResource(VocabularyTermResource.class, this.uriInfo)
```
sdumitriu: This doesn't work, but I'm not sure making it work is required.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
```diff
@@ -151,12 +120,6 @@ document.observe('xwiki:dom:loading', function() {
                 item._restriction = null;
               }
             }
-            if (item._customOptions.queryProcessor && item._restriction) {
-              item._customOptions.queryProcessor = Object.clone(item._customOptions.queryProcessor);
-              item._customOptions.queryProcessor.restriction = {
-                'term_category' : item._restriction
```
sdumitriu: Regression: searching in only one subsection of HPO no longer works, results from the entire HPO are returned.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -145,7 +145,7 @@
     matchVarname : "id",
     // The CSS classname of the suggest list
     className : "ajaxsuggest",
-    timeout : 2500,
+    timeout : 3000,
```
sdumitriu: The value used everywhere else had an extra 0, for 30 seconds.

components/patient-data/ui/src/main/resources/PhenoTips/AllergiesDataClass.xml
```diff
@@ -63,29 +63,12 @@
...
-1. #_fms__displayValue($v ${foreach.count})
+      <customDisplay>{{include reference="PhenoTips.PatientSheetMacros" /}}
 
-#end
-#if ((!$value || $value.size() == 0) &amp;&amp; ($xcontext.action == 'edit' || $xcontext.action == 'inline'))
-1. #_fms__displayValue('' 1)
-#end
-{{/velocity}}</customDisplay>
+{{velocity filter="indent"}}{{html wiki=true clean=false}}
+#set ($suggest_script = $xwiki.getURL('PhenoTips.AllergySearch', 'get', 'outputSyntax=plain'))
```
sdumitriu: Regression: previously entered allergies are no longer suggested. This is because `AllergySearch` returns XML, and the suggest widget was changed to expect JSON by default. It should be simple enough to change `AllergySearch` to return JSON instead.

Actually there's another reason: the URL ends up as `...AllergySearch?outputSyntax=plaininput=asd`, missing an `&` before `input=`.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetMacros.xml
```diff
@@ -238,6 +238,31 @@
...
+      &lt;input class="xwiki-free-multiselect-value" type="text" size="30" name="${prefix}${name}" id="${prefix}${name}_${counter}" value="$!escapetool.xml($v)"/&gt;
+    {{/html}}
+  #else
+    $!{services.rendering.escape($v, $doc.syntax)}
+  #end
+#end
+(% class="xwiki-free-multiselect"%)
+#foreach ($v in $value)
+1. #_fms__displayValue($v ${foreach.count})
+
```
sdumitriu: Remove this newline, lists are numbered `1, 1, 1...` instead of `1, 2, 3...` because of it.

veronikaslc:

sdumitriu:
@veronikaslc Restricting search in only one subsection still doesn't work. Adding back the query processor code isn't enough, the code that actually restricts to the right categories needs to be there as well.

PR #2296, opened 2017-05-04
PT-3086: Added option to filter out genes that have been ruled out in gene panels

a-gvo:


components/gene-panels/ui/src/main/resources/PhenoTips/PanelsDisplay.xml
```diff
@@ -484,6 +504,31 @@
...
+#genePanelsDownloadWrapper {
+  float: left;
+}
+
+#genePanelsExcludeGenesWrapper {
+  float: right;
+  font-size: .7em;
+  text-transform: capitalize;
+  margin-right: -20px;
+  color: #4D5860;
```
marta-: We're trying to avoid hardcoded colors. Please use one from the active color theme, e.g. `$theme.textColor` instead (available thanks to the inclusion of the  `colorThemeInit.vm` template at the beginning of tis CSS block).

marta-:
This seems to work great. A couple of minor comments:
* the checkbox and its label are a bit misaligned (at least for me, see attached screenshot), so the alignment should be improved
* could we make the checkbox blend in a bit better by making it look more like the show/hide buttons? What do others think? @sdumitriu @sashaandjic @veronikaslc 

image

sashaandjic:
Like this?
screen shot 2017-05-04 at 3 19 58 pm

marta-:
@sashaandjic that's not quite what i had in mind. While your suggestion is perfectly consistent with the other buttons, I think it may be confusing because we're using identical visuals for two very different things: showing/hiding a whole form section versus filtering a list of results.

What I had discussed with Alina looks more like this (up for further discussion):

image

image

However, we decided to wait a bit before making the change, for several reasons (the exclusion of carrier or confirmed genes may also make sense so we need to factor that in, not everyone is happy with the currently chosen position for the checkbox, etc).

sashaandjic:
tested

components/gene-panels/api/src/main/java/org/phenotips/panels/internal/TermsForGeneBuilder.java
```diff
@@ -57,11 +107,10 @@
      * @param geneSymbol the gene symbol, which will be used as the key
      * @param term the {@link VocabularyTerm vocabulary term} with which to update the stored {@link TermsForGene}
      * @throws NullPointerException if key {@code geneSymbol} has not yet been added to the {@link TermsForGeneBuilder},
-     *             or if {@code term} or {@code geneSymbol} are null
+     *             or if {@code term} is null
      */
-    void update(@Nonnull final String geneSymbol, @Nonnull final VocabularyTerm term)
+    void update(@Nullable final String geneSymbol, @Nonnull final VocabularyTerm term)
```
sdumitriu: Are you sure? Further down `add` still requires it to be nonnull.

components/gene-panels/api/src/main/java/org/phenotips/panels/GenePanelFactory.java
```diff
@@ -39,29 +38,49 @@
...
-        @Nonnull Collection<VocabularyTerm> absentTerms);
+    GenePanel build(
+        @Nonnull Collection<VocabularyTerm> presentTerms,
+        @Nonnull Collection<VocabularyTerm> absentTerms,
+        @Nonnull Collection<VocabularyTerm> rejectedGenes);
 
     /**
-     * Create an object of {@link GenePanel} class for a given {@link Patient} object.
+     * Create an object of {@link GenePanel} class for a given {@link Patient} object. Any genes marked as rejected
+     * in the {@code patient} will be included in the panel data.
```
a-gvo: Fixed. Should I implement excluding rejected genes?

components/gene-panels/api/src/main/java/org/phenotips/panels/internal/TermsForGeneBuilder.java
a-gvo: Right, fixed.

PR #2298, opened 2017-05-05
Pt 3214 patient data tests

AlexLapena:


sdumitriu:
Did you build without `-Pquick`?

PR #2300, opened 2017-05-05
PT-3211: Move a DiffDiagnosisService into separate component

veronikaslc:


components/phenotype-suggest-service/api/src/main/java/org/phenotips/solr/PhenotypeSuggestService.java
```diff
@@ -54,18 +57,19 @@
 @Component
 @Named("omim")
```
sdumitriu: Wrong name.

components/phenotype-suggest-service/api/src/main/java/org/phenotips/solr/PhenotypeSuggestService.java
```diff
@@ -54,18 +57,19 @@
 @Component
```
sdumitriu: Every component must implement a role, but this doesn't implement anything, so it will fail to be instantiated.

components/phenotype-suggest-service/api/src/main/java/org/phenotips/solr/PhenotypeSuggestService.java
```diff
@@ -148,20 +151,56 @@ protected String getName()
...
 
-        return new MapSolrParams(params);
+    /**
+     * Get the HPO IDs of the specified phenotype and all its ancestors.
+     *
+     * @param id the HPO identifier to search for, in the {@code HP:1234567} format
+     * @return the full set of ancestors-or-self IDs, or an empty set if the requested ID was not found in the index
+     */
+    @SuppressWarnings("unchecked")
+    public Set<String> getAllAncestorsAndSelfIDs(final String id)
```
sdumitriu: A vocabulary term already has a `term.getAllAncestorsAndSelf` method. Instead of duplicating code, you should just call that method, then collect the IDs of each ancestor term.

components/phenotype-suggest-service/api/src/main/java/org/phenotips/solr/PhenotypeSuggestService.java
```diff
@@ -19,32 +19,35 @@
```
sdumitriu: The package is wrong, this isn't about `solr`, but about phenotypes. @marta- what's a good name for the package? `org.phenotips.diagnosis.differentialPhenotypes` perhaps?

components/solr-access-service/api/src/main/resources/ApplicationResources.properties
```diff
@@ -1,2 +0,0 @@
-phenotips.solrAccessService.noMatches=No matches found
```
sdumitriu: These translations are still used, perhaps they should be removed in #2293 instead.

veronikaslc:
codereview comments are addressed, branch is rebased on top of latest changes in master and PT-1403

PR #2301, opened 2017-05-08
PT-3187: Rename Disorders column in excel export, create separate header for comments

katpavlova:
Diagnosis comments/notes will still only export if the Final Diagnosis field is selected to be exported. I did not create an option to export them separately because the export menu will be redone soon, but I did place them in the render list for the export menu so that they are given their own export option in the future.

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -233,8 +233,7 @@ public class DataToCellConverter
                 { "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp", "transcript", "protein",
                     "cdna" },
                 { "segregation", "evidence", "inheritance", "interpretation", "effect", "zygosity", "dbsnp",
-                    "transcript",
-                    "protein", "cdna" },
+                    "transcript", "protein", "cdna" },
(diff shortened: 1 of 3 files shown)


sashaandjic:
I'm also not sure about Notes becoming Comments. Users may find this confusing.

sashaandjic:
On a second thought "Notes" heading was equally too. In the Patient form it is called "Additional comments". So "Diagnoses Comments" are not that bad, as it gives user perspective as where the comments(notest) are coming from. I'd keep it this way.

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -190,7 +190,8 @@
        'variants_evidence' : "$services.localization.render('phenotips.exportPreferences.field.variantsEvidence')",
        'variants_segregation' : "$services.localization.render('phenotips.exportPreferences.field.variantsSegregation')",
        'variants_sanger' : "$services.localization.render('phenotips.exportPreferences.field.variantsSanger')",
-       'variants_coordinates' : "$services.localization.render('phenotips.exportPreferences.field.variantsCoordinates')"
+       'variants_coordinates' : "$services.localization.render('phenotips.exportPreferences.field.variantsCoordinates')",
+       'diagnosis_notes' : "$services.localization.render('phenotips.exportPreferences.field.diagnosisNotes')"
```
sdumitriu: `diagnosis_notes` is also included in the autogenerated list, so now it appears twice, with different labels.

PR #2304, opened 2017-05-11
PT-3107: Switched to standard XWiki pagination look for gene panels

a-gvo:


components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene_suggestions.xml
```diff
@@ -121,16 +121,67 @@
...
+    === $escapetool.xml($services.localization.render('phenotips.UIXField.geneSuggestions.geneSearch.matchingGenes'))  {{html wiki=false clean=false}}&lt;span class="buttonwrapper"&gt;&lt;a href="#" title="$services.localization.render('phenotips.UIXField.geneSuggestions.download.hint')" id="genePanelsExport" class="button gene-panels-button csv-export-button"&gt;&lt;span class="fa fa-fw fa-download"&gt;&lt;/span&gt; $services.localization.render('phenotips.UIXField.geneSuggestions.download')&lt;/a&gt;&lt;/span&gt;{{/html}}{{icon name="question-circle" title="$services.localization.render('phenotips.UIXField.geneSuggestions.download.hint')" cssClass="xHelpButton"/}} ===
```
sdumitriu: `$escapetool.xml` only makes sense when inside a `{{html}}` macro, otherwise it's not needed. You should remove it from here and two lines down.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene_suggestions.xml
```diff
@@ -121,16 +121,67 @@
...
-      (% class="hint"%)$services.localization.render('phenotips.UIXField.geneSuggestions.geneSearch.Hint')(%%)##
+      (% class="hint"%)$escapetool.xml($services.localization.render('phenotips.UIXField.geneSuggestions.geneSearch.Hint'))(%%)##
       {{html wiki=false clean=false}}&lt;input type="hidden" id="panelsSearchTermsInput" value="" /&gt;&lt;input type="hidden" id="panelsSearchTermIDsInput" value="" /&gt;{{/html}}##
       {{html wiki=false clean=false}}&lt;input type="hidden" id="panels-phenotypes-summary" value="$escapetool.xml($jsontool.serialize($phenotypesSummary))" /&gt;{{/html}}##
       (% id="panelsSearchTerms" %)((()))
       (% class="background-search gene-search hidden" %)(((
-        (% id="gene-search-results" class="background-suggestions" %)((()))
+        (% id="gene-search-results" class="background-suggestions" %)(((
+          {{html wiki=false clean=false}}
+            &lt;table id="panels-livetable" class="xwiki-livetable"&gt;
```
sdumitriu: What prevents you from using the `#livetable` macro which normally generates all this HTML?

components/gene-panels/rest/src/main/java/org/phenotips/panels/rest/GenePanelsLiveTableResource.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param absentTerms a list of term IDs that are observed to be absent
+     * @param offset the offset for the results, numbering starts from 1
+     * @param limit the number of results to display, must be an integer
+     * @param reqNo the request number, must be an integer
+     * @return associated genes and counts data if successful, an error code otherwise
+     */
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    Response getGeneCountsFromPhenotypes(
+        @QueryParam("present-term") List<String> presentTerms,
```
sdumitriu: I don't like that the parameters are sent in they query string. Ideally:

- this should use the request directly to get the values, like the `DefaultGenePanelsResourceImpl`, which accepts both GET and POST parameters
- the livetable should send data as POST body, not in the URL.

Is that doable?

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene_suggestions.xml
a-gvo: The macro creates a new LiveTable js object. I'm using my own implementation (I added a bug fix and a custom method).

components/gene-panels/rest/src/main/java/org/phenotips/panels/rest/GenePanelsLiveTableResource.java
a-gvo: The current livetable implementation sends data in the query string, but i can override the method in my implementation.

a-gvo: Fixed.

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__gene_suggestions.xml
a-gvo: Fixed.

a-gvo:
Addressed code review comments and rebased onto the latest master.

PR #2305, opened 2017-05-11
PT-2925 and PT-2860

allasm:
@veronikaslc please review

@sashaandjic please test.

Since there are changes in all areas related to patient data, a few different things should be tested:
1) PT-2860 as described in the issue
2) pushing in general (smoke test)
3) setting patient access rights, via UI and via REST (if we already have REST for that?)
4) patient data: setting via UI, getting/setting via REST (smoke test)
5) smoke test family studies (create family, edit pedigree, modify data/save, add patients, remove/unlink patient, delete family, list families)
6) JSON and Excel export, JSON import

This is an updated version of https://github.com/phenotips/phenotips/pull/2148 rebased on top of master right after 1.3-final (so the version is 1.4). Due to the size of the chnages/rebase I created a new pull request so that changes can be compared to the old one, if need be

### DIFF
--- a/components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
+++ b/components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
@@ -55,8 +55,7 @@ public interface PrimaryEntity
      * Returns a reference to the document where the entity is stored.
      *
      * @return a valid document reference
-     *
-     * @deprecated use {@link getDocumentReference()} instead
+     * @deprecated use {@link #getDocumentReference()} instead
      */
     @Deprecated
     DocumentReference getDocument();
@@ -76,12 +75,10 @@ public interface PrimaryEntity
     XWikiDocument getXDocument();
 
     /**
-     * For use in script services which should use com.xpn.xwiki.api.Document instead of
-     * com.xpn.xwiki.doc.XWikiDocument.
+     * For use in script services which should use {@code com.xpn.xwiki.api.Document} instead of
+     * {@code com.xpn.xwiki.doc.XWikiDocument}.
      *
-     * @return a "secure" Document which is safe for use i nscripts bcause it does permission
-     *         checks in various getter methods and does not expose any functionality which may
-     *         modify any data unless the user has programming rights.
+     * @return a "secure" Document which is safe for use in scripts because it does permission checks in various methods
      */
     Document getSecureDocument();
 
--- a/components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
+++ b/components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
@@ -23,6 +23,9 @@ import org.xwiki.stability.Unstable;
 
 import org.json.JSONObject;
 
+import com.xpn.xwiki.api.Document;
+import com.xpn.xwiki.doc.XWikiDocument;
+
 /**
  * An XDocument containing a primary XObject, which gives it a special meaning, such as a User, a Patient Record, or a
  * Project.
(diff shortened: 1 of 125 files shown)


allasm:
The new commits that have never been reviewed are the last two. One is to fix rebase issues (and update new code added after the original work was done to use the new API), the other addresses @sdumitriu's last comment in the original pull request about exposing XWikiDocument to velocity via insecure Patient and Family objects by family-studies script services

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/SecureFamily.java
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Constructor that wraps around another (supposedly non-secure) instance of a Family.
+     *
+     * @param family the family to have a secure wrapper around
+     */
+    public SecureFamily(Family family)
+    {
+        if (family == null) {
+            throw new UnsupportedOperationException();
```
veronikaslc: It's either IllegalArgumentException or no exception at all like in all other constructors from this PR, like SecurePatient for example. For consistency I propose removing that check.

components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
```diff
@@ -52,10 +55,34 @@
...
+     * @deprecated use {@link #getDocumentReference()} instead
      */
+    @Deprecated
     DocumentReference getDocument();
 
     /**
+     * Returns a reference to the document where the entity is stored.
+     *
+     * @return a valid document reference
+     */
```
sdumitriu: The new methods should have `@since 1.4`

components/entities/api/src/main/java/org/phenotips/entities/PrimaryEntity.java
```diff
@@ -52,10 +55,34 @@
      * Returns a reference to the document where the entity is stored.
      *
      * @return a valid document reference
+     * @deprecated use {@link #getDocumentReference()} instead
```
sdumitriu: @deprecated since 1.4; use {@link #getDocumentReference()} instead

components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/script/PermissionsManagerScriptService.java
```diff
@@ -92,13 +107,29 @@ public AccessLevel resolveAccessLevel(String name)
         return this.manager.resolveAccessLevel(name);
     }
 
-    public PatientAccess getPatientAccess(Patient targetPatient)
+    public PatientAccess getPatientAccess(String targetPatientId)
```
sdumitriu: The old method should be left in place, too, for backwards compatibility, and it can just do `return getPatientAccess(serializer.serialize(targetPatient.getDocumentReference());`

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/SecureFamily.java
allasm: agree, I think it is the other way around, i.e. `SecurePatient` should throw as well. I just ran into this problem (argument being `null`) here not there, but throwing here simplifies figuring out what happens a lot (as opposed to some random null pointer much later)

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -643,9 +643,7 @@ public DataSection documentInfoHeader(Set<String> enabledFields) throws Exceptio
 
     public DataSection documentInfoBody(Patient patient) throws Exception
     {
-        @SuppressWarnings("deprecation")
-        DocumentAccessBridge dab = Utils.getComponent(DocumentAccessBridge.class);
```
allasm: rebase issues, will fix

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/SecureFamilyTools.java
```diff
@@ -0,0 +1,139 @@
...
+ * A secure wrapper around FamilyTools which exposes a version of a Family object with
+ * no access to underlying XWikiDocument (implemented by {@link SecureFamily}).
+ *
+ * @version $Id$
+ * @since 1.4
+ */
+@Component(roles = FamilyTools.class)
+@Named("secure")
+@Singleton
+public class SecureFamilyTools implements FamilyTools
```
allasm: Agree, just Family classes (e.g. `FamilyTools`) were originally designed in a different way (compared to Patient classes\PatientRepository), and I noticed this discrepancy and did not like it, just as you said that was too many changes for one PR

allasm:
All comments were addressed in the last commit

marta-:
^ Wohoooo!  :fireworks:

PR #2307, opened 2017-05-15
PT-3226 Access Restrictions

AlexLapena:
@sdumitriu @marta- Please review and let me know if changes need to be made in the implementation.


components/patient-access-rules/api/src/main/java/org/phenotips/data/permissions/internal/VisibilityAccessAuthorizationModule.java
```diff
@@ -0,0 +1,77 @@
...
+            return null;
+        }
+
+        Visibility visibility = this.helper.getVisibility(patient);
+        if(visibility == null) {
+            return null;
+        }
+
+        // Checks if the visibility of Patient Record and the access rights
+        if(visibility.getName().equals("open") && access.getName().equals("edit")) {
```
sdumitriu: Also "view". Use this instead:

    if (visibility.getName().equals("open") && (access == Right.EDIT || Right.EDIT.getImpliedRights().contains(access))

AlexLapena:
Applied the changes @sdumitriu

marta-:
@sdumitriu should this PR also remove the code that generates XWikiRights objects for owner/collaborator/visibility objects?

sdumitriu:
Yes it should.

components/patient-access-rules/migrations/src/main/java/org/phenotips/data/permissions/internal/R71502PhenoTips3226DataMigration.java
```diff
@@ -0,0 +1,142 @@
...
+    public Object doInHibernate(final Session session) throws HibernateException, XWikiException
+    {
+        XWikiContext context = getXWikiContext();
+        XWiki xwiki = context.getWiki();
+        final DocumentReference rightsClassReference = this.entityResolver.resolve(RIGHTS_CLASS);
+        // FIXME this query does not generate results
+        Query q =
+            session.createQuery("select distinct doc.fullName from XWikiDocument as doc, BaseObject as p, BaseObject as r where "
+                + "p.name = doc.fullName and p.className = '" + this.serializer.serialize(Patient.CLASS_REFERENCE)
+                + "' and r.name = doc.fullName and r.className = '" + this.serializer.serialize(rightsClassReference) + "'");
```
AlexLapena: @sdumitriu It looks like this query is not retrieving any documents for some reason, but it looks okay to me. Do you see anything that may be an issue here?

PR #2312, opened 2017-05-17
PT-3240 Remove Export Redundancies

AlexLapena:
Removes unnecessary options within JSON and Excel export options.

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -34,6 +34,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -89,19 +90,17 @@ public class DataToCellConverter
                 "phenotype",
                 "phenotype_meta"
             };
-        // FIXME These will not work properly in different configurations
         String[][] headerIds =
             {
-                { "name", "id", "positive", "negative" },
-                { "meta.name", "name", "positive", "negative" },
-                { "type" }
+                { "present", "name", "id" },
+                { "present", "name", "id", "meta.name", "meta.id" }
             };
 
         Set<String> present = addHeaders(fieldIds, headerIds, enabledFields);
         this.enabledHeaderIdsBySection.put(sectionName, present);
         this.phenotypeHelper = new ConversionHelpers();
-        this.phenotypeHelper
-            .featureSetUp(present.contains("positive"), present.contains("negative"), present.contains("type"));
+        // FIXME "type" is never exposed
+        this.phenotypeHelper.featureSetUp(present.contains("type"));
     }
 
     public DataSection phenotypeHeader() throws Exception
@@ -114,17 +113,7 @@ public class DataToCellConverter
 
         DataSection section = new DataSection();
         int hX = 0;
-        if (present.contains("positive") && present.contains("negative")) {
-            DataCell cell =
-                new DataCell(this.translationManager.translate("phenotips.export.excel.label.phenotype.present"),
-                    hX, 1, StyleOption.HEADER);
-            section.addCell(cell);
-            hX++;
-        }
-        for (String headerId : Arrays.asList("type", "name", "id", "meta.name", "meta.id")) {
-            if (!present.contains(headerId)) {
-                continue;
-            }
+        for (String headerId : present) {
             DataCell cell = new DataCell(this.translationManager.translate("phenotips.export.excel.label.phenotype."
                 + headerId), hX, 1, StyleOption.HEADER);
             section.addCell(cell);
@@ -152,8 +141,9 @@ public class DataToCellConverter
         int x;
         int y = 0;
         Set<? extends Feature> features = patient.getFeatures();
+
         this.phenotypeHelper.newPatient();
-        Boolean categoriesEnabled = present.contains("type");
+        boolean categoriesEnabled = present.contains("type");
         List<Feature> sortedFeatures;
         Map<String, String> sectionFeatureLookup = new HashMap<>();
         if (!categoriesEnabled) {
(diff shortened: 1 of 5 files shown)


components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -119,34 +121,17 @@
...
         &lt;li&gt;
-          &lt;label for="${targetFieldName}_${field.name}_code_meta"&gt;
-          &lt;input type="checkbox" name="${targetFieldName}" id="${targetFieldName}_${field.name}_code_meta" value="${field.name}_code_meta" #if($valueList.contains("${field.name}_code_meta")) checked="checked" #end /&gt;
-          ${field.translatedPrettyName} $services.localization.render('phenotips.exportPreferences.field.termAndMetaID')
-        &lt;/li&gt;
-        &lt;li&gt;
           &lt;label for="${targetFieldName}_${field.name}_meta"&gt;
           &lt;input type="checkbox" name="${targetFieldName}" id="${targetFieldName}_${field.name}_meta" value="${field.name}_meta" #if($valueList.contains("${field.name}_meta") || $valueListisEmpty) checked="checked" #end /&gt;
-          ${field.translatedPrettyName} $services.localization.render('phenotips.exportPreferences.field.termAndMetaNames')
+          ${field.translatedPrettyName} $services.localization.render('phenotips.exportPreferences.field.qualifiers')
```
sdumitriu: This kind of translation chunking doesn't work, not all languages use the same topic. The full sentence should be translated.

AlexLapena:
@sdumitriu Code review changes added and rebased

sashaandjic:
Looks much better now without redundant options. However, both export dialogs (excel and JSON) are still quite ugly. Many fields that should be grouped together are showin somewhere else (e.g. Clinical, Final diagnosis and Additional comments should be displayed together). We should make both dialogs ui wise consistent with push dialog fields selection.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -86,28 +86,15 @@ public void phenotypeSetup(Set<String> enabledFields) throws Exception
...
-            { "id", "positive" },
-            { "name", "id", "positive" },
-            { "meta.id", "name", "positive" },
-            { "meta.name", "name", "positive" },
-            { "name", "negative" },
-            { "id", "negative" },
-            { "id", "name", "negative" },
-            { "type" }
+                { "name", "id", "positive", "negative" },
+                { "meta.name", "name", "positive", "negative" },
```
sdumitriu: `meta.id` should still be used, don't remove it.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -86,28 +86,15 @@ public void phenotypeSetup(Set<String> enabledFields) throws Exception
...
+                { "type" }
```
sdumitriu: This one has no equivalent trigger in `fieldIds`, so it's dead.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -182,7 +168,7 @@ public DataSection phenotypeBody(Patient patient) throws Exception
         for (Feature feature : sortedFeatures) {
             x = 0;
 
-            if (bothTypes && lastStatus != feature.isPresent()) {
+            if (lastStatus != feature.isPresent()) {
```
sdumitriu: Bug: if the patient only has negative phenotypes, no label will be printed.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -207,48 +191,28 @@ public DataSection phenotypeBody(Patient patient) throws Exception
...
-                        DataCell cell = new DataCell(meta.getName(), mX + offset, y);
-                        section.addCell(cell);
-                        offset += 1;
-                    }
-                    if (present.contains("meta.id")) {
-                        DataCell cell = new DataCell(meta.getId(), mX + offset, y);
-                        section.addCell(cell);
-                        offset += 1;
-                    }
+                    section.addCell(new DataCell(meta.getName(), mX + offset, y));
```
sdumitriu: Now offset is always `0`, so it's useless.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -207,48 +191,28 @@ public DataSection phenotypeBody(Patient patient) throws Exception
...
                 Boolean metaPresent = !featureMetadatum.isEmpty();
                 int offset = 0;
                 for (FeatureMetadatum meta : featureMetadatum) {
                     offset = 0;
-                    if (present.contains("meta.name")) {
-                        DataCell cell = new DataCell(meta.getName(), mX + offset, y);
-                        section.addCell(cell);
-                        offset += 1;
-                    }
-                    if (present.contains("meta.id")) {
```
sdumitriu: The id shouldn't be removed.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -274,27 +238,28 @@ public void variantsSetup(Set<String> enabledFields) throws Exception
         String sectionName = "variants";
         String[] fieldIds =
             { "variants", "variants_protein", "variants_transcript", "variants_dbsnp", "variants_zygosity",
-            "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence",
-            "variants_segregation", "variants_sanger", "variants_coordinates" };
+                "variants_effect", "variants_interpretation", "variants_inheritance", "variants_evidence",
```
sdumitriu: These changes are irrelevant for this issue, so they shouldn't be included in the pull request.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1116,25 +1070,16 @@ public DataSection prenatalPhenotypeBody(Patient patient) throws Exception
...
-                x++;
-            }
-            if (present.contains("id")) {
-                DataCell cell = new DataCell(feature.getId(), x, y, StyleOption.FEATURE_SEPARATOR);
-                section.addCell(cell);
-                x++;
-            }
+            DataCell cell = new DataCell(feature.getName(), ++x, y, StyleOption.FEATURE_SEPARATOR);
+            section.addCell(cell);
+            section.addCell(new DataCell(feature.getId(), ++x, y, StyleOption.FEATURE_SEPARATOR));
```
sdumitriu: These two cells are created the same way, so why two different ways? The name cell should also be added directly, without storing it in a temporary variable.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1147,48 +1092,30 @@ public DataSection prenatalPhenotypeBody(Patient patient) throws Exception
     public DataSection disordersHeaders(Set<String> enabledFields) throws Exception
     {
         String sectionName = "disorders";
+        Set<String> present = new HashSet<>();
+        present.add("omim_id"); //Accepts one option only
```
sdumitriu: Use `Set<String> present = Collections.singleton("omim_id");`

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1205,32 +1132,21 @@ public DataSection disordersBody(Patient patient)
...
-                x++;
-            }
+            DataCell cell = new DataCell("", x, y);
+            bodySection.addCell(cell);
+            x++;
+            bodySection.addCell(new DataCell("", x, y));
+            x++;
         }
         /* Notes export */
         if (present.contains("notes")) {
```
sdumitriu: This `if` is now redundant, and actually leaving it in means that notes will not be displayed anymore.

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -109,6 +110,8 @@
   {{html clean="false"}}
     #set($valueListisEmpty = ($valueList.size() == 0))
     #set($xclass = $xwiki.getDocument($xclassName).getxWikiClass())
+    #set ($discard = $ignored.fields.add('negative_phenotype'))
+    #set ($discard = $ignored.fields.add('negative_prenatal_phenotype'))
```
sdumitriu: Why? Aren't these already included in `$ignored.fields`?

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
AlexLapena: I used my Phenotips formatting tool before the PR, should I have only used it on the section of code I was working on?

sdumitriu: It's OK to use it everywhere, but you can skip these changes when doing a specific pull request.

PR #2316, opened 2017-05-29
PT-2613: Collapse/expand arrow in the current phenotype selection category sub-lists now works for dynamically inserted elements

a-gvo:


components/vocabularies/hpo/ui/src/main/resources/PhenoTips/HPOBranchDisplayer.xml
```diff
@@ -202,6 +202,19 @@ $xwiki.jsx.use('PhenoTips.HPOBranchDisplayer')##
   });
 });
 document.observe('xwiki:dom:loaded', function(){
+  $('current-phenotype-selection').observe('DOMNodeInserted', function(event) {
```
sdumitriu: `DOMNodeInserted` is deprecated and will soon be unsupported.

Instead of adding listeners for every expand tool, since events bubble up another solution is to add a single global listener on the body, which checks if the actual target is such an expand tool.

sashaandjic:
tested.

PR #2317, opened 2017-05-29
PT-1393: Improve pedigree node colouring and legend

allasm:
@sashaandjic please test, @veronikaslc please review

Before merging we need to discuss default colouring for phenotypes (minor point, but wrong colours can be annoying)

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -251,32 +251,30 @@
 .legend-container span {
   font-family: Arial,Verdana,sans-serif;
 }
-.legend-box-controls {
+.legend-box-controls-open {
+  background: #FFF;
+  height: 20px;
+  border-bottom: 1px solid;
+  border-color: #ccc;
+}
+.legend-box-controls-closed {
   background: #FFF;
   height: 25px;
-  margin-bottom: -12px;
+  cursor: pointer;
 }
-.legend-box-controls .legend-box-button {
+.legend-box-button {
   float: right;
   margin-top: 3px;
   margin-right: 3px;
   cursor: pointer;
 }
-.legend-box-controls .legend-box-button:hover {
+.legend-box-button:hover {
   font-weight: bold;
 }
-.legend-hide-show-button {
+.legend-hide-show-checkbox {
   cursor: pointer;
   float: right;
-  font-size: 12px;
-  color: black;
-  margin-top: 2px;
-  background: #ddd;
-  padding: 1px 3px 1px 4px;
-  margin-left: 4px;
-}
-.legend-hide-show-button:hover {
-  background: #B1C2E2;
+  margin-bottom: -1px;
 }
 .legend-box .legend-title {
   font-size: 1.1em;
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -224,6 +224,18 @@
 }
 .legend-info {
   font-size: .8em;
+  padding-top: 14px;
+  padding-bottom: 5px;
+  margin-bottom: -9px;
+}
+.legend-hidden {
+  display:none !important;
+}
+.legend-minimized-title {
+  float: left;
+  padding-top: 3px;
+  padding-left: 10px;
+  font-size: 1.1em;
 }
 .legend-box {
   background: #FFF;
@@ -235,11 +247,36 @@
   position: absolute;
   right: 0;
   z-index: 10000;
-  min-width: 140px;
-  max-width: 255px;
 }
 .legend-container span {
   font-family: Arial,Verdana,sans-serif;
+}
+.legend-box-controls {
+  background: #FFF;
+  height: 25px;
+  margin-bottom: -12px;
+}
+.legend-box-controls .legend-box-button {
+  float: right;
+  margin-top: 3px;
+  margin-right: 3px;
+  cursor: pointer;
+}
+.legend-box-controls .legend-box-button:hover {
+  font-weight: bold;
+}
+.legend-hide-show-button {
+  cursor: pointer;
+  float: right;
+  font-size: 12px;
+  color: black;
+  margin-top: 2px;
+  background: #ddd;
+  padding: 1px 3px 1px 4px;
+  margin-left: 4px;
+}
+.legend-hide-show-button:hover {
+  background: #B1C2E2;
 }
 .legend-box .legend-title {
   font-size: 1.1em;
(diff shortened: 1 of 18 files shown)


marta-:
Some entries in the legend are not quite readable when hovered because the background is too dark for the default text color. It would be nice if dark background on hover would be associated with white text color, while light background would have white text color. It's not an issue introduced in this PR, but since we're improving the legend coloring and interaction here, why not tackle this aspect, too.

FTR, `lightness = (max (r,g,b) + min (r,g,b)) / 2 `.

I see the colors are generated in hex directly, so here's a little function example that returns the lightness for a hex color as a number in `[0, 1]`:

```
var l = function(hexcolor) {
  var decnb = Number.parseInt(hexcolor.substring(hexcolor.indexOf('#') + 1, hexcolor.length), 16);
  var r = (decnb >>> 16) & 255;
  var g = (decnb >>> 8) & 255;
  var b = decnb & 255;
  return (Math.min(r,g,b) + Math.max(r,g,b))/2/255;
}
```

@allasm Would it be possible to also incorporate in this PR changing the text color on hover according to the background lightness?

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -235,12 +247,37 @@
   position: absolute;
   right: 0;
   z-index: 10000;
-  min-width: 140px;
-  max-width: 255px;
```
marta-: Just curious why this change.

sashaandjic:
Can we come up with better wording on "show all" and "hide all" buttons? "Colour on" /"colour off"?

marta-:
@sashaandjic Technically it's a selection, so we could say 'select all | none'.

sashaandjic:
This does not look good
screen shot 2017-06-01 at 11 22 28 am

sashaandjic:
"All" and "none" are fine. "Hide" troubles me, as we are not hiding anything. We are highlighting certain items/categories. Also "+" and "-" may confuse user.

allasm:
@marta- I will play with brightness, I'm not sure what is the problem, but will figure it out

@sashaandjic well, definitely Britney should have a (final) say in this, but we are either showing or hiding graphic for certain disorders, don't we? I tried checkboxes in one of the early implementations and it was ugly-sh.

components/pedigree/resources/src/main/resources/pedigree/editor.css
allasm: The legend used to be more or less fixed size (auto-scaling within 140-255px limits). Now there is anew control which lets the legend be as wide as necessary to avoid splitting a name into two lines, which I think is more convenient on (very) wide displays. By default it is more or less the same size as before, just now it is dynamically controlled (css classes or styles are added in JS depending on user clicks on control buttons - or lack there of)

sashaandjic:
Here is the screenshot for the issue that Marta reported. screen shot 2017-06-01 at 3 03 45 pm

allasm:
I think i addressed most/all of the comments

sashaandjic:
@allasm : here is what the team came up on today's meeting:

marta-:
@allasm can you please attach a screenshot of your work in progress on the requested changes, for discussion? Thanks.

allasm:
pedigree-legend-onhover
pedigree-legend-regular

sashaandjic:
nice!

allasm:
Not any more. Also user preferences (for no - hidden "you can drag" hints) are saved via browser cokies (@sashaandjic please test, @veronikaslc please review)

sashaandjic:
- shrink and maximize buttons are great addition; i would change the "maximize" tooltip to read "expand"
- once legend's infobox is closed, it will never show up (not sure if this is a good idea); maybe we can keep the little "i" button at al times so users can open and read the info
- tooltips should be added on round and square checkboxes
- it bothers me that we have checkboxes on both sides. shall we try to move all of them to the right and make them all square? I would also make unchecked ones white rather than gray

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -235,15 +247,42 @@
...
+  margin-right: 3px;
+  cursor: pointer;
+}
+.legend-box-button:hover {
+  font-weight: bold;
+}
+.legend-hide-show-checkbox {
+  cursor: pointer;
+  float: right;
+  margin-bottom: -1px;
```
marta-: With this negative margin, on my screen (FF54/Linux, all zoom levels) the chechbox looks ugly:

image

If your intention was to prevent the `border-bottom` from being pushed down, I suggest a negative `margin-top` instead.

marta-:
* The expand/collapse buttons (<< and >>) should be at the left edge
* Like Sasha suggested a while ago, when an abnormality is _not_ selected, the marker should be an empty circle rather than a full gray one, since it's closer to the representation on the actual pedigree.
* I'm not a fan of having checkmarks with different styles on two sides of the legend either. We should consider having them all on the right side at least, if you disagree with all on the left side (obviously the alignment needs work, this is just a poc screenshot):
image
* All the newly added controls (the maximize/minimize "buttons", the checkboxes) should have appropriate, clear tooltips.

marta-:
Looks great.

I think there's one more thing that can be done to make the feature even more user friendly easier to discover (and it was discussed before):
* on hovering an entry that is "off", display a check mark inside to show that you can interact with it and hint what the interaction does (the title is great, quite long, and it also has a delay showing up in most browsers);
* on hovering an entry that is "on", display either a minus or a cross inside.

I'm suggesting these because i think that, while people will easily figure out that they can turn on and off ALL diseases/genes/phenotypes thanks to the checkboxes,  they might not think they can do the same for individual entries.

What do others think? @sashaandjic @veronikaslc @sdumitriu

sashaandjic:
https://phenotips.atlassian.net/browse/PT-3321seems to be fixed too

allasm:
@sashaandjic: re: gene colouring: I think there is a separate issue for gene colouring and it should be handled there (btw, there is also an issue for displaying gene names under pedigree node in the comments, I think those two should be sub-tasks of one task to "redesign gene display in pedigree"

@marta- : I will test and check i it looks reasonably clear/not annoying

allasm:
> I think there's one more thing that can be done....
@marta- - like this?

marta-:
Thanks @allasm , looks great. There's one more bug with the new symbols showing up when selecting the whole category.

image

allasm:
Fixed

marta-:
Thanks, looks great, can't wait to see this merged.

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -48,6 +48,16 @@
   color: $theme.notificationSuccessColor !important;
   border-color: $theme.notificationSuccessColor !important;
 }
+
+input[type="checkbox"] {
+  outline: 0 none !important;
+  box-shadow: none !important;
+  -moz-transform: scale(.85) !important;
```
sdumitriu: Why `-moz-`?

components/pedigree/resources/src/main/resources/pedigree/model/helpers.js
```diff
@@ -221,6 +221,39 @@ define ([], function(){
       return obj;
     }
 
+    // expects two strings as input in the format of a web hex color: "#HHHHHH"
+    //
+    // optional minBrightnbess and maxBrightness parameters specify min and max brightness for the individual color
+    // channel of the reuslting color (e.g. background color should not be so dark as to make foreground text unreadable)
+    //
+    // returns a color in the "#HHHHH" format
```
sdumitriu: Should have 6 H, not 5.

components/pedigree/resources/src/main/resources/pedigree/editor.css
marta-: There's a layout bug in Firefox only, arising from the fact that the checkboxes are larger than in other browsers.

PR #2318, opened 2017-05-29
PT-3253: Clean up diagnoses titles in All Patient Data view

katpavlova:
I had trouble finding the view in which the original error was found, so @sashaandjic please test and approve before merging. :)


sashaandjic:
Tested
screen shot 2017-06-01 at 2 00 53 pm

PR #2321, opened 2017-06-07
PT-3216 Add info button for selected phenotypes

AlexLapena:


components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__phenotype.xml
```diff
@@ -578,10 +578,13 @@
...
   width: auto;
-  margin-right: 1em;
+  margin-right: 0.3em;
+}
+#current-phenotype-selection .summary-item .tools .xHelpButton {
+  margin : 0.6em;
 }
 #current-phenotype-selection .summary-item .tools {
-  color: $theme.textPrimaryColor;
+  color: #636B75;
```
sdumitriu: Hardcoded values are bad, they don't play nice with different color themes. Please find a color from the theme to use instead.

AlexLapena: Fixed, sorry about that.

components/patient-data/impl/src/main/resources/ApplicationResources.properties
```diff
@@ -130,6 +130,7 @@ phenotips.PatientSheetCode.removed=Removed '__name__'
 phenotips.PatientSheetCode.identifierExists=This identifier already exists
 phenotips.PatientSheetCode.PDFFooterDocInfo= Patient record last modified by {0} on {1}, exported by {2}.
 phenotips.PatientSheetCode.PDFFooterDisclaimer=This is a PDF of the PhenoTips record on {0} which may have since been modified. For the latest record please check:
+phenotips.patientSheetCode.termSuggest.aboutPhenotype=About this phenotype
```
marta-: Where is this used?

marta-:
There appears to be a z-index issue, as seen in this screenshot:

image

Can anyone else confirm? My instance is actually a mutant, so it's better to test on a clean build. @sashaandjic

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -842,7 +842,12 @@ document.observe('xwiki:dom:loaded', function()
                  newItem.insert({top: new Element('span', {'class': "fa fa-exclamation-triangle", title: "$escapetool.javascript($services.localization.render('phenotips.patientSheetCode.termSuggest.nonStandardPhenotype'))"}).update(' ')})
                }
                newItem.__source = inputElt;
+
                var tools = new Element('span', {'class' : 'tools hint'});
```
marta-: Let's drop the `hint` css class from here. It's from an early version of the tools, which had plain text not buttons, and it was supposed to make the font smaller and the color lighter. But now we're already overwriting the font size, and we want the default text color for the (i).

components/patient-data/ui/src/main/resources/PhenoTips/UIX_Field__phenotype.xml
```diff
@@ -578,10 +578,13 @@
...
   width: auto;
-  margin-right: 1em;
+  margin-right: 0.3em;
+}
+#current-phenotype-selection .summary-item .tools .xHelpButton {
+  margin : 0.6em;
 }
 #current-phenotype-selection .summary-item .tools {
-  color: $theme.textPrimaryColor;
+  color: $theme.textColor;
```
marta-: If we remove the `hint` css class for the `tools` element as sugested in the previous comment, the color will default to `$theme.textColor`, so this line doesn't need to be added (just remove `$theme.textPrimaryColor`).

marta-:
The z-index bug is caused by the [Touchscreen drag and drop](https://github.com/phenotips/phenotips/blob/master/components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml#L2907) code, which actually doesn't appear to be functional anymore. Is everyone OK with just removing it? @sdumitriu

sdumitriu:
OK for removing that code.

AlexLapena:
Should I remove that code in this PR?

AlexLapena:
Updated.

sashaandjic:
tested. problem Marta previously reported is not longer present.

PR #2325, opened 2017-06-13
PT-3261: Fixed multi-patient import through "Import JSON" option.

a-gvo:
Fixed as per @sdumitriu suggestion.


sashaandjic:
Tested! Works as expected!

PR #2326, opened 2017-06-14
PT-3215 Add info buttons for already selected diagnoses

AlexLapena:


sashaandjic:
Tested. Looks very nice. however, I noticed that our formatting in the info box is a bit inconsistent. For example one has button "READ ABOUT THIS ON OMIM.ORG..." and the other "OMIM:12345". Both buttons will take you to the same OMIM page. It would be nice to have same buttons labels as well. I will log this as a separate task.

PR #2327, opened 2017-06-15
[misc] Removed empty DocumentSheetBinding extension

veronikaslc:


sdumitriu:
This is not an empty object, it specifically tells XWiki to display this document as an UIX.

veronikaslc:
oh, sorry about that! didn't realize

PR #2335, opened 2017-06-22
PT-3240 Simplify the export dialog

AlexLapena:


AlexLapena:
@sashaandjic Could you please test this?

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -245,12 +245,13 @@ public void variantsSetup(Set<String> enabledFields) throws Exception
...
-                { "comments", "strategy", "status", "genes" } };
-        Set<String> present = addHeaders(fieldIds, headerIds, enabledFields);
+        Set<String> present = new LinkedHashSet<>();
+        if (enabledFields.remove(sectionName)) {
+            present.add("genes");
+            present.add("genes_status");
+            present.add("genes_strategy");
+            present.add("genes_comments");
+        }
         this.enabledHeaderIdsBySection.put(sectionName, present);
```
sdumitriu: Only if `sectionName` is in `enabledFields`.

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -120,11 +124,19 @@
       #end
       #foreach ($field in $xclass.properties)
         #if (($activeFields.size() == 0 || $activeFields.contains(${field.name})) &amp;&amp; !$ignored.fields.contains(${field.name}) &amp;&amp; !$ignored.types.contains(${field.type}))
-        &lt;li&gt;
-          &lt;label for="${targetFieldName}_${field.name}"&gt;
-          &lt;input type="checkbox" name="${targetFieldName}" id="${targetFieldName}_${field.name}" value="${field.name}" #if($valueList.contains(${field.name}) || $valueListisEmpty) checked="checked" #end /&gt;
-          ${field.translatedPrettyName}
-        &lt;/li&gt;
+          #if ($field.name == "apgar1")
```
sdumitriu: This is spaghetti code, a custom hack, don't do this. Instead mark both fields as ignored, and add `apgar` as a custom field, like `genes` and `variants`.

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -109,6 +109,10 @@
   {{html clean="false"}}
     #set($valueListisEmpty = ($valueList.size() == 0))
     #set($xclass = $xwiki.getDocument($xclassName).getxWikiClass())
+    #set($ignored = {
+      'fields' : ['apgar5', 'negative_phenotype', 'negative_prenatal_phenotype'],
```
sdumitriu: Don't override the value for this parameter, since it's defined someplace else. Instead, modify the place where `ignored` is defined. It's in two places lower in this file.

components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
AlexLapena: Ok, I'll make the fix.

sashaandjic:
Genetic data notes is shown twice.

AlexLapena:
@sdumitriu Made all the changes. @sashaandjic Would you please re-test?

AlexLapena:
Actually, it's looking like I need to address the JSON export now for apgar scores, I will do that now.

AlexLapena:
@sashaandjic I fixed the JSON, so it looks like it should be ready now!

PR #2338, opened 2017-06-26
PT-3279: "Delete all x" option does not work in the Advanced Search a…

marta-:
…nd in the Export dialog

### DIFF
--- a/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
@@ -1788,8 +1788,9 @@
   onItemAdded : function (element) {
   },
 
-  removeItem : function(event) {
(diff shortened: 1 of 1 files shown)


sashaandjic:
tested

sashaandjic:
tested. works as expected.

sashaandjic:
Push and Export dialogs doesn't take into account the user who's logged when filtering. E.g. in the following xar https://drive.google.com/open?id=0B1t2Eth-aBrha3JkRG5FcG9SbG8 I have user JohnDoe (password 123456) who has 4 patients with BRCA1 gene. When I log as MaryGreen (password 123456) who has no patients with BRCA1, export and push dialog show 4 at the bottom (John's patients). If I proceed with push/export nothing gets pushed/exported. Push fails, export file is empty. Advanced filtering option works as expected.

PR #2344, opened 2017-06-30
PT-3278: Export doesn't take into account the filters for genes and f…

marta-:
…or clinical diagnosis

Refactored the ExportFilter service to reuse the generic query builder of the LiveTable instead of generating the query from scratch filter by filter.


sashaandjic:
Push and Export dialogs doesn't take into account the user who's logged when filtering. E.g. in the following xar https://drive.google.com/open?id=0B1t2Eth-aBrha3JkRG5FcG9SbG8 I have user JohnDoe (password 123456) who has 4 patients with BRCA1 gene. When I log as MaryGreen (password 123456) who has no patients with BRCA1, export and push dialog show 4 at the bottom (John's patients). If I proceed with push/export nothing gets pushed/exported. Push fails, export file is empty. Advanced filtering option works as expected.

marta-:
@sashaandjic The problem you reported should be fixed by the last commit, please retest.

sashaandjic:
Tested. No problems found.

components/navigation/ui/src/main/resources/PhenoTips/LiveTableMacros.xml
```diff
@@ -0,0 +1,672 @@
...
+ #end
+#end
+
+#**
+ * Adds information about each live table row to the JSON returned by a live table data call.
+ * NOTE: This macro uses Velocity variables defined by gridresultwithfilter_buildQuery.
+ *
+ * @param $map stores the JSON in memory so that it can be adjusted before serialization
+ *#
+#macro(gridresult_buildRowsJSON $map)
```
sdumitriu: This is also defined identically at line 152, why the duplicate?

components/export/ui/src/main/resources/PhenoTips/ExportFilter.xml
```diff
@@ -37,231 +37,34 @@
...
-      #set ($query = ", DBStringListProperty disease ${query} and disease.id.id = obj.id and disease.id.name = 'omim_id' and (? in elements(disease.list)")
-      #set ($isFirstDisease = false)
-    #else
-      #set ($query = $query.concat(" ${joinOperator} ? in elements(disease.list)"))
-    #end
-    #set ($discard = $parameters.add($disease))
-  #end
+#set ($extra = " $!{request.filterWhere} ")
+#set ($params = [])
+#if ("$!request.space" != '')
```
sdumitriu: I wonder where these came from...

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatient.xml
```diff
@@ -1224,7 +1224,7 @@
                     onSuccess: function(response) {
                         var patients = [];
                         response.responseJSON.each(function(item) {
-                            patients.push(item.replace("data.",""));
+                            patients.push(item.substring(item.lastIndexOf('.') + 1));
```
sdumitriu: Would it be safer to check first if `contains('.')`?

marta-: Unnecessary. `lastIndexOf` returns -1 if `.` is not there, and then `substring` just returns the whole string.

components/navigation/ui/src/main/resources/PhenoTips/LiveTableMacros.xml
marta-: Leftover copy/paste, thanks for spotting it.

components/export/ui/src/main/resources/PhenoTips/ExportFilter.xml
marta-: From nowhere... that's why the filtering didn't work properly in Push and Export for non-admins. Sasha reported the issue and it was fixed in 67b4373 .

marta-:
@sdumitriu your comments were addressed.

PR #2345, opened 2017-07-03
PT-3186: Improve advanced data filtering form

marta-:


sashaandjic:
If we shorten text, we may be able to display options on their own separate line. For example:
Filter records with:
all
at least one diagnosis

all
at least one phenotype

all subcategories
exact phenotypes

Candidate genes
Confirmed Causative genes

The other suggestion is to italicize the text so it is consistent with format of "after" and "before"?

sashaandjic:
It appears a bit cluttered when having long and multiple disease/phenotype descriptions. Would it be possible to display only the part of the description that fits on one line with a tooltip showing the full text?
new filters
old

marta-:
@sashaandjic Changed long names as per your suggestion in latest commit. Not sure I agree with displaying filter options one per line, I personally don't mind the current look in the side filters (please note i fixed the positioning of "Delete all") and i think one per line would look bad in the export/push ui. Other opinions are welcome.

sashaandjic:
"Delete all" option is somewhat difficult to click on

sashaandjic:
under Phenotypes replace "Filter records O with all subcategories of O only matching exactly the selected phenotypes." with "Filter records that match the selected phenotypes O exactly O including subcategories"

marta-:
@sashaandjic Forgot to mention: I only tested my changes in Chrome & FF. For all I know it all might look like a train wreck in IE & Safari. Please make sure you or someone else tests in those browsers too.

marta-:
@sashaandjic Comments addressed.

sashaandjic:
looks very good! few minor complains:

marta-:
@sashaandjic :
1 and 2 are not specific to filters, that happens everywhere in PhenoTips. Please log them separately.
3: maybe we should have kept the "empty means no constraints" hints?

marta-:
@sashaandjic @veronikaslc @katpavlova did anyone test in Safari & IE?

sashaandjic:
sure...oh there is one more that I forgot: if I select more than one owner(referer) it will not return any records as it applies "AND" between them. Meanwhile, everywhere else OR is applied.

sashaandjic:
In Safari check boxes and radio buttons appear unaligned with their respective labels
screen shot 2017-07-06 at 12 20 10 pm

marta-:
@sashaandjic 
> if I select more than one owner(referer) it will not return any records as it applies "AND" between them. Meanwhile, everywhere else OR is applied.

The referrer seems to work fine, can you confirm there are problems only with the owner filter?

The issue with the owner is not that the values are AND-ed, but that when there's more than one owner value we're searching for local references (e.g. `XWiki.Admin`, `XWiki.jdoe`) while storing full references (`xwiki:XWiki:Admn`). This bug needs to be logged separately, as it affects the stable branch 1.3.x as well.

sashaandjic:
confirming the problem with the owner; by referrer and by last author are fine
screen shot 2017-07-06 at 2 30 35 pm

veronikaslc:
1. I still think we can use placeholders for dates "before:" and "after:", it's not that hard to learn
2. There is no way I can install windows 9 on my machine to test, I have downloaded IETester that allows to test in multiple versions of IE and all of them worked fine except IE9 that crashed on the page.

sashaandjic:
In Safari I noticed that radio buttons on group and study filters look much worse than on the patient and family filters (probably because they are thinner).
group filter
patient filter

sashaandjic:
it not the width, group and study filters have bullet points that are causing this anomaly.

sashaandjic:
One more thing, the date fields on groups and study filters need to be done as in the patient filter. 

Group/study filter:
dates on groups and studies

marta-:
@sashaandjic the last two issues are actually related to (and fixed by) #2346. I should rebase, since that one is already merged into master.

sashaandjic:
ok

veronikaslc:
should we re-review?

marta-:
@veronikaslc yes, please

components/navigation/ui/src/main/resources/PhenoTips/DBWebHomeSheet.xml
```diff
@@ -374,21 +377,144 @@ document.observe('xwiki:livetable:newrow', function(tableEvent){
...
+.entity-directory .xwiki-livetable-topfilters-tip .tippointer {
+  display: none;
+}
+
+/*---------------------------------------------------------------*/
+.entity-directory &gt; .xwiki-livetable-container.hidden-filters {
+    margin-left: 20px;
+}
+.entity-directory .hidden-filters .tipfilters {
+    xtransform: rotate(-90deg);
```
sdumitriu: Trash?

marta-: yep, sorry, removing.

components/navigation/ui/src/main/resources/PhenoTips/DatabaseFilters.xml
```diff
@@ -51,41 +51,41 @@
...
+* $services.localization.render('phenotips.databaseFilters.externalID') {{html clean=false}}&lt;input type="text" name="external_id" value=""/&gt;{{/html}}
 #end
 #if (!$xwiki.getDocument('PhenoTips.VisibilityClass').isNew())
-  * $services.localization.render('phenotips.databaseFilters.visibility') {{html clean=false}}#foreach ($visibility in $services.permissions.listVisibilityOptions())&lt;br /&gt;&lt;input type="checkbox" class="multipleSelection privacyExport" name="visibility" value="$visibility.name"#if ($visibility.name != 'hidden') checked="checked"#end /&gt; $visibility.label#end&lt;input type="hidden" name="visibility/class" value="PhenoTips.VisibilityClass"/&gt;{{/html}}
+  * $services.localization.render('phenotips.databaseFilters.visibility') {{html clean=false}}#foreach ($visibility in $services.permissions.listVisibilityOptions())&lt;br /&gt;&lt;label&gt;&lt;input type="checkbox" class="multipleSelection privacyExport" name="visibility" value="$visibility.name"#if ($visibility.name != 'hidden') checked="checked"#end /&gt; $visibility.label&lt;label&gt;#end&lt;input type="hidden" name="visibility/class" value="PhenoTips.VisibilityClass"/&gt;{{/html}}
```
sdumitriu: /label (you forgot the `/`)

PR #2346, opened 2017-07-03
PT-3284: Suggest widgets for vocabularies do not work on the filters for groups' owned and shared cases tables, and for study cases

marta-:
Fixed.


sashaandjic:
tested filtering on group cases, group shared cases and studies. works well.

PR #2348, opened 2017-07-04
PT-3081: Updating patient from JSON should allow updates and overwrites

a-gvo:


veronikaslc:
We should add that option explicitly to the UI. My vision:

veronikaslc:
How to test it? Later on we also need to update [PhenoTips RESTful API](https://phenotips.org/DevGuide/RESTfulAPI#HUpdatingapatientrecord).

a-gvo:
The REST endpoint for updating the patient from JSON can take a query param (policy=update/merge/replace). Though there should probably be a separate REST service for each? I think i may have seen a JIRA task for that.

veronikaslc:
Can you give example how you tested it?

a-gvo:
So, for example using ```rest/patients/P0000005?method=PUT&policy=update``` with a JSON in the body. The "update" policy should do what the JSON dialogue does currently, "merge" should combine old data with new data if it's in some collection (like list of genes), and "replace" should replace all patient data with that provided in the JSON.

veronikaslc:
What I mean is, if you use curl or something else how exactly you test?
Example of curl command line or other method you used.

a-gvo:
Well, for update I'd just use the JSON import dialog. For "merge" and "replace", i piggybacked off an existing POST in Firefox (so, ```Network -> click on some POST request -> Edit & Resend```)

firefox-post

That way it's easier to see the changes, just refresh the patient record :).

veronikaslc:
Thanks. Tested, works.

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
```diff
@@ -111,36 +121,139 @@
...
+     * @return a singleton list of medications, if medication name is known, an accumulated list otherwise
+     */
+    private List<Medication> mergeMedications(
+        @Nonnull final List<Medication> storedMeds,
+        @Nonnull final List<Medication> newMeds)
+    {
+        // If a medication does not have a specified name, then can't compare between medications. Store all.
+        // Otherwise, just store the new value.
+        return storedMeds.get(0).getName() == null
+            ? Stream.of(storedMeds, newMeds).flatMap(Collection::stream).collect(Collectors.toList())
```
veronikaslc: would `storedMeds.addAll(newMeds)` be more simple in that case?

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
```diff
@@ -111,36 +121,139 @@
...
+        // A stream of medication objects to save.
+        final Stream<Medication> medsStream;
+        if (PatientWritePolicy.MERGE.equals(policy)) {
+            final PatientData<Medication> storedMedications = load(patient);
+            medsStream = storedMedications != null
+                ? buildMergedMedicationsMap(medications, storedMedications).entrySet().stream()
+                    .flatMap(mapEntry -> mapEntry.getValue().stream())
+                : StreamSupport.stream(medications.spliterator(), false)
+                    .filter(Objects::nonNull);
+            docX.removeXObjects(Medication.CLASS_REFERENCE);
```
veronikaslc: if we anyways `docX.removeXObjects(Medication.CLASS_REFERENCE);` in both `if` and `else`, may be we can pool it out and place before `if`?

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
```diff
@@ -111,36 +121,139 @@
...
+        } else {
+            medsStream = StreamSupport.stream(medications.spliterator(), false)
```
veronikaslc: why not simply use `medications.stream()` if sequential stream is needed?

veronikaslc: *( in all other such places as well, where applicable)

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/FeaturesController.java
```diff
@@ -281,96 +297,205 @@ private JSONArray joinArrays(JSONArray jsonOne, JSONArray jsonTwo)
...
+        @Nonnull final XWikiContext context)
+    {
+        final Stream<Feature> featureStream = PatientWritePolicy.MERGE.equals(policy)
+            ? buildFeaturesStream(load(patient), features)
+            : StreamSupport.stream(features.spliterator(), false);
+
+        clearFeatureData(docX, dataHolder, context);
+
+        featureStream
+            .map(feature -> saveFeatureData(docX, feature, context))
```
veronikaslc: I'm kind of debating here. On one hand, looks short&nice. On the other - this is essentially `.forEach()` masked to `map()` to make it intermediate operation instead of terminal operation. The debate is: separating steps makes reading code easier but it results in couple extra lines. Personally, when everything is one concatenated chain of operations it looks cool but hard to parse whats actually going on in the code. Open to discussion.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/FeaturesController.java
```diff
@@ -281,96 +297,205 @@ private JSONArray joinArrays(JSONArray jsonOne, JSONArray jsonTwo)
...
+     * @return a merged map of feature ID to {@link Feature}
+     */
+    private Stream<Feature> buildFeaturesStream(
+        @Nullable final PatientData<Feature> storedFeatures,
+        @Nonnull final PatientData<Feature> features)
+    {
+        return storedFeatures == null
+            ? StreamSupport.stream(features.spliterator(), false)
+            : Stream.of(storedFeatures, features)
+                .flatMap(s -> StreamSupport.stream(s.spliterator(), false))
```
veronikaslc: would `.flatMap(s :: stream)` work?

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
a-gvo: Fixed

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
a-gvo: Fixed

components/patient-data-extra/medication/api/src/main/java/org/phenotips/data/internal/controller/MedicationController.java
a-gvo: PatientData extends Iterable, so can't call ```stream()``` on it directly.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/FeaturesController.java
a-gvo: In this case, I mostly wanted to take advantage of ```groupingBy```, since it's nice and short.
I haven't thought of this before, but the saving of feature data could also be moved to the following ```Collectors.mapping``` call, where i could save feature and return feature value. So that's 3 potential ways to do this. 

Implemented last, but open to discussion re: the best way to do this.

a-gvo:
Rebased and refactored.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/PhenoTipsPatient.java
```diff
@@ -220,6 +221,12 @@ public JSONObject toJSON(Collection<String> selectedFields)
     @Override
     public void updateFromJSON(JSONObject json)
     {
+        updateFromJSON(json, PatientWritePolicy.UPDATE);
+    }
+
+    @Override
+    public void updateFromJSON(JSONObject json, PatientWritePolicy policy)
```
allasm: I only realized this while working on the "improve entities" issue: `updateFromJSON()` is actually a `PrimaryEntity` method, should we have a more generic `EntityWritePolicy` instead of `PatientWritePolicy` and correspondingly have `updateFromJSON(JSON, Policy)` be a `PrimaryEntity` method? The default implementation may throw an "unimplemented exception" for all policies which are different from current default to make it compatible with all other current entities which do not support an "update" yet

allasm: Update: discussed this with @a-gvo, she thinks this is a good idea. I will implement that as part of the update entities branch. Note, since Entity update will be in 1.4m3, 1.4m2 will have this API which will change again for 1.4m3, hope that does not cause any problems (@sdumitriu?)

sdumitriu: This should indeed be part of the entities API. OK to have this temporary API in 1.4M2.

PR #2349, opened 2017-07-05
PT-3265: getTerm in HGNC vocabulary now matches on exact input provided

a-gvo:
Before testing, please reindex the HGNC vocabulary.


sashaandjic:
tested

sashaandjic:
@a-gvo I'm not sure what is happening, please see the screenshots 
screen shot 2017-07-14 at 1 00 53 am
screen shot 2017-07-14 at 12 49 08 am
screen shot 2017-07-14 at 12 49 30 am
screen shot 2017-07-14 at 12 53 40 am
screen shot 2017-07-14 at 12 56 36 am

marta-:
@sashaandjic did you reindex HGNC after building and before testing?

sashaandjic:
@marta- no, i didn't... hope it is a false alarm...

sashaandjic:
reindexed. now i'm getting different behaviour when typing "ANCR" and "lncRNA-ANCR". they are both aliases for DANCR.

a-gvo:
@sashaandjic Is the different behaviour you're talking about is that you get ANCR when you type ANCR, and DANCR when you type lncRNA-ANCR?

sashaandjic:
@a-gvo yes. I'm not sure why ANCR is still available as an option in the dropdown.

sashaandjic:
also when you type IncRNA-ANCR it offers ANCR; but when you pick it, it turns into DANCR

a-gvo:
@sashaandjic ANCR is available as an option in the dropdown because the dropdown uses the search function, rather than getTerm. I believe this takes into account possible spelling mistakes...so for example, if you type DRNCR, DANCR will be displayed. 

If you want to test getTerm directly, you have to do it through REST (```/rest/vocabularies/hgnc/ANCR```). In this case, If i input ANCR, i should get back a term for ANCR. If I input lncRNA-ANCR, I should get a term where this is an alias because there is no vocabulary term with lncRNA-ANCR as its symbol.

As for the second problem, I can't reproduce. Pick it how? Click, enter, etc?

sashaandjic:
I'm confused with ANCR. It is listed in HGNC, but has no ensebleID. How should we treat this one? Also  why am I getting the triangle before ANCR and DANCR in edit mode? it dissapears once i save the record?
screen shot 2017-07-14 at 12 54 51 pm

sashaandjic:
ANCR, AGU2 and lncRNA-ANCR are aliases for DANCR; when I type ANCR or lncRNA-ANCR I'm offered to pick ANCR and DANCR; if I type AGU2 only DANCR is shown up as option.

a-gvo:
ANCR is treated alike all genes without the ensemblID. We store ANCR as the ID. I can't seem to reproduce the triangle issue.

Re: ANCR, AGU2, and lncRNA-ANCR. When you search for ANCR you get ANCR because there is a vocabulary term with that symbol and DANCR because it has ANCR as an alias. Similarly with lncRNA-ANCR (it matches on the ANCR part of the entire term).

AGU2 is an alias for DANCR, which is why it appears. ANCR doesn't have AGU2 in its aliases.

sashaandjic:
to reproduce triangles:
- click on ADD GENE
- type "dancr"
- click on <enter>
- cclick on <tab>

a-gvo:
@sashaandjic Ohhh. Interesting. I can only reproduce it in Chrome, nowhere else. I think that's a separate issue though (same thing happens in Playground).

sashaandjic:
@a-gvo i went through first 44 suspicious genes and I got expected behaviour in all cases. the problem with triangles is very minor. i would not hold the release because of it.

sashaandjic:
one thing that i just discovered while testing lower cases of the genes of 17 pc patients that need to be fixed. if the following genes are typed in lower cases: cdh1, magi1, rai1, hpca, slep the uppercased matching gene won't be highlighted in the dropdown. So if user presses enter, the wrong gene will end up being entered.
screen shot 2017-07-14 at 2 49 32 pm

marta-:
@sashaandjic that is also happening in master, so it's unrelated to this PR. To be reported separately.

PR #2351, opened 2017-07-10
PT-3287: In the advanced data flters, when filtering by more than one…

marta-:
… owner no results are returned

Fixed by adding a special case for owners in the query builder.

### DIFF
--- a/components/navigation/ui/src/main/resources/PhenoTips/LiveTableMacros.xml
+++ b/components/navigation/ui/src/main/resources/PhenoTips/LiveTableMacros.xml
@@ -621,8 +621,8 @@
       #set($discard = $whereParams.add("${colname}"))
       #foreach ($value in $filterValueList)
         ## Owners are always stored as full references, and the query may come as a partial reference
-        #if ($colname == 'owner' &amp;&amp; !$filterValue.startsWith("$!{xcontext.database}:"))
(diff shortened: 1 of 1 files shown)


components/navigation/ui/src/main/resources/PhenoTips/LiveTableMacros.xml
```diff
@@ -620,7 +620,12 @@
       #set($whereSql = "${whereSql} and ${targetedObject}.id=${safe_tableAlias}.id.id and ${safe_tableAlias}.id.name = ? and ${safe_tableAlias}.value in (${stringtool.repeat('?', ', ', $filterValueList.size())})")
       #set($discard = $whereParams.add("${colname}"))
       #foreach ($value in $filterValueList)
-        #set($discard = $whereParams.add($value))
+        ## Owners are always stored as full references, and the query may come as a partial reference
+        #if ($colname == 'owner' &amp;&amp; !$filterValue.startsWith("$!{xcontext.database}:"))
```
sdumitriu: If the user is from another instance and has a different prefix, this fails.

marta-:
@sdumitriu comment addressed.

sashaandjic:
Not working as expected. If I pick more than one owner, it shows me the patients of the first owner that was entered only (which is better than what we had before - none).

PR #2353, opened 2017-07-11
PT-3294: A phenotype can appear as both positive and negative in the …

marta-:
…"Current selection" in certain conditions


components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -827,7 +827,7 @@ document.observe('xwiki:dom:loaded', function()
              if (forceUpdate) {
                 this._silentRemove(inputElt.value, type);
              }
-             var exists = this._find(inputElt.value, type);
+             var exists = this._find(inputElt.value, !inputElt.up('.accepted-suggestions, .dropdown') && type);
```
sdumitriu: Won't this search for the `.false` class?

PR #2354, opened 2017-07-11
PT-1853: Add fields for US/LS measurements and ratio

katpavlova:


components/patient-measurements/rest/src/main/java/org/phenotips/data/rest/internal/DefaultComputedMeasurementResourceImpl.java
```diff
@@ -57,37 +59,54 @@ public Response getComputedMeasurement(UriInfo uriInfo)
...
+                value = ((BMIMeasurementHandler) this.handlers.get(BMI))
+                .computeBMI(Double.parseDouble(dependencies[0]), Double.parseDouble(dependencies[1]));
+            } else if (USLS.equals(measurement)) {
+                // US/LS is the measurement to be computed. Pass deps 0 and 1 as the respective upperSeg and lowerSeg
+                value = ((USLSMeasurementHandler) this.handlers.get(USLS))
+                .computeUSLS(Double.parseDouble(dependencies[0]), Double.parseDouble(dependencies[1]));
+            } else {
+                // The measurement param passed to this method does not match any known computed measurement types
+                throw new IllegalArgumentException();
+            }
```
marta-: This implementation was wrong from the beginning from an architectural pov and did not provide a good example.
There should be no code that is specific to a certain measurement handler, because this approach is not scalable and configurable enough. The goal is to drop in a new handler and have the service work without additional changes. A few suggestions to achieve that:
* The method for getting the value for computed measurements should be generically named `compute` (what else is the BMI handler going to compute if not the BMI?).
* it should get a collection of named values as parameter (that part was good in the original and shouldn't have been changed), and handle the processing internally. Keep in mind that different measurements might have different numbers of dependencies.
* there should be no "if-else" for calling the appropriate handler and no casting to the specific handler's class. It should be as simple as `this.handlers.get(measurement).compute(params)` (inside an `if` or a `try/catch` in case someone tries to call this for a non-computed measurement or a non-existing measurement).

To handle the computed measurements more elegantly I suggest the following:
* the computed measurements should implement a second interface e.g. `ComputedMeasurementHandler` which provides the `compute` method
* the `isComputed` method in the abstract handler should return true if the the measurement implements this interface (which means removing the `isComputed` override from all computed measurements, since we already mark them as computed when we make them implement `ComputedMeasurementHandler`),

katpavlova: This comment will be addressed in a new task: [PT-3296](https://phenotips.atlassian.net/browse/PT-3296)

marta-:
@katpavlova Why did you merge this? The reviewers didn't approve it yet.

PR #2355, opened 2017-07-12
PT-3295: Selecting from the quick search suggestions a phenotype that is already selected and had details attached appears to cause the loss of the details

marta-:


sashaandjic:
tested. no problems found.

PR #2356, opened 2017-07-12
PT-3127: Rearrange pedigree import dialog to put the options above the actual file selection

AlexLapena:
This also redesigns the dialogue so that the file selection is similar to patient JSON import.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -447,9 +447,8 @@
 .import-selector .import-textarea {
   height: 10em;
   resize: none;
-  width: 84%;
   /*border: 1px solid #E8E8E8;*/
-  margin: .2em .2em 0 7.5em;
+  margin: .5em .2em 0 ;
   display: block;
 }
 .import-block {
@@ -460,12 +459,14 @@
   margin-top: 4px !important;
 }
 .import-file-button-wrapper {
-  margin-left: -494px;
+  position: relative;
+  margin-left: -350px;
 }
 .import-block-bottom {
   margin: 1.2em 0 0.5em 0;
 }
 .import-section {
+  margin-top: 0.5em;
   font-weight: bold;
   float: left;
 }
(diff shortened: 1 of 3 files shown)


AlexLapena:
@sashaandjic brought up an interesting point that our JSON import menu looks a lot better than our current pedigree import. I think rather than fixing the current spacing, we re-order the bottom section to say `Paste a {format} in the box below, or [Select A Local File]`

AlexLapena:
Furthermore, we also discussed that in the future, we could implement some sort of drop down menu for `Data Format` and `Options`. Currently, the import menu looks a bit less friendly than an import menu should, in my opinion.

sashaandjic:
It would also make sense to do the IMPORT and CANCEL buttons alignments the same way as in the Import dialog.

sashaandjic:
Do we need "," between "below" and SELECT LOCAL FILE bttn?

AlexLapena:
Fixed UI.

sashaandjic:
looks great. 2 minor things:
- remove comma from "Paste data in the box below, or"
- move the "Paste dta..." and the corresponding text box few lines down

AlexLapena:
@sashaandjic Fixed.

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -460,12 +459,14 @@
   margin-top: 4px !important;
 }
 .import-file-button-wrapper {
-  margin-left: -494px;
+  position: relative;
+  margin-left: -350px;
```
sdumitriu: This is extremely fragile... This whole CSS must be rewritten.

components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -460,12 +459,14 @@
...
 }
 .import-block-bottom {
   margin: 1.2em 0 0.5em 0;
 }
 .import-section {
+  margin-top: 0.5em;
```
sdumitriu: I don't like this, now the label and the radio buttons are misaligned.

sdumitriu:
Misaligned label:

screenshot_20170804_001714

AlexLapena:
I wasn't able to have enough time tonight to address all the issues unfortunately, but I will rewrite the fragile CSS part asap.
label

PR #2359, opened 2017-07-16
PT-3305 and PT-3306: Improved homepage without permissions module

sdumitriu:
To test without the permissions module: edit `resources/ui/pom.xml` and remove the dependency on `patient-access-rules-ui`, then rebuild.


sashaandjic:
Export to png fails with error msg "The file “FAM0000001_pedigree.png” could not be opened because it is empty."

marta-:
FTR the issue reported above (empty png) is not related to this PR and will be logged separately.

veronikaslc:
I have the following missing translations

PR #2361, opened 2017-07-17
PT-3304 Underlines new patient button on hover

AlexLapena:


AlexLapena:
@sdumitriu We have a CSS rule for it, however it seems to not want to work unless I create the a tag.

https://github.com/phenotips/phenotips/blob/79c2c7c4fbd47d4be9d335352e56f0bb71218520/components/skin/ui/src/main/resources/PhenoTips/Skin.xml#L871

marta-:
@AlexLapena actually no, that rule would not apply to this element. It's for elements with `id="create-patient-record"`.

AlexLapena:
@marta- I just noticed that ^^

AlexLapena:
Fixed and rebased.

PR #2362, opened 2017-07-17
PT-3309: Remove quotation marks from Export png/svg buttons

marta-:
Done. Accidentally introduced when translating the pedigree thumbnail menu in commit 6379c7e93e4fb22 for https://phenotips.atlassian.net/browse/PT-2849 .


veronikaslc:
I think this was could be merged without codereview :)

PR #2367, opened 2017-07-18
PT-3308: Export to PNG SVG initiated in pedigree editor is missing person's details other than internalID

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
```diff
@@ -1055,6 +1055,12 @@ define([
                 }
             }
 
+            // move white background up if it is not the first child element of <svg>, otherwise it will hide labels
```
sdumitriu: This seems misplaced, why does it have to be done in the per-node `drawLabels`? The background is global, so it should, in theory, be placed as the first element only once. If it doesn't stay there, I'd say that other code is buggy, and this is just a hacky fix.

veronikaslc: you're right, I placed it at the place where the nodes are actually shifted. I better move it up in stack.

veronikaslc:
@sdumitriu updated with new commit

PR #2374, opened 2017-07-20
PT-3315: Phenotypes from the predefined list disappear from phenotype…

marta-:
… selection after JSON import from edit mode


marta-:
@veronikaslc Please test In IE before and after the fix. Thanks!

PR #2375, opened 2017-07-20
PT-3051: Allow assigning Alive and Well status to a group of people in pedigree editor

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/personGroupVisuals.js
```diff
@@ -26,14 +26,6 @@ define([
             this.setNumPersons(node.getNumPersons());
         },
 
-        generateHoverbox: function(x, y) {
-            if (editor.isReadOnlyMode()) {
-                return new ReadOnlyHoverbox(this.getNode(), x, y, this.getGenderGraphics());
-            } else {
-                return new PersonGroupHoverbox(this.getNode(), x, y, this.getGenderGraphics());
```
allasm: I think with this change PersonGroupHoverbox is never used. It may not be a wrong thing in itself, but then we don't need the corresponding class/file. I need a bit more time to read the code, requesting a change now to make sure it does not get merged until checked

veronikaslc: @allasm I better just remove this separate [misc] commit, it's not related to the issue anyways.

veronikaslc: commit removed

allasm: it was related to the issue as all your changes to PersonGroupHoverbox were negated by the commit (JFYI - the hoverbox type for a node is decided by the method you removed, so after the change group nodes would inherit the parent method, which generates a PersonHoverbox, which has stuff not relevant to groups). Now looks good

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -496,7 +496,7 @@ define([
                 editor.getUndoRedoManager().addState( event, undoEvent );
             }
 
-            editor.getNodeMenu().update();  // for example, user selected a wrong gender in the nodeMenu, which
+            ((node.getType() == 'PersonGroup') ? editor.getNodeGroupMenu() : editor.getNodeMenu()).update();  // for example, user selected a wrong gender in the nodeMenu, which
```
allasm: we should improve it and make it more object-oriented - e.g. a `node` should know which "node menu" it uses (which it does, as it creates it, just does not have a way to report it as of right now) instead of using an explicit `if/else`, but that can be a `TODO`

components/pedigree/resources/src/main/resources/pedigree/pedigreeEditorParameters.js
```diff
@@ -51,6 +51,7 @@ define([], function(){
         touchOrbRadius: 8,
         personHoverBoxWidth: 180,  // 160 for old handles
         personHoverBoxHeight: 180, // 160 for old handles
+        personGroupHoverBoxHeight: 110,
```
allasm: Actually, this looks a bit ugly, lets make it look a bit more like it used to

allasm: `A&W` label is also somewhat misplaced for groups

allasm:
@veronikaslc please review

veronikaslc:
+1, I can not submit review for my own PR other way.

allasm:
@sashaandjic please test (PT-3320 & PT-PT-3051, but also general node menu operation, including relationships, as the under-the-hood code has changed a bit)

allasm:
re: commit for PT-3320: yes and no, they are somewhat related and may have side effects affecting each other, so testing together in one branch is useful IMHO

allasm:
As discussed, new issues were done in separate branches, and I locally tested the commits separately as well - is this good to go or we need to split it into two branches (which at this point for these particular commits seems like a bureaucracy work to me)?

PR #2377, opened 2017-07-25
PT-2323 Refactor REST to use the delete method from PatientRepository

AlexLapena:


components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -340,23 +338,19 @@ public void deletePatientCatchesXWikiException() throws XWikiException
...
-        XWiki wiki = mock(XWiki.class);
-        XWikiDocument patientXWikiDoc = mock(XWikiDocument.class);
-        doReturn(wiki).when(this.context).getWiki();
-        doReturn(patientXWikiDoc).when(this.patient).getXDocument();
         doReturn(true).when(this.access).hasAccess(Right.DELETE, this.userProfileDocument, this.patientDocument);
 
         Response response = this.patientResource.deletePatient(this.id);
 
-        verify(wiki).deleteDocument(patientXWikiDoc, this.context);
+        verify(this.logger).debug("Deleted patient record [{}]", this.id);
```
allasm: a minor point: I would rather check that no exceptions were thrown, or just removed this check and only left the check below (for the return type)

allasm: ...also not sure if we want to have this method be declared as `throws XWikiException`, maybe use `WebApplicationException` or catch all exceptions?

AlexLapena:
Updated & rebased.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientResourceImplTest.java
```diff
@@ -340,23 +335,18 @@ public void deletePatientCatchesXWikiException() throws XWikiException
...
     {
-        XWiki wiki = mock(XWiki.class);
-        XWikiDocument patientXWikiDoc = mock(XWikiDocument.class);
-        doReturn(wiki).when(this.context).getWiki();
-        doReturn(patientXWikiDoc).when(this.patient).getXDocument();
         doReturn(true).when(this.access).hasAccess(Right.DELETE, this.userProfileDocument, this.patientDocument);
 
         Response response = this.patientResource.deletePatient(this.id);
 
-        verify(wiki).deleteDocument(patientXWikiDoc, this.context);
```
sdumitriu: You should replace this with a verification that the repository was invoked.

sdumitriu:
You should do the same for the other resource, `DefaultPatientByExternalIdResourceImpl.java`.

AlexLapena:
Added the change for the other resource.

PR #2381, opened 2017-07-28
PT-3248 and PT-3275

allasm:
(both are simple text changes)

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
@@ -105,7 +105,7 @@ define([
             },
             {
                 'name' : 'disorders',
-                'label' : 'Known disorders of this individual',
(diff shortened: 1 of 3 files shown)


components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
```diff
@@ -368,7 +368,7 @@ define([
             },
             {
                 'name' : 'disorders',
-                'label' : 'Known disorders<br/>(common to all individuals in the group)',
+                'label' : 'Final diagnos<br/>(common to all individuals in the group)',
```
marta-: Typo?

allasm: yes

allasm: fixed

PR #2382, opened 2017-07-28
PT-1462 REST Pedigree format converter

AlexLapena:


components/pedigree/rest/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+            <version>${xwiki.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>patient-data-rest</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>family-studies-api</artifactId>
```
allasm: I think this is an unnecessary dependency. @veronikaslc, how is that project to analyze  dependencies doing, I remember yo did a nice UI for browsing the results of the analysis?

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/PedigreeConverter.java
```diff
@@ -0,0 +1,48 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.pedigree.rest;
+
+import javax.ws.rs.*;
```
allasm: why do we need it,  why a `.*`?

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/PedigreeConverter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.phenotips.rest.ParentResource;
+
+import javax.ws.rs.core.MediaType;
+import org.xwiki.rest.resources.RootResource;
+import javax.ws.rs.core.Response;
+
+/**
+ * Resource for converting PED files to JSON with REST
```
allasm: I would reword to something more detailed, e.g. "convert pedigree PED files to PhenoTips pedigree SImpleJSON format" (since there may be many different kinds of JSON  =however probably the actual output format will change before this gets released :) )

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/internal/PedigreeConverterImpl.java
```diff
@@ -0,0 +1,137 @@
...
+@Named("org.phenotips.pedigree.rest.internal.PedigreeConverterImpl")
+@Singleton
+public class PedigreeConverterImpl extends XWikiResource implements PedigreeConverter{
+    @Inject
+    private Logger logger;
+
+    @Override
+    public Response convertPedigree(String ped) {
+        if (ped == null) {
+            this.logger.error("PED file is empty");
```
allasm: I contemplated a bit about asking to replace `error()`-s with `debug()`-s, but I think this (and the other case below)  ARE errors worth logging

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/internal/PedigreeConverterImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                jsonObject.put("sex", "male");
+            }
+            else if(dataSplit[4].equals("2")) {
+                jsonObject.put("sex", "female");
+            }
+            else {
+                jsonObject.put("sex", "unknown");
+            }
+            if (dataSplit[5].equals("0") || dataSplit[5].equals("-9")) {
+                jsonObject.put("carrierStatus", "missing").put("disorders", new JSONArray().put("missing"));
```
allasm: "missing" is not a supported status, there is "unknown" instead

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/internal/PedigreeConverterImpl.java
```diff
@@ -0,0 +1,137 @@
...
+     *   Maternal ID`
+     *   Sex (1=male; 2=female; other=unknown)
+     *   Phenotype
+     *
+     *   Phenotype, by default, should be coded as:
+     *      -9 missing
+     *       0 missing
+     *       1 unaffected
+     *       2 affected
+     *       */
```
allasm: Ideally we should document what happens if data is "missing" - is it treated as "unaffected" or "unknown" in the JSON? We can also add a POST parameter option to specify how the data should be treated (good as an exercise as well)

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/PedigreeConverter.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @version $Id$
+ * @since 1.4m2
+ */
+// Fix annotations
+@Path("/pedigreeConverter")
```
allasm: I see Patient REST interfaces use a `ParentResource` annotation (e.g. `@ParentResource(RootResource.class)`). It works without it, and I'm not sure if it is a good practice to include it or not (I know we removed it while debugging another problem, maybe we should add it back?). @sdumitriu?

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/PedigreeConverter.java
AlexLapena: +1 to this,  these javadocs will be fixed.

components/pedigree/rest/src/main/java/org/phenotips/pedigree/rest/PedigreeConverter.java
AlexLapena: This must have come from my IDE auto import, I'll change that

components/pedigree/rest/pom.xml
veronikaslc: `mvn dependency:analyze` does the job, `mvn dependency:analyze-report` will even generate HTML page in `target` directory

PR #2383, opened 2017-07-29
PT-3328: Fail the build if there are dependency misconfigurations

sdumitriu:
Done
Also fixed existing dependency issues

### DIFF
--- a/components/application-configuration/api/pom.xml
+++ b/components/application-configuration/api/pom.xml
@@ -44,6 +44,10 @@
       <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-collections4</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.xwiki.platform</groupId>
       <artifactId>xwiki-platform-uiextension-api</artifactId>
       <version>${xwiki.version}</version>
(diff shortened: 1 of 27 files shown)


sashaandjic:
Tests failing. Was able to build with -Pquick option.
https://drive.google.com/open?id=1V9fb_1nFJAymwQGOvUxecavKShDdfTnIxQibninctB4

sdumitriu:
@sashaandjic Building with `-Pquick` defeats the purpose, this is exactly what's supposed to be tested, if the build passes with checks enabled.

The build error is unrelated to this.

sdumitriu:
Did you try building again? You can just resume from where it failed, as it says at the bottom:

`mvn clean install -rf :patient-access-rules-api`

sashaandjic:
@sdumitriu  Build passed.

allasm:
A clean `git pull` and build of the branch resulted in an error:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.0.0:analyze-only (check-dependencies) on project patient-access-rules-api: Dependency problems found -> [Help 1]
`

sdumitriu:
@allasm That's strange, `org.json` is not a declared dependency for that module in that branch. Did you accidentally switch to a different branch while building?

allasm:
OK, problem resolved. The issue was that I had a local `.java` file in `patient-access-rules-api` which was not in `git`, and stayed there for all git branches, and it used a `json` library which was no declared as a dependency. So the actual build would pass and file would go unused, but the dependency checker found the file and complained. So this pull request is ok.

PR #2385, opened 2017-08-01
PT-3330 and PT-3332: Extra files controllers

sdumitriu:
To test, first you must enable the Additional Files section from the administration.


components/export/ui/src/main/resources/PhenoTips/ExportPreferences.xml
```diff
@@ -49,7 +49,9 @@
 #set($customExporters = {
     'PhenoTips.GeneClass' : {'name' : 'genes', 'label': "$services.localization.render('phenotips.exportPreferences.field.genes')"},
     'PhenoTips.InvestigationClass' : {'name' : 'genes', 'label': "$services.localization.render('phenotips.exportPreferences.field.genes')"},
-    'PhenoTips.AllergiesDataClass' : {'name': 'allergies', 'label' : "$services.localization.render('PhenoTips.AllergiesDataClass_allergies')"}
+    'PhenoTips.AllergiesDataClass' : {'name': 'allergies', 'label' : "$services.localization.render('PhenoTips.AllergiesDataClass_allergies')"},
+    'PhenoTips.ExternalFileClass' : {'name': 'additionalDocuments', 'label' : "$services.localization.render('phenotips.exportPreferences.field.additionalDocuments')"},
+    'PhenoTips.ExternalImageClass' : {'name': 'additionalImages', 'label' : "$services.localization.render('phenotips.exportPreferences.field.additionalImages')"}
```
marta-: These last two shouldn't show up in the Excel export dialog.

sdumitriu: Fiiiine, I'll try to remove them.

sdumitriu: Done.

marta-:
I finally tested pushing briefly and it worked for me as expected.

marta-:
The document size limitation will be addressed separately: [PT-3336](https://phenotips.atlassian.net/browse/PT-3336). The actual controllers work as expected and the data is serialized/deserialized correctly.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/AdditionalImagesController.java
```diff
@@ -0,0 +1,204 @@
...
+            for (Attachment image : images) {
+                BaseObject xobject = doc.newXObject(CLASS_REFERENCE, this.contextProvider.get());
+                XWikiAttachment xattachment = doc.getAttachment(image.getFilename());
+                if (xattachment == null) {
+                    xattachment = new XWikiAttachment(doc, image.getFilename());
+                    doc.addAttachment(xattachment);
+                }
+                xattachment.setContent(image.getContent());
+                DocumentReference author = image.getAuthorReference();
+                if (author != null
```
allasm: just co confirm, the logic here is to set image author to current user if there is any author, and leave no author if there is none?

sdumitriu: No:
- if there is no author, then let the imported image also have no author (which is considered as Guest)
- if there is an author specified, and that author exists on the local instance, then use the same author
- if there is an author specified, but it doesn't exist in the local instance, then set the current author

allasm: Ah, right, missed one of the clauses in the condition

PR #2386, opened 2017-08-01
PT-3286: When hovering an individual in the pedigree, highlight all its abnormalities/annotations in the legend

allasm:
Done. @veronikaslc please review, @sashaandjic please test

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/editor.css
+++ b/components/pedigree/resources/src/main/resources/pedigree/editor.css
@@ -303,7 +303,10 @@ input[type="radio"] {
   padding-left: 1px;
 }
 li.abnormality:hover {
-  background-color: #FFF9EE;
+  background-color: #FBF5EA;
+}
+.abnormality.highlighted {
+  background-color: #FBF5EA;
 }
 .legend-box .abnormality-list {
   list-style-type: none;
--- a/components/pedigree/resources/src/main/resources/pedigree/pedigree.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/pedigree.js
@@ -621,6 +621,16 @@ define([
         },
 
         /**
+         * Returns a list of all available legends
+         */
+        getAllLegends: function() {
+            return [ this.getCancerLegend(),
+                     this.getHPOLegend(),
+                     this.getDisorderLegend(),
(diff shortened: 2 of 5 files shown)


components/pedigree/resources/src/main/resources/pedigree/editor.css
```diff
@@ -305,6 +305,9 @@ input[type="radio"] {
 li.abnormality:hover {
   background-color: #FFF9EE;
 }
+.abnormality.highlighted {
+  background-color: #FBF5EA;
+}
```
marta-: IMO both `:hover` and `.highlighted` should have the same color. If you think the new one is better just use it for both.

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
```diff
@@ -621,6 +621,16 @@ define([
...
 
         /**
+         * Returns a list of all available legends
+         */
+        getAllLegends: function() {
+            return [ this.getCancerLegend(),
+                     this.getHPOLegend(),
+                     this.getDisorderLegend(),
+                     this.getCarrierGeneLegend(), this.getRejectedGeneLegend(), this.getCandidateGeneLegend(), this.getCausalGeneLegend() ];
+        },
```
marta-: There has to be a better way... This is not scalable. Different legends should somehow register when instatiated.

components/pedigree/resources/src/main/resources/pedigree/editor.css
allasm: agree, will update now

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
allasm: agree, but this is a larger fix than this issue name implies, I will create a new issue for that and think a bit how to do it right instead of doing something better but still a quick-fix-type (may still make it to m1, but will assign m2 for now). The change here at the very least makes sure all the _other_ code only uses this one method instead of manually calling 5 legends (as it used to, though only in one place - but now there are more places)

allasm:
Addressed code reviews

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
marta-: Sure, it should be a separate thing. That's why I didn't request changes and only left a comment.

PR #2388, opened 2017-08-02
PT-3319: Long custom entries in the genes table cause layout issues

marta-:
Better fix.

### DIFF
--- a/components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm
+++ b/components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm
@@ -158,7 +158,7 @@ $doc.display($prop.getName(), 'view', $fakeObj )##
 #set($variantObjects = $doc.getObjects($variantClass.name, "gene", $geneObj.getProperty('gene').value))
 |(% class="row-count" %)$geneNumber##
 #foreach($prop in $geneObj.xWikiClass.properties)##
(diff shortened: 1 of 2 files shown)


components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm
```diff
@@ -158,7 +158,7 @@ $doc.display($prop.getName(), 'view', $fakeObj )##
 #set($variantObjects = $doc.getObjects($variantClass.name, "gene", $geneObj.getProperty('gene').value))
 |(% class="row-count" %)$geneNumber##
 #foreach($prop in $geneObj.xWikiClass.properties)##
-|(% class="$prop.translatedPrettyName gene #if ($prop.name == 'gene')mandatory#end gene-$geneObjNumber"##
+|(% class="$prop.translatedPrettyName gene #if ($prop.name == 'gene')mandatory#if($options.mode != 'edit' && !$services.vocabularies.hgnc.getTerm($geneObj.getProperty($prop.name).value)) custom-entry#end#end gene-$geneObjNumber"##
```
sdumitriu: `$geneObj.getProperty($prop.name).value` can be written as `$geneObj.getValue($prop.name)`.

marta-:
@sdumitriu Changed as you suggested.

PR #2392, opened 2017-08-08
[refactoring] PT-3296: Create interface for ComputedMeasurements

katpavlova:
Authored with @marta- and @sdumitriu

### DIFF
--- a/components/patient-measurements/api/pom.xml
+++ b/components/patient-measurements/api/pom.xml
@@ -78,6 +78,10 @@
       <version>${xwiki.version}</version>
     </dependency>
     <dependency>
+      <groupId>javax.ws.rs</groupId>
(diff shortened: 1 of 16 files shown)


katpavlova:
I could generalize the handleComputation method by passing it both the map of params and the list of computational dependencies, and having it check each param listed in the dependencies for being non-null and parsable. Then it would pass an array containing all of the parsed dependencies pulled from the map to the compute method, which would use those dependencies to compute the value.

This would result in less code duplication, but passing an array like that sounds messy. I could pass a map, but then we would need initialize and fetch values from two different hashmaps containing similar information. Also, I would need to place the handleComputation method in the AbstractMeasurementHandler superclass rather than the ComputedMeasurementHandler interface, where it should ideally be. What do you think @marta- ?

PR #2394, opened 2017-08-10
PT-3343: 'Owner' and 'visibility' should be available for displaying in the patients table

marta-:


sashaandjic:
@marta Column Owner cannot be sorted in DESC order.
screen shot 2017-08-10 at 10 05 01 pm

sashaandjic:
More details and problems: when attempting to do a DESC order by "Owner", it seems the grid gets sorted by "Reported by" column in DESC order. ASC order appears to be working, however some patient records show twice in the grid  (see P0000033 and P0000034 on rows: 1,6,7,8 in the screenshot)
33 and 34

marta-:
The problem reported above is unrelated to PT-3343. Created [PT-3345](https://phenotips.atlassian.net/browse/PT-3345).

PR #2396, opened 2017-08-11
PT-3344: Collaborators should be available for displaying in the pati…

marta-:
…ents table, and filterable in the advanced filters


sashaandjic:
Problem found, steps to reproduce:
- filter by Mary as owner
- filter by Peter as collaborator
- open one of the patient records and go back
- Peter is listed twice as collaborator in the filtering panel
screen shot 2017-08-14 at 2 46 57 pm

marta-:
@sashaandjic this is https://phenotips.atlassian.net/browse/PT-2822 . Happens to all filters with multiple values, so it's not directly related to PT-3344.

sashaandjic:
It would be nice if we can enter groups in the owner, referer, collaborator,... filter fileds

marta-:
@sashaandjic agreed. already looked into this, there's a service for suggesting both users and workgroups simultaneously deep inside itai's Projects&Templates implementation. Somebody needs to get it out of there and make it usable separately. Can you create a task?

PR #2397, opened 2017-08-14
PT-3346: Filter panel should remain collapsed until user either opens it or logs out of PhenoTips

marta-:


components/navigation/ui/src/main/resources/PhenoTips/DBWebHomeSheet.xml
```diff
@@ -251,25 +251,44 @@ document.observe('xwiki:livetable:newrow', function(tableEvent){
...
+    var toggleFilters = function (filtersElt, forceHide) {
+      if (filtersElt) {
+        filtersElt.toggleClassName('collapsed', forceHide);
+        filtersElt.up('.xwiki-livetable-container').toggleClassName('hidden-filters', forceHide);
+        var key = getCookieKey(filtersElt.up('.entity-directory').id);
+        if (filtersElt.hasClassName('collapsed')) {
+          XWiki.cookies.create(key, 'hidden', '')
+        } else {
+          XWiki.cookies.erase(key)
+       }
```
veronikaslc: couple of very minor comment:  indentation is missing one space; eye got used to `;` at the end of a line although they aren't necessary in those `if-else` one lines

veronikaslc: other than that looks fine to me

marta-: @veronikaslc thanks, fixed.

components/navigation/ui/src/main/resources/PhenoTips/DBWebHomeSheet.xml
```diff
@@ -251,25 +251,44 @@ document.observe('xwiki:livetable:newrow', function(tableEvent){
       </use>
     </class>
     <property>
-      <cache>long</cache>
+      <cache>forbid</cache>
```
sdumitriu: I don't like this.

components/navigation/ui/src/main/resources/PhenoTips/DBWebHomeSheet.xml
```diff
@@ -251,25 +251,44 @@ document.observe('xwiki:livetable:newrow', function(tableEvent){
...
-    if (filtersElt) {
-      filtersElt.toggleClassName('collapsed');
-      filtersElt.up('.xwiki-livetable-container').toggleClassName('hidden-filters');
-    }
-  });
-});</code>
+      <code>(function() {
+  var init = function(event) {
+    var getCookieKey = function (tableId) {
+      return "$!{xcontext.userReference.hashCode()}" + '_' + tableId + '_filters_state';
```
sdumitriu: Better place the user name somewhere in the HTML.

marta-:
@sdumitriu comments addressed.

@sashaandjic @veronikaslc this needs to be reviewed and tested again. Thanks.

veronikaslc:
tested, works

sashaandjic:
Behaviour changed a bit...Previously we had 
Filter panels should on browse all records pages were be open by default when user logs in. Once a specific filter panel is collapsed, it remained collapsed unless:
1. user opens it again
2. user logs out and logs in again
Now #2 is no longer valid.

marta-:
Thanks @sashaandjic , it was a bug and I committed the fix.

sashaandjic:
There is still discrepancy between present behaviour and JIRA.
Currently the state of filters is saved per user, and whenever user logs in the filters will look the same as in the last session.
In JIRA rask it was specified that every time someone logs in, the filters should be open.

marta-:
@sashaandjic That was by design. From my experience it makes more sense to find the filters the way you left them when you log back in, than to have everything reset to defaults. I should have changed the jira task to reflect this.

sashaandjic:
Ok. I will adjust task description in JIRA. Thank you.

PR #2402, opened 2017-08-15
PT-3353: Terms added from the ontology browser are always added in the main clinical symptoms section, regardless of where the browser was triggered from

marta-:


marta-:
@sdumitriu Fixed the first issue.

marta-:
I suggest we leave the second issue for later and report it as a separate bug. It's pretty unlikely to happen and when it does there are simple ways around it.

PR #2403, opened 2017-08-15
PT-3351: REST Service for suggesting both users and workgroups simultaneously

veronikaslc:


components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
```diff
@@ -0,0 +1,249 @@
...
+        UsersAndGroups.groupsQueryString = groupsQuerySb.toString();
+    }
+
+    /**
+     * Checks whether an {@link userOrGroup} is an entity that represents a user or a group.
+     *
+     * @param userOrGroup entity to check
+     * @return {@link USER}, @{link GROUP}, or {@link UNKNOWN}
+     */
+    public String getType(EntityReference userOrGroup)
```
allasm: Just curious when this method may be useful and if return type of `String` is the best choice - e.g. maybe for users of the class `isUser()` and `isGroup()` may be more useful without the constants?

components/users/rest/src/main/java/org/phenotips/users/rest/UsersResource.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+/**
+ * Root resource for working with users and groups of users.
+ *
+ * @version $Id$
+ * @since 1.4
+ */
+@Path("/users")
```
marta-: I don't like the API.
* First, `users` is inappropriate. According to @sdumitriu , the correct term that captures both users and groups is `principals`.
* Second, it's inconsistent with our other services and it doesn't follow best practices. A better path is `principals/suggest`. Moreover, the root `/principals` should list all users and groups, and `/principals/{id}` should provide information about that particular principal, but this is out of the scope of this task for now.
* Both groups and users should be suggested by default.

components/users/rest/src/main/java/org/phenotips/users/rest/UsersResource.java
```diff
@@ -0,0 +1,48 @@
...
+     * Searches for users and/or groups matching the input parameter. Result is returned as JSON
+     *
+     * @param input string to look for
+     * @param searchUsers if true, includes users in result
+     * @param searchGroups if true, includes groups in result
+     * @return a json object containing all results found
+     */
+    @GET
+    Response searchUsersAndGroups(
+        @QueryParam("input") @DefaultValue("0") String input,
```
marta-: I think `@DefaultValue` belongs to a different parameter, since the input is not optional, and it's not a number. Perhaps a `maxResults` with a default value of `10`?

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
marta-: In general `is____` is less extensible, but I don't think we need that much extensibility here. I'd say replacing `getType()` with `isUser` & `isGroup` makes the code friendlier. @sdumitriu Thoughts?

sdumitriu: +1 for `isUser` and `isGroup`.

allasm: my reasoning was that since the return value is not directly usable (it is a new constant, so can't just use return type as a parameter to other code) the code that uses this would have to have a pattern like:

```var type = getType()
if (type == A) ... else if (type == B) .... ```
(or `if (getType() == A) ... else if (getType() == B) ...` which is less efficient)

which can be replaced by more obvious non-custom constant `if (isUser()) ... else if (isGroup()) ...`

File: components/users/rest/src/main/java/org/phenotips/users/rest/UsersResource.java (hunk 3 above)
Comment (veronikaslc) [inline components/users/rest/src/main/java/org/phenotips/users/rest/UsersResource.java]: misplaces here, I will add `maxResults`, not sure if I should add `sort` though

**Comment by veronikaslc** [comment]
@marta- codereview comments addressed

File: components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java [hunk 4]
```diff
@@ -0,0 +1,304 @@
...
+@Singleton
+public class UsersAndGroups
+{
+    private static final EntityReference USER_CLASS = new EntityReference("XWikiUsers", EntityType.DOCUMENT,
+        Constants.XWIKI_SPACE_REFERENCE);
+
+    private static final EntityReference GROUP_CLASS = new EntityReference("PhenoTipsGroupClass", EntityType.DOCUMENT,
+        Constants.CODE_SPACE_REFERENCE);
+
+    private static final String USER = "user";
```
allasm: it seems `USER` and `GROUP` constants are not used any more

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
```diff
@@ -0,0 +1,304 @@
...
+                XWikiContext xcontext = this.xcontextProvider.get();
+                XWiki xwiki = xcontext.getWiki();
+                avatarURL = xwiki.getSkinFile("icons/xwiki/noavatar.png", xcontext);
+            }
+            String name = user.getName();
+            if (StringUtils.isBlank(name)) {
+                name = user.getUsername();
+            }
+            JSONObject o = createObject(user.getProfileDocument().toString(), name, avatarURL, USER);
+            resultArray.put(o);
```
allasm: This is not very important, but if `resultArray` already has `maxResults` entries it will end up with `maxResults + 1` entries. Making this check before the loop iteration may actually save some work (if there are `maxResults` users any group search will be skipped when this method is called for the second time for groups)

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
```diff
@@ -0,0 +1,304 @@
...
+            String avatarURL = "";
+            if (attachmentRefs.size() > 0) {
+                avatarURL = this.bridge.getAttachmentURL(attachmentRefs.get(0), true);
+            } else {
+                XWikiContext xcontext = this.xcontextProvider.get();
+                XWiki xwiki = xcontext.getWiki();
+                avatarURL = xwiki.getSkinFile("icons/xwiki/noavatargroup.png", xcontext);
+            }
+            JSONObject o = createObject(groupName, group.getReference().getName(), avatarURL, GROUP);
+            resultArray.put(o);
```
allasm: This is not very important, but if `resultArray` already has `maxResults` entries it will end up with `maxResults + 1` entries. Making this check at the start of the loop iteration may actually save some work (if there are `maxResults` users then group search will be skipped completely when this method is called)

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
```diff
@@ -0,0 +1,304 @@
...
+            if (resultArray.length() == maxResults) {
+                break;
+            }
+        }
+    }
+
+    private void runGroupsQuery(JSONArray resultArray, String queryString, String formattedInput, int maxResults)
+        throws Exception
+    {
+        List<String> queryResult = runQuery(queryString, formattedInput, maxResults);
```
allasm: can also optimize this as `maxResults - resultArray.length()`

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
veronikaslc: they are used

veronikaslc:
@marta- @allasm codereview comments addressed

components/users/api/src/main/java/org/phenotips/groups/internal/UsersAndGroups.java
```diff
@@ -0,0 +1,307 @@
...
+        return queryResults;
+    }
+
+    private JSONObject createObject(String id, String value, String avatar, String type)
+    {
+        JSONObject o = new JSONObject();
+
+        StringBuilder idWithType = new StringBuilder();
+        idWithType.append(id).append(";").append(type);
+        o.put(UsersAndGroups.ID_KEY, idWithType.toString());
```
allasm: @marta- just curious why such a strange ID format ("ID;type")?

marta-: No idea, I didn't get this far with the review. Now that you mention it, it does look wrong. I assume it has something to do with how the projects&templates code is using this service (reminder that this is where @veronikaslc extracted the code from).

veronikaslc: I'll remove `;type` appending

sdumitriu:
Jenkins, test this please.

veronikaslc:
Jenkins, test this please.

veronikaslc:
@sashaandjic tests are fixed, ready for testing :)

sashaandjic:
@veronikaslc I'm not getting any groups, see the screenshots
screen shot 2017-12-06 at 12 58 40 pm
screen shot 2017-12-06 at 12 59 29 pm

screen shot 2017-12-06 at 1 00 57 pm

veronikaslc:
Seems that old version of PR without last commit was tested. @sashaandjic is rebuilding, meanwhile I will update one test. (@sdumitriu I discovered interesting thing: `group` is a reserved word in xwiki xwql queries. When I run `from doc.object(PhenoTips.PhenoTipsGroupClass) as group..` stuff breaks at query translation moment with error `id is expected`)

sashaandjic:
tested. all good.

PR #2407, opened 2017-08-16
PT-3357: In pedigree editor do not show A&W options on hoverboxes for patients which are read-only for current user

allasm:
@sashaandjic please test


allasm:
Box sizes should be adjusted in a few other cases ad well (e.g. nodes assigned to PT patients where the box overlaps PT id somewhat). We can/should rethink all sizes in a separate issue

marta-:
Sounds good.

PR #2411, opened 2017-08-17
PT-3292: Migrating from cancer names to HPO identifiers

a-gvo:


allasm:
@sashaandjic, did you test this?

sashaandjic:
no. how do i test it?

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71504PhenoTips3292DataMigration.java
```diff
@@ -0,0 +1,426 @@
...
+
+            if (xDocument != null) {
+                this.logger.debug("Updating pedigree for document {}.", docName);
+                final BaseObject pedigreeXObject = xDocument.getXObject(PEDIGREE_CLASS_REFERENCE);
+                if (pedigreeXObject != null) {
+                    updatePedigree(xDocument, pedigreeXObject, context, session);
+                }
+            }
+        } catch (final JSONException ex) {
+            this.logger.error("Error updating JSON for document {} : [{}]", docName, ex.getMessage());
```
allasm: In theory this may be caused by reading existing JSON (not writing new stuff), so the error message is a bit misleading - though the end result is the same, the data does not get updated, so I think this is OK

allasm:
@sashaandjic: testing instructions:

1) get the branch as usual
2) `git reset aa6dd41bc8f896c0311cca5d64588269e1fee90 --hard`
3) build, install, create some pedigrees with cancers
4) `git pull`
5) build, install, copy DB from the installation installed in step #3
6) test

sashaandjic:
i'm on step 5

sashaandjic:
tested

PR #2415, opened 2017-08-24
PT-3368 Extend Entities API

allasm:
Seems to work fine, though see some comments below about some design issues


allasm:
A few things can be improved, though if we treat the API as `@unstable` that can probably be done later:

1) `AbstractExternalPrimaryEntityGroupManager.java` has no built-in support for uniqueness, e.g. that a `Patient` is only in one `Family`. I think it is best to address this when refactoring Families to use Entities, since the requirements and quirks will be more clear at that point

2) Entities-with-parameters are only supported in `AbstractInternalPrimaryEntityGroupManager.java` and the support is non-straightforward (need to overwrite non-obvious `setMemberParameters` and also `getMembers()`, e.g. see [https://github.com/phenotips/phenotips/pull/2413/files#diff-42c1075c49afc4ea4f492e427583cdfcR76](CollaboratorsInProjectManager) ) - as described [https://github.com/phenotips/phenotips/pull/2415/files#diff-6f0d303e9be452f6273331bbb1c6ec07R211](here) 

This can probably be refactored to be more obvious, e.g. a new internal-with-aprameters manager, which has an explicit `add(member, parameter(s) )` method and an obvious callback to implement to assemble a member back

3) I didn ot address the problem @danielpgross described [https://github.com/phenotips/phenotips/pull/2115/files#diff-3b798f34b13cdc4e8c590145f7f9f36aR88](here). Since projects that use all of the functionality here work fine, I suspect @danielpgross 's code could have been modified with appropriate annotations to make it work, e.g. [https://github.com/phenotips/phenotips/pull/2413/files#diff-6438e89c5a183f3290afa7c155499271R37](this) and [https://github.com/phenotips/phenotips/pull/2413/files#diff-05f0dd7e0e53b6665d780e20a47f7667R37](this) seem to work, and able to find the necessary managers.

sdumitriu:
See other comments on #2115

sdumitriu:
Redone in #2617

PR #2416, opened 2017-08-25
PT-3341: Allow administrators to permanently redefine the location of vocabulary sources

veronikaslc:


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractSolrVocabulary.java
```diff
@@ -241,6 +245,18 @@ public long getDistance(VocabularyTerm fromTerm, VocabularyTerm toTerm)
...
+    @Override
+    public String getSourceLocation()
+    {
+        return this.relocationService.getRelocation(this.getDefaultSourceLocation());
+    }
+
+    @Override
+    public List<VocabularyExtension> getExtensions()
+    {
+        return this.extensions.get();
```
sdumitriu: Does this return all extensions, or just those valid for this vocabulary? It should be the latter.

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/GeneForPhenotypeAnnotationsExtensionTest.java
```diff
@@ -119,7 +119,7 @@ public void getTargetVocabularyIds()
     @Test
     public void getAnnotationSource()
     {
-        Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
+        // Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
```
sdumitriu: This isn't good.

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -122,7 +122,7 @@ public void getTargetVocabularyIds() throws Exception
     @Test
     public void getAnnotationSource() throws Exception
     {
-        Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
+        // Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
     }
```
sdumitriu: This isn't good.

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/PositivePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -122,7 +122,7 @@ public void getTargetVocabularyIds() throws Exception
     @Test
     public void getAnnotationSource() throws Exception
     {
-        Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
+        // Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
```
sdumitriu: This isn't good.

components/vocabularies/rest/src/main/java/org/phenotips/vocabularies/rest/internal/DefaultVocabularyResource.java
```diff
@@ -128,10 +130,18 @@ private Autolinker getCategoryLinks()
     @Override
     public Response reindex(String vocabularyId, String url)
     {
+        // Validate URL before loading any extensions
```
sdumitriu: This has already been committed, does it need to be rebased?

veronikaslc:
rebased, codereview comments addressed

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractSolrVocabulary.java
```diff
@@ -254,7 +254,13 @@ public String getSourceLocation()
     @Override
```
allasm: Please add at least a brief comment with each commit (e.g. "addressing codereview comments")

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -122,6 +158,17 @@ public void getTargetVocabularyIds() throws Exception
...
+        when(xcontextProvider.get()).thenReturn(context);
+        when(context.getWiki()).thenReturn(this.xwiki);
+        when(this.xwiki.getDocument(any(DocumentReference.class), any(XWikiContext.class)))
+            .thenReturn(this.document);
+        when(this.document.isNew()).thenReturn(false);
+        when(this.document.getXObject(VOCABULARY_RELOCATION_CLASS))
+            .thenReturn(this.object);
+        String source = this.serv.getRelocation(ANNOTATION_SOURCE);
+        Assert.assertEquals(ANNOTATION_SOURCE, source);
         // Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
```
allasm: I think Sergiu would ask you to remove the commented out code

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -83,6 +105,17 @@
...
     private Vocabulary vocabulary;
 
+    @Mock
+    private XWiki xwiki;
+
+    @Mock
+    private XWikiDocument document;
+
+    @Mock
+    private BaseObject object;
```
allasm: I would rename this to relocationObject, since it is only used as such. It is harder to understand the code when there are generic names like `object` around

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -83,6 +105,17 @@
...
+
+    private VocabularySourceRelocationService serv;
```
allasm: ...and similarly it is easier to understand whats going on if this is a `relocationServer` (or something similar), not a generic "serv"\"server"

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -122,6 +158,17 @@ public void getTargetVocabularyIds() throws Exception
     @Test
     public void getAnnotationSource() throws Exception
     {
+        final Provider<XWikiContext> xcontextProvider = this.mocker2.getInstance(XWikiContext.TYPE_PROVIDER);
```
allasm: I'm not even sure this test belongs here now, since relocation service should have tests of its own, and this test has nothing to do with `NegativePhenotypeForDiseaseAnnotationsExtension`. So I would just remove it altogether

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/PositivePhenotypeForDiseaseAnnotationsExtensionTest.java
```diff
@@ -122,7 +157,17 @@ public void getTargetVocabularyIds() throws Exception
     @Test
     public void getAnnotationSource() throws Exception
     {
-        // Assert.assertEquals(ANNOTATION_SOURCE, this.extension.getAnnotationSource());
+        final Provider<XWikiContext> xcontextProvider = this.mocker2.getInstance(XWikiContext.TYPE_PROVIDER);
```
allasm: Same comments here - I think this test should be removed from here (and variables named better if it stays)

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
allasm: Update: I'm 100% sure this test should be removed. `getAnnotationSource()` method was removed from `AbstractCSVAnnotationsExtension`, so the subject of this test - that method - is no longer here, and no reason for this test to exist.

components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/DefaultVocabularySourceRelocationService.java
```diff
@@ -0,0 +1,86 @@
...
+        new EntityReference("XWikiPreferences", EntityType.DOCUMENT, Constants.XWIKI_SPACE_REFERENCE);
+
+    @Inject
+    private Provider<XWikiContext> provider;
+
+    @Inject
+    private Logger logger;
+
+    @Override
+    public String getRelocation(String original)
```
allasm: @sdumitriul I discussed this with @veronikaslc and she said you should know better: why so we pass URLs around in the new design? I'e. when a vocabulary needs to be reindexed, don't we want to reindex "vocabulary named X" and not "vocabulary which used to be indexed from URL http://x.y.z"? It seems in this design we always need to remember the old URL, instead of just new URL for each vocabulary type?

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
veronikaslc: may be it should be relocated :)

sashaandjic:
- When I edit the source and hit enter the reindexing starts. I believe it would be better to reindex only after user hits button START REINDEXING.
- I edited the source for my HPO, ORPHANET and THE DISEASE ANNOTATIONS WITH HPO. Where did my REINDEX button go? Even after restoring to the default value the button did not appear.
screen shot 2017-10-11 at 12 09 55 am
- I should not be able to start reindexing in this situation (invalid file name) 
screen shot 2017-10-11 at 12 07 18 am
- ORPHANET buttons are misplaced

veronikaslc:
@sashaandjic  reindexing starts only if you click the button START REINDEXING. When you click the checkmark button - it only saves the new URL. Also I do not observe the other things. I have just tested a fresh build.
@allasm can you please test a build as well?

sashaandjic:
@veronikaslc I was not talking about checkmark. When you edit source of HPO for example, and hit enter the reindexing starts immediately. Also HPO is missing the START REINDEXING button.

components/vocabularies/hpo/annotations/src/test/java/org/phenotips/vocabulary/internal/hpoannotations/NegativePhenotypeForDiseaseAnnotationsExtensionTest.java
veronikaslc: Actually `getAnnotationSource()` method was not removed completely, it was moved up from `AbstractCSVAnnotationsExtension` to `VocabularyExtension.java` interface. But I will remove all tests if you think they are unnecessary.

veronikaslc:
@sashaandjic codereview comments addressed. I disabled reindex button until user saved new location URL. Also enter button disabled as well.

sashaandjic:
Not looking good. I think we should replace all "ONTOLOGY)" titles with "The URL of a file containing the vocabulary you wish to index". + optional "in .xxx format")
screen shot 2017-10-12 at 10 50 47 am

sashaandjic:
Why CHEBI has "START **INDEXING**" not "START REINDEXING" like others?

veronikaslc:
CHEBI has "START INDEXING" because it has no version, thus, was never indexed.

veronikaslc:
IMHO, saving then indexing is more clear workflow. Editing URL is one step, indexing is another, mixing them is not good UX experience, it leave user with guesses and assumptions.

veronikaslc:
If we will replace "ONTOLOGY)" titles with "The URL of a file containing the vocabulary you wish to index", we need to replace all other titles with he same style. For example, "HPO ANNOTATION WITH ASSOCIATED GENES" should be replaces with "The URL of a file containing the HPO ANNOTATION WITH ASSOCIATED GENES". They all are URLs. Otherwise titles will look inconsistent. But then the part "The URL of a file containing the" will be repeated in each of them which is annoying. IMHO, the short and nice way is just to say "ONTOLOGY" above the ontologu URL.

veronikaslc:
Rebased on latest  master and squashed last 3 codereview commits into one.

allasm:
Jenkins, build this please

PR #2420, opened 2017-08-28
PT-3396: Improve the usefulness of the data and links in the homepage gadgets related to patients and families

marta-:


marta-:
1/ Right, forgot about those, will add them.

2/ Yes. If the user cannot delete, it simply wob't show up. If the user has manager level, it would be useful to have the action link available.

3/ I think the link on the id makes more sense. We can add the view action if we ever end up removing the id column and there's nothing to click, but for now IMO it would just unnecessarily crowd the table.

marta-:
@sdumitriu Comment 1 addressed in the last commit.

PR #2427, opened 2017-08-31
PT-3359: Replaced "Patient" with "PrimaryEntity" in the permissions module

a-gvo:


components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/internal/DefaultPermissionsManager.java
```diff
@@ -0,0 +1,157 @@
...
+        @Nullable final Visibility requiredVisibility)
+    {
+        return this.visibilityHelper.filterByVisibility(entities, requiredVisibility);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Iterator<Patient> filterByVisibility(Iterator<Patient> patients, Visibility requiredVisibility)
+    {
+        return (Iterator<Patient>) (Iterator<?>) filterVisible((Iterator<PrimaryEntity>) (Iterator<?>) patients,
```
allasm: I'm wondering if there is a nicer way to cast thing around, e.g. declare `filterVisible()` as accepting `Iterator<? extends PrimaryEntity>` or something else (could not find anything but `? extends`, but maybe new Java 8 ways were still deep down in the list of results)

components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/PermissionsManager.java
```diff
@@ -0,0 +1,166 @@
...
+     *         {@link org.phenotips.data.permissions.internal.visibility.PrivateVisibility} if {@code name} is invalid
+     */
+    @Nonnull
+    Visibility resolveVisibility(@Nullable String name);
+
+    @Nonnull
+    Collection<AccessLevel> listAccessLevels();
+
+    @Nullable
+    AccessLevel resolveAccessLevel(@Nullable String name);
```
allasm: What i sit supposed to mean mean when a `null` is returned - is it "no access"?

allasm: OK, never mind, I found a comment in another class, probably should add it here:

`@return the {@link AccessLevel} associated with the provided {@code name}, or null if the provided {@code name}`
 +     * is not valid

components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/script/PermissionsManagerScriptService.java
```diff
@@ -135,7 +142,28 @@ public PatientAccess getPatientAccess(String targetPatientId)
...
+                // scripts have only access to a SecurePatient implementation of a Patient,
+        // which does not support all the functionality PatientAccess needs. So
+        // need to get the full Patient object here instead of taking it as an argument
+        //
+        // Since this is a script service, need to check access rights the same way SecurePatientReporistory does.
+        //
+        // TODO: rights management should be refactored so that less is done from velocity
+        //       and this method won't be needed any more
+
+        final PrimaryEntity entity = this.patientRepository.get(targetEntityId);
```
allasm: Can this be a problem? The input now is not a `patientId`, it is now a `targetEntityId`, so using `patientRepository` to get the object may be a mistake when entity is not a patient. Or if this is only used for patients this class should be renamed.

...as a side note it may be a very difficult task to convert an entity ID to an entity object for a generic unknown Entity type, @itaiGershtansky tried to solve this problem in this entities branch and it is hairy (and may be buggy, still under review)

components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/script/PermissionsManagerScriptService.java
```diff
@@ -112,12 +114,17 @@ public AccessLevel resolveAccessLevel(String name)
...
-    public PatientAccess getPatientAccess(Patient targetPatient)
+    public EntityAccess getPatientAccess(Patient targetPatient)
     {
         return this.getPatientAccess(targetPatient.getId());
     }
 
-    public PatientAccess getPatientAccess(String targetPatientId)
+    public EntityAccess getEntityAccess(@Nullable final PrimaryEntity targetEntity)
+    {
+        return targetEntity != null ? getEntityAccess(targetEntity.getId()) : null;
```
allasm: This seems very inefficient, to convert an entity to an ID, and then back to entity in `getEntityAccess(String id)`. I think this should be implemented the other way around, i.e. `getEntityAccess(String id)` should resolve the ID to an object, and then cal _this_ method instead to perform actual work.

allasm:
I reviewed the pull request, but it is very large and I may have missed something, so need to be tested very well.

sashaandjic:
tested and haven't found any issues

components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/script/PermissionsManagerScriptService.java
allasm: Update: after thinking for a little I think we need to have a central "repository of repositories" (RoR), and whenever we use an entity method with an ID we should also be passing the entity type, which will be used to get the correct repository from that RoR. This can be done as part of Entity refactoring, for now this IS a bug, but since in practice is only used for patients we can let this through if we really need this in this milestone

marta-: An automatic entity type detection is much more useful in practice than requiring to specify the entity type.

allasm: It is true, but do we want to do it? This would require figuring out that a "Pxxxxx" is a patient and "FAMxxx" is a family etc., which is doable, but requires a very strong scheme for entity IDs (since security may depend on this)

marta-: We need to think more from the perspective of those using the API, not those implementing the API. The whole point of the entities architecture was to write less code.

For example, if there's a component implementing a feature for more than one type of entity (in our immediate case, permissions), I shouldn't have to figure out which repository to ask for that entity. I have the entity id, i should get the entity (whatever it is) and process permissions for it. Right now it has to be the usual, ugly "is this a patient? if not, is this a family?"

PR #2428, opened 2017-08-31
PT-3380: Make families basic entities

a-gvo:


components/family-studies/api/src/main/java/org/phenotips/studies/family/Family.java
```diff
@@ -38,7 +39,7 @@
  * @version $Id$
  * @since 1.4
  */
-public interface Family
+public interface Family extends PrimaryEntity
```
allasm: I think to make it more clear what `Family` is about we should remove all duplicate (available in `PrimaryEntity`) methods from this interface since they are inherited (`getId()`, `getXDocument()`, `getDocumentReference()`, `toJSON()`)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
```diff
@@ -87,25 +87,14 @@
      */
     public PhenotipsFamily(XWikiDocument familyDocument)
     {
+        super(familyDocument);
```
allasm: i think we can get rid of `this.familyDocument` as well and just use the one inherited from `AbstractPrimaryEntity`

components/family-studies/api/src/main/java/org/phenotips/studies/family/Family.java
a-gvo: Right, fixed.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
a-gvo: Fixed.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
```diff
@@ -133,22 +119,16 @@ public DocumentReference getDocumentReference()
...
         List<String> memberIds = this.getMembersIds();
-        List<Patient> members = new ArrayList<>(memberIds.size());
-        for (String memberId : memberIds) {
-            Patient patient = PhenotipsFamily.patientRepository.get(memberId);
-            members.add(patient);
-        }
-        return members;
+        return memberIds.stream()
+            .map(memberId -> PhenotipsFamily.patientRepository.get(memberId))
+            .collect(Collectors.toCollection(() -> new ArrayList<>(memberIds.size())));
```
allasm: This probably should be a subject of a separate discussion, but just curious why rewrite an already written `for` loop in favor of streams - from what I read about streams, the main benefit is ability to parallelize them with `parallel()`, which is not very relevant here, otherwise old school `for` loops usually perform faster, put less pressure on garbage collector due to less wrapper objects, and are easier to debug (this is why I guess there used to be an unnecessary temporary variable `patient` in the original code), and the code is about equally easy to read in both cases. Streams and maps and lambdas are useful/cool, but why use them in such a simple case?

marta-: @allasm my guess would be: fewer lines of code :)

PR #2432, opened 2017-09-07
PT-3382: Refactored the Permissions widget in preparation for supporting family entities

marta-:


sashaandjic:
not sure how to test this one

sdumitriu:
Just test that the permissions UI still looks and works fine, both for patients and families.

PR #2436, opened 2017-09-11
PT-3421: Change "Unborn" to "Pregnancy" in pedigree editor person menu

marta-:


sashaandjic:
doesn't look neat, perhaps we change "A pregnancy" to "Pregnancy"
screen shot 2017-09-11 at 10 49 24 pm

allasm:
We can also try to make the dialogue wider. I recall it used to be not very neat when very wide as well, but things have changed since last time I played with it, maybe it will look good (note that in some browsers/zoom levels text will scale slightly differently, so if it fits with a very small margin it may not work on some other screen/browser combination)

marta-:
@allasm how about 2 colums instead of 3?

marta-:
Closing and postponing. The layout needs to change.

PR #2437, opened 2017-09-12
PT-2804, PT-2824 and PT-3419: Show the correct families in "Families shared with me" and "All families"

sdumitriu:


sashaandjic:
admin
johndoe

marta-:
@sashaandjic 
"As an admin I should see everyone": I disagree. The purpose of these boxes is (as their title suggests) to provide quick access to data specifically shared with that user. There's the "browse all" link for accessing all the available data. We can discuss changing this behavior, but 1/ it should be part of a separate task and 2/ it should involve changing the box titles as well. Please note that "patients shared with me" always behaved like this.

"FAM0000001 is shared with JohnDoe": i see in your screenshots that the owner is also JohnDoe. Does it really make sense to show that family in both boxes? We consider ownership prioritary.

sashaandjic:
@marta-  Sorry I wasn't clear... Patients in FAM0000001 are owned by the group Clinical Genetics (MaryGreen & PeterBrown) and I'm assuming the owner of the family should be the same?
Also, for an Admin, families shared with me only shows families whose patients are owned by Clinical Genetics.

marta-:
The owner of a family is (for now, until we implement ownership transfer for families) the user who created that family.

The owner of FAM0000001 is displayed in your 1st screenshot: it's JohnDoe. You can also check this by viewing the family document with `?xpage=xml` and looking at the owner object.

Are there any explicitely shared families that are missing for Admin? More precisely, is there any family that:
* is not owned by Admin
* has a member (patient) where Admin or a group containing Admin is either owner or collaborator

but does not appear in shared families?

sdumitriu:
> Also, for an Admin, families shared with me only shows families whose patients are owned by Clinical Genetics.

This is normal. No patient is explicitly shared with Admin, but Admin is a member of Clinical Genetics Administrators.

sashaandjic:
Ok.

sashaandjic:
Tested. I didn't find any problem.

marta-:
Thanks @sashaandjic.  Added the missing translation key for family owner in the last commit.

PR #2438, opened 2017-09-13
PT-3429: Do not propagate click events beyond the modal popup

marta-:
How to test:
- Open a patient record
- Go to phenotyping
- Click on an (i) button for a predefined phenotype -> Tooltip opens
- Click on "Browse related terms" -> Ontology browser opens
- Click _anywhere_ in the ontology browser dialog or on the dark screen behind it.

Result before this change: the tooltip disapears
Result after this change: the tooltip remains open

Known side effect: Tooltips from the ontology browser can only be closed by clicking on their x button (or by opening another tooltip, and not by clicking outside.

### DIFF
--- a/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
@@ -2073,7 +2073,7 @@ widgets.ModalPopup = Class.create({
 
     // Stop the propagation of mouse events beyond the dialog element to prevent unwanted interaction with anything in the back
     var _this = this;
-    ['mousedown', 'mouseup', 'click'].each(function(eventName) {
+    ['mousedown', 'click'].each(function(eventName) {
       _this.dialog.observe(eventName, function(event) {
         event.stop();
       });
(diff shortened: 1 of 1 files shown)


marta-:
@sashaandjic :
> if I attempt to move HPO browser, I can no longer click on it (it seems to be stuck and the only way out is by pressing esc)

Thanks, fixed in latest commit.

> pedigree editor: if I accidentally click on an item in the hpo browser, it will be selected and I cannot undo it while hpo browser is open

This is the issue I was talking about in [this comment](https://phenotips.atlassian.net/browse/PT-3428?focusedCommentId=38447&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-384). and one of the reasons why we're fixing [PT-3428](https://phenotips.atlassian.net/browse/PT-3428) by not showing the browse button at all (and consequently never activating the ontology browser in the pedigree editor), because there are too many issues with checkboxes (instead of Y/N/NA pickers) in the ontology browser. This is unelated to making dialogs truly modal, which is the purpose of this PR. 

> pedigree editor: when pressing on "i" button while searching, BROWSE button is hidden. It is only available on already selected terms

Again, this is unrelated and is handled by the quick fix for [PT-3428](https://phenotips.atlassian.net/browse/PT-3428), which is to not show the browse button anywhere.

Please note that this is PR affects _all_ modal dialogs, not just the ontology browser.

marta-:
This is no longer urgent. I'll close it for now and until I fix the side effect involving non-disappearing tooltips issue (mentioned in the description).

PR #2443, opened 2017-09-15
PT-3402: Add an authorisation module for allowing the owner and collaborators of family member (patient) records to access the family record and pedigree

sdumitriu:


marta-:
@sashaandjic 

For quick testing on existing data without migration, import[ Clean.Rights.xar](https://phenotips.atlassian.net/secure/attachment/30130/Clean.Rights.xar) and view the document it contains to remove the XWikiRights objects from family documents.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DenyAccessByDefaultAuthorizationModule.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return 110;
+    }
+
+    @Override
+    public Boolean hasAccess(User user, Right access, EntityReference entity)
+    {
+        if (!ObjectUtils.allNotNull(access, entity) || access.getTargetedEntityType() == null
+            || !access.getTargetedEntityType().contains(EntityType.DOCUMENT)) {
+            return null;
```
allasm: Just curious what is the point of doing any checks and returning `null` instead of just always returning `false` (presumably "deny by default" should always return `false`?)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/MemberAccessAuthorizationModule.java
```diff
@@ -0,0 +1,88 @@
...
+ * Implementation that allows view and edit access to families to the users that have access to at least one patient
+ * that is a member of that family (ignoring access granted through the patient visibility).
+ *
+ * @version $Id$
+ * @since 1.4
+ */
+@Component
+@Named("family-member-access")
+@Singleton
+public class MemberAccessAuthorizationModule implements AuthorizationModule
```
allasm: Since this is all about families having a `Family`somewhere in the name (e.g. `FamilyAccessAuthorizationModule`?) may help us find the right component to modify in the future, after all it provides access to families (and the fact that it is based on family members is internal logic)

allasm: ..actually, given we also have OwnerAccess module below, and all these modules are in family component, I was probably wrong about naming above

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DenyAccessByDefaultAuthorizationModule.java
marta-: * `ObjectUtils.allNotNull(access, entity)` - sanity check
* `access.getTargetedEntityType() == null || || !access.getTargetedEntityType().contains(EntityType.DOCUMENT)` - XWiki supports rights that are not specific to a document, e.g. administration, registration, scripting, programming. It is not the responsibility of this module to restrict those rights.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/MemberAccessAuthorizationModule.java
allasm: The name is a bit misleading, since this does not give access to family members or by family members, but instead access _to family_ by _users_ who have access to family members

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DenyAccessByDefaultAuthorizationModule.java
allasm: I see. I understood the sanity check - but just thought that even in "insane" conditions the "deny" module should deny

("not the responsibility of this module to restrict those right" - aha, this I did not know and one of the reasons I asked)

allasm: BTW, which module should be responsible for that? Can someone please describe the way the entire permission system works now?

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/MemberAccessAuthorizationModule.java
marta-: True, the name could be improved. Suggestions are welcome. Both @sdumitriu have had trouble coming up with a short but relevant name.

`AccessThroughFamilyMemberAuthorisationModule`? :)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DenyAccessByDefaultAuthorizationModule.java
marta-: These are instance-level rights. They do not concern documents, but actions such accessing `/register/` for creating accounts, or `/admin/` for administering. They are the responsibility of the XWiki platform authorization module, which is the last one interrogated when we return null here.

allasm: I see, and I presume it will return `false` in all the insane cases

marta-:
@sashaandjic did you also test family access via rest? (the version of rest that we have now)

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DenyAccessByDefaultAuthorizationModule.java
marta-: yes

sdumitriu: > just thought that even in "insane" conditions the "deny" module should deny
This is the deny module _for families_, so it should not deny everything, just access to family documents. But if the pieces required to determine if this is a request to access a document and of that document is a family are missing, this module can not (and should not) decide anything.

Even after the [XWikiACL](https://github.com/phenotips/phenotips/blob/master/components/security/authorization/src/main/java/org/phenotips/security/authorization/internal/XWikiACLAuthorizationModule.java) module runs, there is [https://github.com/phenotips/phenotips/blob/master/components/security/authorization/src/main/java/org/phenotips/security/authorization/internal/BaseAuthorizationModule.java](another failsafe module) that returns false unconditionally.

allasm: right

PR #2444, opened 2017-09-18
PT-3427: Open patients with private families should not show the pedigree at all for non-collaborators

marta-:
The problem affects the master branch after the merge of #2443.
Steps to reproduce on master:
* log in as Alice
* create a new patient
* draw a pedigree
* make the patient open
* log in as Bob: Bob can edit the patient, but cannot see the pedigree or that the patient has a family
* BUG: Bob can start drawing a new pedigree, with a new family, but when trying to save an error is displayed: "Unable to save pedigree: Patient P123 already belongs to a different family FAM123, and therefore can not be added to this one"

With this change, Bob does not have the option to draw a pedigree in this case, and there is a warning message informing him that the patient is already part of a family he can't access and that he doesn't have permission to draw a new pedigree.


marta-:
@sashaandjic how about "You do not have permission to access this patient's family nor pedigree."? Please note that currently the message is displayed only in edit mode, so it makes sense IMO to talk about access in general rather than just viewing. Should we also show the message in view mode?

sashaandjic:
Yes, that's better. I would show it in the view mode too.

marta-:
@sashaandjic Thanks, changes committed.

sashaandjic:
Tested. Looks very good.

sashaandjic:
- logging out while in edit mode of an open patient record, leaves the record/form in a weird state
- this anomaly is not presented in playground not in PhenoTips 1.3.2
screen shot 2017-09-19 at 1 22 36 pm
screen shot 2017-09-19 at 1 22 47 pm

marta-:
@sashaandjic can't be related to this PR. I assume it is present on master, correct?

sashaandjic:
i'm building master right now

sashaandjic:
yes, the problem present in master

components/pedigree/ui/src/main/resources/PhenoTips/UIX_Field__pedigree.xml
```diff
@@ -124,7 +124,13 @@
 {{velocity}}
 #if ($xcontext.action != 'export')
   #set ($family = $services.families.getFamilyForPatient($doc.name))
-  #pedigreeThumbnail($family, $doc.name)
+  #set ($familyRef = "$!doc.getObject('PhenoTips.FamilyReferenceClass').getValue('reference')")
+  #if ("$!family" == '' &amp;&amp; $familyRef != '' &amp;&amp; !$xwiki.getDocument($familyRef).isNew())
```
veronikaslc: just curious why do we use `isNew()` here?

marta-: `$services.families.getFamilyForPatient` returns `null` in 2 cases:
- there is no family for that patient
- there is a family but the current user doesn't have access to it

So we need to figure out when the user doesn't have access to the family. That would be when the patient has a family reference, and that corresponds to an actual family document. OTOH, if the patient has some invalid family reference, then we should allow to create a new family.

Invalid family reference means the document doesn't exist (`$xwiki.getDocument($familyRef).isNew()`) or is not a family document. The document referred by family reference can be absent for example on XAR import, when you import a patient with a family reference but you don't import that family, or (I think) when you delete the family document using XWiki's delete functionality rather than using the families service.

(Now that I'm detailing this, we could also capture the situation where the reference document exists but it's not a family document, i.e. it doesn't have a family object, which can happen when manually deleting the family object from it. But that's far less likely to happen than the other 2 scenarios.)

PR #2446, opened 2017-09-20
PT-3381: REST API for family permissions

a-gvo:


a-gvo:
Rebased on top of PT-3359-v2.

a-gvo:
Rebased on top of PT-3359-v2.

a-gvo:
Rebased on top of the latest master.

components/entity-access-rules/rest/src/main/java/org/phenotips/data/permissions/rest/CollaboratorResource.java
```diff
@@ -35,78 +35,83 @@
...
     /**
-     * Retrieve information about a particular collaborator. If the indicated patient record doesn't exist, or if the
+     * Retrieve information about a particular collaborator. If the indicated entity record doesn't exist, or if the
      * indicated principal is not a collaborator, or if the user sending the request doesn't have the right to view the
-     * target patient record, an error is returned.
+     * target entity record, an error is returned.
      *
-     * @param patientId internal identifier of a patient record
+     * @param entityId internal identifier of an entity record
+     * @param entityType the type of entity
```
allasm: One has to jump through a lot of code to figure out where to find which type to use (e.g. should it be "patients" or "patient" or what?). Apparently the type is defined and hard-coded in various XXXEntityManager.getType(), and only there, I think there must be a better way to get this information (it was OK for internal use, but should be more clear how to get the string value of the type for the entity one needs once external APIs are involved)

a-gvo: Added to the javadoc for now.

a-gvo:
@sashaandjic I just changed the javadoc, so no :)

PR #2447, opened 2017-09-20
PT-3438: Pedigree editor breaks when using non-default date input format

marta-:


allasm:
This seems to have fixed the problem. You are right about hairy code though, the true solution is probably to make date widget handle everything, and accept updated data as input which it does not at the moment (and so nodeMenu has to manually set the values of various dropdowns)

PR #2455, opened 2017-09-28
PT-3446 and PT-3449: Pedigree functionality breaks when attempting to display dates with decade (not year) plus day and month

marta-:
For steps to reproduce, see [PT-3446](https://phenotips.atlassian.net/browse/PT-3446) and [PT-3449](https://phenotips.atlassian.net/browse/PT-3449).

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
@@ -146,7 +146,7 @@ define([], function(){
 
         /** Returns true iff the minimum precision level is set, ie. the year. Useful for displaying the date. */
         isComplete: function() {
-            return (this.year !== null);
+            return (this.year !== null && (!this.range || this.range.years == 1));
         },
 
         onlyDecadeAvailable: function() {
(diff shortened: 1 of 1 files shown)


marta-:
Actually this fixes the bugs PT-3446 and PT-3449, put prevents dates entered as just decades from being displayed in the pedigree. Closing the PR.

marta-:
PR reopened with a fix that hopefully doesn't have any other unwanted side effects.

marta-:
Jenkins, build this please

sashaandjic:
PT-3446: tested, problem fixed, however the pedigree isn't shown in the tumbnail after successful save 
PT-3449; tested; problem fixed

marta-:
>however the pedigree isn't shown in the tumbnail after successful save

That's expected when saving the JSON from the object editor, as the code which generates the image is not executed. Fortunately this is not something users would normally do.

components/pedigree/resources/src/main/resources/pedigree/pedigreeDate.js
```diff
@@ -146,7 +146,7 @@ define([], function(){
 
         /** Returns true iff the minimum precision level is set, ie. the year. Useful for displaying the date. */
         isComplete: function() {
-            return (this.year !== null);
+            return (this.year !== null && (!this.range || this.range.years == 1));
```
allasm: Good you reverted this, the fact that range is "1 year" does not mean the date is "complete" in the sense we know it with at least some precision (e.g. `range == 1` and `day == `1 but `month == null && year == null`)

marta-: Yep, I had misunderstood and figured out the mistake when birth years stopped displaying.

PR #2456, opened 2017-10-03
PT-3440: Add the option to list other family members on the Patient record.

veronikaslc:


allasm:
Approved the code and tested (quickly)

sashaandjic:
- new option "Display family members" needs to be moved to Administration: Patient form structure/ Family history and pedigree
- label "Family members:" needs to be displayed in bold fonts
- the whole Family members block needs to be aligned with Maternal ethnicity
- unrelated ( shall we move Global mode of inheritance: to to right and align it to Maternal ethnicity as well? if so i'll create a new JIRA issue)

screen shot 2017-10-05 at 11 29 34 am
)

allasm:
a) > new option "Display family members" needs to be moved to Administration: Patient form structure/ Family history and pedigree; i think it be set to true by default

This comment makes sense, just in order for it to be in patient form configuration it needs tobe a separate UIX element I think, and it is extra work, plus logically it is just an additional Ui element for the pedigree UIX, so I would say there is no need to change it at least for now

b) label "Family members:" needs to be displayed in bold fonts

Not sure if it will look better, since again, this is not a separate section, and we only bold section titles. We can try to see if it looks better visually

c) the whole Family members block needs to be aligned with Maternal ethnicity

Not sure about aligning (may be technically different), but I agree a bit more whitespace between this block and pedigre eimage may be better

d) unrelated ( shall we move Global mode of inheritance: to to right and align it to Maternal ethnicity as well? if so i'll create a new JIRA issue)

IMHO definitely not. (Very) hard to implement (to put two unrelated blocks in two columns), and multi-column layouts tend to look different in different browsers and break more easily. Not to mention it will be ugly as well, when "global mode" block is taller than the other block.

sashaandjic:
The label font is bolded in view mode only.  Also in view mode it is wrapped. 
screen shot 2017-11-07 at 2 05 31 pm
screen shot 2017-11-07 at 2 05 55 pm

allasm:
we should keep in mind today's discussion (about adding a patient to a family without a pedigree) in mind when redesigning this part of the UI. Current place where family ID is seems like a good place for that new functionality (button "add to family" or smth like that - and correspondingly family link maybe, when there is no pedigree?)

sashaandjic:
@veronikaslc  reminder:
align and use bold font
don t wrap

allasm:
I'm not sure aligning will work well when this section is disabled?

veronikaslc:
@sashaandjic codereview comments addressed, please retest

sashaandjic:
It would be perfect if "Family members" label does not wrap in view mode.
screen shot 2017-12-05 at 11 53 22 am
screen shot 2017-12-05 at 11 53 33 am

veronikaslc:
@sashaandjic wrapping fixed

sashaandjic:
tested. perfect!

components/family-studies/ui/src/main/resources/PhenoTips/FormMacros.xml
```diff
@@ -76,5 +76,22 @@
 #end
 )))
 #end
+
+#macro (__displayFamilyMembers $family)
```
sdumitriu: This is not the right place for this macro. `FormMacros` is supposed to be a container of just macros used for displaying a generic form, and not something specific to families or patients. Create a new `FamilyMacros` instead.

components/pedigree/api/src/main/resources/ApplicationResources.properties
```diff
@@ -19,6 +19,7 @@ phenotips.patientRecord.deleteWarning=Are you sure you want to delete this patie
 phenotips.patientRecord.deletePatient=Delete patient
 phenotips.patientRecord.deletePatientAndFamily=Delete patient and family
 phenotips.patientRecord.deleteWarning.cancelButton=Cancel
+phenotips.patientRecord.familyListTitle=Family members:
```
sdumitriu: Use this key instead: `phenotips.patientRecord.family.members.title`.

sdumitriu:
Two more things:

screenshot_20171206_164607

- the label "Family members" shows up inside the "Draw Pedigree" image for a new patient
- the label should not show up at all if there are no family members yet (because there's no family)

sdumitriu:
Ah, and in view mode nothing should be displayed if there's no family:

screenshot_20171206_164925

veronikaslc:
@sdumitriu I can not reproduce the UI bug on my Firefox. But other comments are addressed.

sashaandjic:
@veronikaslc  The UI bug was visible in Chrome and it is now gone. However I found this problem:
1. create simple family 
screen shot 2017-12-07 at 10 27 49 am
2. Click on "x" button next to the tumbnail and select any offered option, this happens:
screen shot 2017-12-07 at 10 28 07 am

There is also problem with pedigree after the deletion, but I this may have nothing to do with this pull request. If you pick option "REMOVE PATIENT FROM FAMILIY" and then click on "CANCEL CHANGES SINCE LAST SAVE" the pedigree seem to lose the link to the patient.
screen shot 2017-12-07 at 10 39 13 am

veronikaslc:
@sashaandjic issue comments addressed

sashaandjic:
retested. all good!

PR #2458, opened 2017-10-06
PT-3454: Uniqueness check for family external ID (JS/UI only)

veronikaslc:
`UniquenessValidator` is moved to the `Widgets.xml` to be used by both patient and family forms.

### DIFF
--- a/components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml
+++ b/components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml
@@ -160,15 +160,6 @@
       );
     }
   });
-
-  // Family External ID uniqueness check
-  ([$('body')]).each(function(element) {
-      element.select('.external_id input').each(function(input) {
-        if (!input.__uniqueness_validator) {
-          input.__uniqueness_validator = new XWiki.widgets.UniquenessValidator(input, "family");
-        }
-      });
-  });
 });</code>
     </property>
     <property>
--- a/components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml
+++ b/components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml
@@ -160,6 +160,15 @@
       );
     }
   });
+
+  // Family External ID uniqueness check
+  ([$('body')]).each(function(element) {
+      element.select('.external_id input').each(function(input) {
+        if (!input.__uniqueness_validator) {
+          input.__uniqueness_validator = new XWiki.widgets.UniquenessValidator(input, "family");
+        }
+      });
+  });
 });</code>
     </property>
     <property>
--- a/components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
+++ b/components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
@@ -108,16 +108,10 @@
       </scope>
     </class>
     <property>
-      <content>{{include reference="PhenoTips.FormMacros" /}}
+      <content>{{include reference="PhenoTips.PatientSheetMacros" /}}
 
 {{velocity}}
-#if($xcontext.action == 'edit' || $doc.display('external_id') != '')
-  (% class="fieldset external_id" %)(((
-    #__form_label('external_id' 'section')
-
-    #__form_field('external_id')
-  )))
-#end
+#__displayFormatted('2-col' 'external_id' '' '' '')
 {{/velocity}}</content>
     </property>
     <property>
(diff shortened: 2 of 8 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/CheckID.xml
```diff
@@ -44,11 +44,16 @@
 #if ($request.getParameter('eid') &amp;&amp; $request.getParameter('id'))
   #set ($externalIdentifier = "$!{request.eid}")
   #set ($identifier = "$!{request.id}")
-  #if ($externalIdentifier == $xwiki.getDocument($services.model.resolveDocument($identifier, $services.model.resolveSpace('data'))).getObject('PhenoTips.PatientClass').getProperty('external_id').value)
+  #set ($principal = "$!{request.principal}")
```
marta-: `principal` is not the right concept here, that's for users and groups. Should be `entity`.

components/family-studies/ui/src/main/resources/PhenoTips/FamilyRecordField_externalId.xml
```diff
@@ -108,16 +108,10 @@
...
 
 {{velocity}}
-#if($xcontext.action == 'edit' || $doc.display('external_id') != '')
-  (% class="fieldset external_id" %)(((
-    #__form_label('external_id' 'section')
-
-    #__form_field('external_id')
-  )))
-#end
+#__displayFormatted('2-col' 'external_id' '' '' '')
```
marta-: Why this change? The 2 column layout doesn't look well in the family page, please see https://github.com/phenotips/phenotips/commit/9e4d44967232e0ff4cfecabe6d4d8679497b4ea8#diff-67005384f8134222422f15ae2a7ac382 .

components/family-studies/ui/src/main/resources/PhenoTips/FamilyContentTopMenu.xml
```diff
@@ -160,6 +160,15 @@
...
   });
+
+  // Family External ID uniqueness check
+  ([$('body')]).each(function(element) {
+      element.select('.external_id input').each(function(input) {
+        if (!input.__uniqueness_validator) {
+          input.__uniqueness_validator = new XWiki.widgets.UniquenessValidator(input, "family");
+        }
+      });
+  });
```
marta-: Semantically, the menu is not the right place for this piece of code. Since it is strictly related to the Family external id, I would put it in a new jsx in the `FamilyRecordField_externalId.xml` document.

Even better, since the backend code only seems to be able to deal with external ids anyway, i would leave this in the UniquenessValidator widget where it was originally, and figure out the entity type (patient or family) from a css class or a hidden input that is generated by the external id uix.

components/patient-data/ui/src/main/resources/PhenoTips/CheckID.xml
```diff
@@ -44,11 +44,16 @@
...
+  #set ($className = 'PhenoTips.PatientClass')
+  #if ($principal == 'family')
+    #set ($className = 'PhenoTips.FamilyClass')
+  #end
+  #if ($externalIdentifier == $xwiki.getDocument($services.model.resolveDocument($identifier, $services.model.resolveSpace('data'))).getObject($className).getProperty('external_id').value)
```
marta-: This always looks in the data space, for both patients and fsmilies. It doesn't look right.

veronikaslc:
@marta- Thank toy, codereview comments addressed @sashaandjic please, test

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -5590,6 +5590,18 @@ document.observe("xwiki:dom:updated", init);
     }
   });
 
+  var init = function(event) {
+    ((event &amp;&amp; event.memo.elements) || [$('body')]).each(function(element) {
+      element.select('input.check-unique, .external_id input').each(function(input) {
+        if (!input.__uniqueness_validator) {
+          var entity = $("entity-type") ? $("entity-type").value : "patients";
```
marta-: This is relying on the hidden input with `id="entity-type"` defined by the access rights management UIX, correct? If that's the case, then external_id uniqueness won't be properly checked for families when the permissions manager is disabled (and we do have PhenoTips instalations configured this way at the moment). Now that we want to use that input in a new context, maybe it should be defined in the patient and family sheets instead of the permissions UIX?

Also, if we can't figure outthe entitytype I would just do nothing instead of defaulting to `patients`.

allasm: The new entity resolver component can resolve entity type by document name (e.g.  P0000xxx or FAM000xxx). Maybe we should use that in one way or another?

And +1 to having entity-type always defined, regardless of optional component presence

veronikaslc:
@marta- @allasm codereview comments addressed

PR #2459, opened 2017-10-10
PT-3457: Unable to open node that contains cancers with unspecified age [fixed]

a-gvo:


components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
```diff
@@ -872,7 +872,7 @@ define([
             var toDisplayedLateralityMap = {"bi" : "Bilateral", "u" : "Unilateral", "r" : "Right", "l" : "Left"};
 
             var toStoredAgeFx = function(value) {
-                if (isNaN(parseInt(value))) {
+                if (value && isNaN(parseInt(value))) {
```
allasm: if `!value` the `else` branch will raise an exception

a-gvo: Right, fixed.

PR #2460, opened 2017-10-10
PT-3423: Implement a migrator from old internal pedigree format to new

allasm:
@veronikaslc: this branch is based on top of PT-2617 branch - the only actual commits that need to be code-reviewed are the last two (clearly marked as PT-3423). However since it has to be tested together all the commits from PT-2617 branch are included in this pull request.

### DIFF
--- a/components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
+++ b/components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
@@ -1,0 +1,558 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+
+package org.phenotips.studies.family.migrations;
+
+import org.phenotips.Constants;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.model.EntityType;
+import org.xwiki.model.reference.DocumentReferenceResolver;
+import org.xwiki.model.reference.EntityReference;
+import org.xwiki.model.reference.EntityReferenceSerializer;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.apache.commons.lang3.StringUtils;
+import org.hibernate.HibernateException;
+import org.hibernate.Query;
+import org.hibernate.Session;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+
+import com.xpn.xwiki.XWiki;
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.store.XWikiHibernateBaseStore.HibernateCallback;
+import com.xpn.xwiki.store.XWikiHibernateStore;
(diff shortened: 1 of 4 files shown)


components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
```diff
@@ -0,0 +1,558 @@
...
+import com.xpn.xwiki.store.migration.XWikiDBVersion;
+import com.xpn.xwiki.store.migration.hibernate.AbstractHibernateDataMigration;
+
+/**
+ * Migrate from old internal pedigree format to the new.
+ *
+ * The migration also includes all pedigree migrations previously implemented in javascript (in versionUpdater.js)
+ *
+ * @version $Id$
+ * @since 1.4M2
```
veronikaslc: in JIRA Fix Version for PT-3423 is 1.4-milestone-3

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
```diff
@@ -0,0 +1,558 @@
...
+            } else {
+                return this.parentID.toString() + " -> <" + this.path.toString() + ">";
+            }
+        }
+    }
+
+    private boolean updatePedigree(BaseObject pedigreeXObject, XWikiContext context, String documentName)
+    {
+        String oldPedigreeAsText = pedigreeXObject.getStringValue(PEDIGREECLASS_JSONDATA_KEY);
+        if (!StringUtils.isEmpty(oldPedigreeAsText)) {
```
veronikaslc: why not `StringUtils.isBlank()`

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
```diff
@@ -0,0 +1,558 @@
...
+        if (pedigreeSettings != null) {
+            newPedigree.put(PEDIGREE_SETTINGS_KEY, pedigreeSettings);
+        }
+
+        // Note1: the assumption is that pedigree is consistent and correct, if a required key
+        //        is missing an exception will be thrown (e.g. by getJSONArray()), which is expected
+        //
+        // Note2: migrated pedigree will preserve all node IDs, so no ID mapping is required
+
+        JSONArray oldData = oldPedigreeData.getJSONArray(OLD_PEDIGREE_GRAPH_KEY);
```
veronikaslc: why not `optJSONArray()`?

allasm: because if there is no array we want to throw an exception (and abort conversion for this family), this is an abnormal situation that we can't handle

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
allasm: Because "empty" means no data (which is OK, just "do nothing"), while blank (but not empty) means corrupt data - and it is best to handle corrupt data together with other cases of corrupt data e.g. non-blank, non-empty, but still wrong

allasm: (to be precise, empty is also "not OK" here, but it is still a different kind of error compared to wrong data)

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
allasm: Will update

components/family-studies/migrations/src/main/java/org/phenotips/studies/family/migrations/R71507PhenoTips3423DataMigration.java
```diff
@@ -0,0 +1,558 @@
...
+        // 3. process all relationships, now that we know all the partners
+        if (!relationshipParents.isEmpty()) {
+            newPedigree.put(PEDIGREE_RELATIONS_KEY, new JSONArray());
+        }
+        for (Map.Entry<Integer, List<PartnerPath>> entry : relationshipParents.entrySet()) {
+            Integer id = entry.getKey();
+            List<PartnerPath> parents = entry.getValue();
+            if (parents.size() != 2) {
+                throw new IllegalArgumentException("Unsupported pedigree: a relationship with ID="
+                    + id.toString() + "doesn't have exactly two partners");
```
veronikaslc: need a space between " and d, otherwise exception message will be hard to read, id is merging with text

allasm: I agree, but this is a one-time message (during data migration) which will likely never happen (for it to happen, need to have an invalid pedigree to start with), so to avoid further delay with merging I will commit as is

PR #2464, opened 2017-10-10
PT-3460: Non-categorised custom phenotypes disappear from saved patient record

a-gvo:


components/patient-tools/src/main/java/org/phenotips/tools/PropertyDisplayer.java
```diff
@@ -110,12 +112,12 @@ protected void putTermsInSections(List<String> selectedTerms, Map<String, List<S
...
         for (String value : selectedTerms) {
             VocabularyTerm term = this.ontologyService.getTerm(value);
-            VocabularyTerm root = this.ontologyService.getTerm("HP:0000001");
+            VocabularyTerm root = this.ontologyService.getTerm(ROOT_ID);
             List<String> categories = new LinkedList<>();
             categories.addAll(this.getCategoriesFromOntology(value));
             categories.addAll(this.getCategoriesFromCustomMapping(value, customCategories));
             if (categories.isEmpty()) {
-                categories.add("HP:0000118");
+                categories.add(ROOT_ID);
```
allasm: I'm not sure if we should fix this part and replace `HP:0000118` with `HP:0000001`, or fix the UI (javascript/velocity?) instead to display terms in the 118 category in the "all" section on the patient form, which seems like a better fix (@sdumitriu?)

allasm: (both fixes seem to be small, so it should not make a huge difference in terms of development time, so we should do what makes more sense)

PR #2465, opened 2017-10-10
PT-3461: Pedigree crashes when patient sheet has free-text disorders

a-gvo:


a-gvo:

PR #2466, opened 2017-10-10
PT-3462: Custom disorders are not displayed correctly in view and edit mode

a-gvo:


a-gvo:

PR #2467, opened 2017-10-11
PT-3399: Add a REST service for suggesting patients matching an input. 

veronikaslc:
Initial implementation.


components/navigation/ui/src/main/resources/XWiki/SearchSuggestConfig.xml
```diff
@@ -526,7 +526,7 @@
       <resultsNumber>5</resultsNumber>
     </property>
     <property>
-      <url>$xwiki.getURL('PhenoTips.SuggestPatientsService', 'get', 'outputSyntax=plain&amp;permission=view')</url>
+      <url>$xcontext.getURLFactory().getServerURL($xcontext.getContext())/rest/patients/suggest?permission=view</url>
```
sdumitriu: You don't need to specify `$xcontext.getURLFactory().getServerURL($xcontext.getContext())`, use `${request.contextPath}/rest/patients/suggest?permission=view` works.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsSuggestionsResourceImpl.java
```diff
@@ -0,0 +1,266 @@
...
+
+        xmlResult.append("<rs id=\"").append(getURL(patient)).append("\" ");
+        xmlResult.append("info=\"").append(escapedReference).append("\">");
+
+        String description = getDescription(patient);
+
+        // Add family info
+        Family family = this.familyRepository.getFamilyForPatient(patient);
+        if (family != null) {
+            xmlResult.append("familyId=\"").append(family.getId()).append("\" ");
```
sdumitriu: This shows up in the label, since the tag was closed already on line 221.

veronikaslc: Seems we did not use `familyId` anyways (based on [old code](https://github.com/phenotips/phenotips/pull/2467/files#diff-5422a2cd9f7ae082de4ecf1ee418661dL136) ) so I'll just remove it in next commit

PR #2469, opened 2017-10-12
SPO-5: Expose ncr.ccm.sickkids.ca through services.phenotips.org/ncr/

a-gvo:


components/annotation-services/ncr/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.8.6</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore</artifactId>
+      <version>4.4.5</version>
```
veronikaslc: no need to specify some dependency versions, for ex. this one is specified here https://github.com/phenotips/phenotips/blob/master/pom.xml#L261

components/annotation-services/ncr/src/main/java/org/phenotips/services/annotations/ncr/internal/NCRServlet.java
```diff
@@ -0,0 +1,184 @@
...
+     * @param request the incoming request
+     * @return a list of name-value pairs, in format expected by the NCR service
+     * @throws JsonProcessingException if there was an error converting request data to json string
+     */
+    @Nonnull
+    private String translateRequest(@Nonnull final HttpServletRequest request) throws JsonProcessingException
+    {
+        final String text = request.getParameter(CONTENT_LABEL);
+        final Map<String, String> params = new HashMap<>();
+        params.put(TEXT_LABEL, StringUtils.isNotBlank(text) ? text : StringUtils.EMPTY);
```
veronikaslc: you may use `StringUtils.defaultIfBlank(text, StringUtils.EMPTY)`

veronikaslc:
code looks good, some very minor comments

components/annotation-services/ncr/src/main/java/org/phenotips/services/annotations/ncr/internal/NCRServlet.java
```diff
@@ -0,0 +1,184 @@
...
+            LOGGER.error("An unexpected error occurred: {}", e.getMessage());
+            response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    /**
+     * Translates the request parameters to the format expected by the NCR service located at {@link #SERVICE_URL}.
+     *
+     * @param request the incoming request
+     * @return a list of name-value pairs, in format expected by the NCR service
```
allasm: To make it more clear both what the code does and which version of NCR this is supposed to work with I would add an example (or specification) of the expected format in the comments. Same with reply format

components/annotation-services/ncr/pom.xml
a-gvo: Fixed

components/annotation-services/ncr/src/main/java/org/phenotips/services/annotations/ncr/internal/NCRServlet.java
a-gvo: Fixed

components/annotation-services/ncr/src/main/java/org/phenotips/services/annotations/ncr/internal/NCRServlet.java
a-gvo: Fixed

PR #2479, opened 2017-10-16
PT-3463: Add a REST service for suggesting families matching an input

veronikaslc:


components/navigation/ui/src/main/resources/XWiki/SearchSuggestConfig.xml
```diff
@@ -970,7 +970,7 @@
       <resultsNumber>5</resultsNumber>
     </property>
     <property>
-      <url>$xwiki.getURL('PhenoTips.FamilySearch', 'get', 'outputSyntax=plain&amp;permission=view')</url>
+      <url>${request.contextPath}/rest/families/suggest?permission=view</url>
```
sdumitriu: The global search doesn't work anymore. Trying to write something in that input shows a modal exception (`Not acceptable`) because the search widget requests `application/xml`, and the rest services can only produce JSON. Then, even if I get it to request JSON instead, nothing is actually displayed since the search widget expects a specific kind of response, which isn't compatible with the output generated by the patients/families suggest REST services.

components/family-studies/rest/src/main/java/org/phenotips/studies/family/rest/FamiliesSuggestionsResource.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param input criterion to select families by
+     * @param maxResults maximal number of results for each query
+     * @param requiredPermission permission a user has to have over each family in the result
+     * @param orderField field used for ordering the families, can be one of {@code id} (default) or {@code eid}
+     * @param order the sorting order, can be one of {@code asc} (default) or {@code desc}
+     * @return a list of family records
+     */
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @RequiredAccess("view")
```
sdumitriu: This shouldn't be here. The `RequiredAccess` annotation only makes sense on a resource for a specific patient. With this annotation in place, each access triggers an ERROR in the logs: `Failed to read document [null]: null`.

components/family-studies/rest/src/main/java/org/phenotips/studies/family/rest/internal/DefaultFamiliesSuggestionsResourceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+@Named("org.phenotips.data.rest.internal.DefaultFamiliesSuggestionsResourceImpl")
+@Singleton
+public class DefaultFamiliesSuggestionsResourceImpl extends XWikiResource implements FamiliesSuggestionsResource
+{
+    @Inject
+    private PhenotipsFamilyExport familyExport;
+
+    @Override
+    public String suggest(String input, @DefaultValue("10") int maxResults,
+        @DefaultValue("view") String requiredPermission, @DefaultValue("id") String orderField,
```
sdumitriu: `DefaultValue` annotations are only read from the interface, so you should remove them from here.

veronikaslc:
@sdumitriu codereview comments addressed & rebased on master
@sashaandjic please, test

veronikaslc:
@sdumitriu same codereview comments also addressed in PT-3399, so PRs can be tested together

components/family-studies/ui/src/main/resources/PhenoTips/FamilySelector.xml
```diff
@@ -129,8 +129,7 @@
 
             var _this = this;
 
-            this.familySearch = new XWiki.Document('FamilySearch', 'PhenoTips');
-            this.familySearchUrl = this.familySearch.getURL('get', 'outputSyntax=plain&amp;rand='+ Math.random());
+            this.familySearchUrl = XWiki.contextPath + "/rest/families/suggest";
```
sdumitriu: Still doesn't work, append `?` at the end: `"/rest/families/suggest?"`

veronikaslc: @sdumitriu indeed, sorry, URL corrected in next commit

PR #2480, opened 2017-10-16
PT-3471: Make the patients and families script services more similar to the underlying entity repository interface

sdumitriu:


components/family-studies/api/src/main/java/org/phenotips/studies/family/script/FamilyScriptService.java
```diff
@@ -53,12 +53,34 @@
...
+    public Family create()
     {
         return this.familyTools.createFamily();
     }
 
     /**
+     * Creates an empty family.
+     *
+     * @return Family object corresponding to the newly created family.
+     */
```
marta-: Deprecate?

sdumitriu: Yes, but in next milestone, to avoid warnings in the logs.

PR #2482, opened 2017-10-17
PT-3104: Different/inconsistent behaviour when filtering in Advanced data filtering and Export [fixed]

a-gvo:


components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
```diff
@@ -357,6 +357,8 @@ document.observe('xwiki:dom:loading', function() {
       document.observe('xwiki:multisuggestpicker:selectionchanged', updateMatchCounter);
       content.select('input[type=radio]').invoke('observe', 'click', updateMatchCounter);
       content.select('input[type=checkbox]').invoke('observe', 'click', updateMatchCounter);
+      content.select('input[name=external_id]').invoke('observe', 'input', updateMatchCounter);
```
sdumitriu: Why not `input[type=text]`?

PR #2487, opened 2017-10-19
PT-3480: Better sizing of the livetable actions column

marta-:


components/navigation/ui/src/main/resources/PhenoTips/DBWebHomeSheet.xml
```diff
@@ -420,6 +420,10 @@ document.observe('xwiki:livetable:newrow', function(tableEvent){
   display: none;
 }
 
+xwiki-livetable td.actions {
```
sdumitriu: Missing a dot at the start of the line.

marta-: Ouch, wrong copy/paste... Fixed, thanks!

veronikaslc:
looks better!

PR #2490, opened 2017-10-24
PT-3477: Script service for PrimaryEntityResolver

a-gvo:


components/entities/api/src/test/java/org/phenotips/entities/internal/DefaultPrimaryEntityResolverTest.java
```diff
@@ -119,7 +121,10 @@ public void setUp() throws Exception
 
         this.referenceResolver = this.mocker.getInstance(DocumentReferenceResolver.TYPE_STRING, "current");
 
-        this.component = this.mocker.getComponentUnderTest();
+        this.component = (DefaultPrimaryEntityResolver) spy(this.mocker.getComponentUnderTest());
+
+        doReturn(true).when(this.component).isValidManager(this.familyResolver);
+        doReturn(true).when(this.component).isValidManager(this.patientResolver);
```
sdumitriu: I'm not a fan of this. You're bypassing an important check in the class.

...However, there's no easy way of not doing this.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/SecurePatientEntityManager.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Secure implementation of patient data access service using XWiki as the storage backend.
+ *
+ * @version $Id$
+ * @since 1.4
+ */
+@Named("Patient/secure")
+@Component(roles = {PrimaryEntityManager.class})
+@Singleton
+public class SecurePatientEntityManager extends PatientEntityManager
```
sdumitriu: This is not secured in any way. You need to also move the relevant code from `SecurePatientRepository` here.

a-gvo: Fixed

PR #2493, opened 2017-10-26
PT-3424: Implement a migrator from auto-generated internal pedigree format

allasm:
For testing, first need to migrate "old family studies" data, then used migrated pedigrees (which will be in "simple JSON" format, as opposed to "old internal") as input for this migrator

For reviewers: only the last two commits are for this issue, all others are from branches this is based on


sashaandjic:
old family studies are not migrating

allasm:
Fixed migrations

PR #2500, opened 2017-11-14
PT-3285: In pedigree, offer an option to delete unlinked patient records

allasm:
When a patient is unlinked, there is now a new button in the corner of the dialogue to remove the patient record being unlinked.

In addition to the obvious functionality (record deletion/error handling) need to make sure undo/redo works as expected, which in this case means "previous pedigree states look as they used to, except that the deleted record is removed". This should be so for all "undo" states, including those produced by changes unrelated to this patient record.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/controller.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/controller.js
@@ -23,6 +23,7 @@ define([
             document.observe("pedigree:node:remove",               this.handleRemove);
             document.observe("pedigree:node:setproperty",          this.handleSetProperty);
             document.observe("pedigree:node:modify",               this.handleModification);
+            document.observe("pedigree:patient:deleterequest",     this.handleDeletePatientRecords);
             document.observe("pedigree:person:drag:newparent",     this.handlePersonDragToNewParent);
             document.observe("pedigree:person:drag:newpartner",    this.handlePersonDragToNewPartner);
             document.observe("pedigree:person:drag:newsibling",    this.handlePersonDragToNewSibling);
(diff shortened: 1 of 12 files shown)


components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -1016,12 +1096,24 @@ define([
...
                 var oldLinkID = editor.getNode(nodeID).getPhenotipsPatientId();
                 editor.getOkCancelDialogue().showWithCheckbox("<br/><b>Do you want to remove the connection between this<br/>patient </b>(" + editor.getGraph().getPatientDescription(nodeID, true) +
                         ")<b> and this pedigree node?</b><br/><br/><br/>" +
                         "<div style='margin-left: 30px; margin-right: 30px; text-align: center'>" +
                         "Please note that if you do not assign this patient to another pedigree<br/>node this patient will be removed from this family</div><br/>",
-                        'Remove the connection?', 'Clear data from this pedigree node', true, "Remove link", processLinkCallback, "Cancel", onCancelAssignPatient );
+                        'Remove the connection?', 'Clear data from this pedigree node', true,
+                        "Remove link", processLinkCallback,
+                        "Cancel", onCancelAssignPatient,
+                        "Delete patient record", deletePatientRecord);
```
sdumitriu: `Cancel` in the middle?

components/pedigree/resources/src/main/resources/pedigree/externalEndpoints.js
```diff
@@ -107,6 +107,10 @@ define([
             return this.vocabulariesService + "/omim";
         },
 
+        getPatientDeleteURL: function(patientId) {
+            return XWiki.contextPath + "/rest/patients/" + patientId + "?method=DELETE";
```
sdumitriu: I wonder if we need to add that `rand` parameter for IE. TODO: check if deleting the same patient twice works or not in IE.

components/pedigree/resources/src/main/resources/pedigree/undoRedoManager.js
```diff
@@ -15,6 +15,51 @@ define([
...
+            document.observe("pedigree:patient:created", this.handlePatientCreated.bind(this));
+        },
+
+        handlePatientDeleted: function(event)
+        {
+            var removedID = event.memo.phenotipsPatientID;
+
+            this._deletedPatients[removedID] = true;
+
+            // Need to modify the undo/redo stack and remoe all references o the deleted patient:
```
sdumitriu: Typo: `o -> to`.

components/pedigree/resources/src/main/resources/pedigree/view/patientDropLegend.js
```diff
@@ -51,21 +51,26 @@ define([
...
             document.observe('pedigree:patient:linked', function (event){
-                if (this._notLinkedPatients.hasOwnProperty(event.memo.phenotipsID)) {
-                    this._deletePatientElement(event.memo.phenotipsID, this._notLinkedPatients[event.memo.phenotipsID].type);
-                    delete this._notLinkedPatients[event.memo.phenotipsID];
-                }
-            }.bind(this));
+                removeCase(event.memo.phenotipsID);
+            });
+
+            // remove patient from a legen on patient record deleted event
```
sdumitriu: Typo: legen.

components/pedigree/resources/src/main/resources/pedigree/view/legend.js
```diff
@@ -683,6 +683,7 @@ define([
             if (editor.isReadOnlyMode()) {
                 return;
             }
+            editor.getNodeMenu().hide();
```
sdumitriu: Is this related to this pull request?

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -149,6 +150,85 @@ define([
...
+                        editor.getOkCancelDialogue().showError(message, title, "OK", null);
+                    }
+                }
+
+                var completed = 0;
+                phenotipsIdListToRemove.forEach(function(phenotipsPatientID) {
+                    new Ajax.Request(editor.getExternalEndpoint().getPatientDeleteURL(phenotipsPatientID), {
+                        method: 'POST',
+                        onCreate: function() {
+                            document.fire("pedigree:load:start", {"message": "Deleting patient " + phenotipsPatientID + "..."});
```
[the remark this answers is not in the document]
sdumitriu: Sending these events like this won't work nicely. Since the loading modal is a singleton, if 3 patients are deleted, this event will be sent 3 times in a row, but only one dialog is shown, with only the last patient ID shown. Then the first deletion to complete will close the dialog, even though other requests are in progress. Maybe trigger the dialog once before the loop, and close it when all requests are completed.

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -149,6 +150,85 @@ define([
             editor.getView().unmarkAll();
         },
 
+        handleDeletePatientRecords: function(event)
+        {
+            console.log("event: " + event.eventName + ", memo: " + Helpers.stringifyObject(event.memo));
```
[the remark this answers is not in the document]
marta-: `console.log` is always defined in the pedigree code, so this should be fine.

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -149,6 +150,85 @@ define([
...
+
+            var title       = "Delete patient record" + (phenotipsIdListToRemove.length > 1 ? "s" : "") + "?";
+            var buttonLabel = "Delete" + (phenotipsIdListToRemove.length > 1 ? " all listed records" : "");
+            var message     = "The deletion of " + (phenotipsIdListToRemove.length > 1 ? "patient records" : "a patient record") +
+                              " (" + patientListString + ") is not reversible. Are you sure you wish to continue?";
+
+            var deleteRecords = function() {
+
+                // perform the callback regardless of if deletion was successful
+                callbackOnApprove();
```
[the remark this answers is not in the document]
allasm: Will fix

components/pedigree/resources/src/main/resources/pedigree/controller.js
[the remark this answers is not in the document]
allasm: The event should definitely be renamed, I just did nto do it as part of this PR, but since I modified relevant code I probably should have

I remember I had some reasoning why a few events in a row is not a problem, but let me re-think

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -149,6 +150,85 @@ define([
...
+                        onSuccess: function() {
+                            successfulDeletions.push(phenotipsPatientID);
+                            var event = { "phenotipsPatientID": phenotipsPatientID };
+                            document.fire("pedigree:patient:deleted", event);
+                        },
+                        onFailure: function() {
+                            failedDeletions.push(phenotipsPatientID);
+                        },
+                        onComplete: function() {
+                            document.fire("pedigree:load:finish");
```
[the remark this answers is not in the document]
allasm: Same as above, it was a bad event name to start with, will rename

components/pedigree/resources/src/main/resources/pedigree/controller.js
allasm: Not sure what is the question? If you mean the porder of buttons in UI, yes, it will look like

```----------------------------------------------------
                     message text

           [Unlink] [Cencel]                 [Delete]
----------------------------------------------------```

File: components/pedigree/resources/src/main/resources/pedigree/externalEndpoints.js (hunk 2 above)
Comment (allasm) [reply components/pedigree/resources/src/main/resources/pedigree/externalEndpoints.js]: I wonder if those `rand`s are there for IE8 or IE9 (pedigree can be used in read-only mode in IE8, not sure if we still care in 2017/2018)

**Comment by allasm** [comment]
Need to be re-tested

File: components/pedigree/resources/src/main/resources/pedigree/undoRedoManager.js [hunk 10]
```diff
@@ -15,6 +15,41 @@ define([
             this._stack        = [];
             this._MAXUNDOSIZE  = 100;
             this._savedState   = "";
+
+            // observe pation deletion events to remove allreferences to a deleted patient from all undo/redo states
```
veronikaslc: typos `pation` and `allreferences`

components/pedigree/resources/src/main/resources/pedigree/undoRedoManager.js
sdumitriu: This is still present.

PR #2503, opened 2017-11-21
PT-3502: Improve the formatting of free text in diagnosis section

a-gvo:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -1220,6 +1222,7 @@
          id: this.fld.value,
          value: this.fld.value,
          category: customItemCategoriesElt,
+         custom: true,
          info: new Element('div', {'class' : 'hint'}).update('(your text, not a standard term)')
```
sdumitriu: This doesn't seem to be translated, something to fix...

PR #2505, opened 2017-11-30
PT-1447: Better usability for the pedigree date picker; PT-2274: Date is now saved before blur

veronikaslc:
- Better behavior for year input; slight improvement to input character capturing

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
@@ -1,264 +1,401 @@
-define([
-    "pedigree/model/helpers"
-  ], function(
-    Helpers
-  ){
+define([], function(){
 
-  var PedigreeFuzzyDatePickerDropdown = Class.create({
-    initialize : function(options) {
-      this.span     = new Element('span');
-      this.options  = options;
-      this.callback = null;
-    },
+  /**
+  * AbstractFuzzyDateInput is an object representing a text input for the @class FuzzyDatePicker
+  *
+  * @class AbstractFuzzyDateInput
+  * @constructor
+  * @param {element} A text input element. The behaviour of the element will be modified.
+  * @param {alwayEnabled} Boolean indicating whether this input should be enabled at all times
+  */
+  var AbstractFuzzyDateInput = Class.create({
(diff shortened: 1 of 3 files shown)


veronikaslc:
The PR is based on the https://github.com/phenotips/phenotips/pull/1944.
For the convenience of codereviewing and rebasing on 2498 commits of master I have squashed commits in one.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,264 +1,401 @@
...
-      return this.daySelector.getElement();
+    onKeyPress : function($super, e) {
+      return $super(e, /[0-9]/, 2, true);
     },
 
-    getAvailableDays : function () {
-      var year = this.yearSelector.getSelectedValue() * 1;
-      var month = this.monthSelector.getSelectedValue() * 1;
+    onDateChange : function($super) {
+      $super();
```
allasm: We need to make sure this plays well with different date formats. I see the logic to make it `alwaysEnabled`, but somehow this is turned off when `yearInput.isFuzzy`. Maybe that is ok, need to think about that.

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -1,264 +1,401 @@
...
+    insertInputs : function(){
+      this.inputFormat.toLocaleLowerCase().split("").forEach(function(identifier){
+        this.container.insert(this.inputs[identifier].getElement());
+      }.bind(this));
+    },
+    updateDateFromInput : function() {
+      var dateObject = JSON.parse(this._input.value);
+      if (dateObject["year"] != null) {
+        this.inputs.y.setValue(dateObject["year"]);
+      } else if (dateObject["decade"] != null) {
```
allasm: This is using an outdated date format I think, we no longer have a `decade` field, instead we have a `range` field

A few things need to be updated (here and in a few other places - at least everywhere there is aword "decade") to support current JSON Fuzzy Date format we use in both pedigree and PT

veronikaslc:
postponing the PR, it was decided to make this datepicker a widget

PR #2506, opened 2017-12-04
PT-3518: Added functionality for displaying additional tabs

a-gvo:


components/family-studies/ui/src/main/resources/PhenoTips/FamilySheet.xml
```diff
@@ -61,6 +61,36 @@ $xwiki.jsfx.use('js/scriptaculous/dragdrop.js')##
...
+## Tabs to be displayed after the family tab
+#set ($tabs = [])
+(% class="hidden" %)(((
+#foreach ($e in $services.uix.getExtensions('org.phenotips.entitySheet.tab', {'sortByParameter' : 'order'}))
+  #set ($extension_disabled = ($e.getParameters().get('enabled') == 'false'))
+  #if (!$extension_disabled)
+    #set ($label = $!{e.getParameters().get('label')})
+    #set ($href = $!{e.getParameters().get('href')})
+    #set ($cssClass = $!{e.getParameters().get('cssClass')})
+    #set ($container = $!{e.getParameters().get('container')})
```
allasm: This `container` is only used to detect which of the tabs is currently being displayed, and so could be named a bit better (or a different current tab detection system can be used - e.g. based on the page URL?). But I can't come up with a better name and not sure what other method is easy to implement, so this is minor and OK

a-gvo:
Addressed code review comments made by @sdumitriu during stand up. This is addressed in a separate commit; @allasm please review again.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -4548,4 +4675,150 @@ input[type="checkbox"] + .LV_validation_message {
...
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>#template('colorThemeInit.vm')
+
+/**********************************************************************************************
+* Tabs css.
```
allasm: I think CSS (and JavaScript) should stay in `Widgets.xml` (or some other global file) instead of being moved to `PatientSheetCode`, since CSS+JS applies to any place which may have tabs, which currently is any entity. It may happen to work for now since Family pages include `PatientSheetCode` (do they?), but this is not generic. Maybe we should create a new `Entity.xml` or something like that to hold UI/CSS/JS stuff which should work for all entities?

components/skin/ui/src/main/resources/PhenoTips/DocumentMeta.xml
```diff
@@ -37,8 +37,30 @@
...
+        &lt;/div&gt;
+      #end
+    #end
+  #end
+#end
+{{/velocity}}
+
+{{velocity}}{{html clean=false wiki=false}}
 #if ($doc.getObject('PhenoTips.PatientClass') || $doc.getObject('PhenoTips.FamilyClass'))
+  #__buildTabs('org.phenotips.entitySheet.tab')
```
allasm: I don't think we can do it now, but just wondering if there is a way to detect if XXXXClass is a class representing an entity (instead of hard-coding PatientClass + FamilyClass)? (@sdumitriu?)

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
a-gvo: Fixed.

components/skin/ui/src/main/resources/PhenoTips/DocumentMeta.xml
marta-: https://github.com/phenotips/phenotips/blob/master/components/entities/api/src/main/java/org/phenotips/entities/script/PrimaryEntityResolverScriptService.java#L63

a-gvo: Fixed.

allasm:
I think you forgot to remove the code that you put into `Entity` from `PatientSheetCode`?

a-gvo:
@allasm I removed it from the instance when I was testing it, and forgot to do it in my IDE. Fixed, thanks!

a-gvo:
Rebased with latest master

components/patient-data/ui/src/main/resources/PhenoTips/Entities.xml
```diff
@@ -0,0 +1,314 @@
...
+    <property>
+      <code>#template('colorThemeInit.vm')
+
+/**********************************************************************************************
+* Tabs css.
+*/
+#document-head .title-tab {
+  display: inline-block;
+  vertical-align: top;
+  line-height: 0;
```
sdumitriu: Add `cursor: default;`

components/patient-data/ui/src/main/resources/PhenoTips/Entities.xml
```diff
@@ -0,0 +1,314 @@
...
+}
+
+#document-head .title-tab h1 {
+  text-transform: none !important;
+}
+
+#document-head .title-tab:not(.active) {
```
sdumitriu: Too much duplication. Move the code for styling the title tab from `PatientSheetCode` into `Skin`, make it apply not to `.space-data #document-head h1` but to `#document-head .title-tab`, and only the colors should be changed for an inactive tab.

components/patient-data/ui/src/main/resources/PhenoTips/Entities.xml
```diff
@@ -0,0 +1,314 @@
...
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>#template('colorThemeInit.vm')
```
sdumitriu: Move this into `PhenoTips.Skin`.

a-gvo: Hm. There is a document-title on the home page that will be affected.

a-gvo:
@sdumitriu Code review comments addressed.

PR #2509, opened 2017-12-07
PT-3515: Add ability to show external identifiers in place of internal ones in the Patient and Family record forms

veronikaslc:


sdumitriu:
Please split into two pull requests.

PR #2511, opened 2017-12-07
PT-3529: New API for providing metadata about entities

sdumitriu:


components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PatientFamilyMetadataProvider.java
```diff
@@ -0,0 +1,60 @@
...
+public class PatientFamilyMetadataProvider implements PrimaryEntityMetadataProvider
+{
+    @Inject
+    private FamilyRepository familyRepository;
+
+    @Override
+    public Map<String, Object> provideMetadata(PrimaryEntity entity)
+    {
+        if (entity instanceof Patient) {
+            Family f = this.familyRepository.getFamilyForPatient((Patient) entity);
```
marta-: Can you remind me what happens if the user can access the patient but not its family? Does the family repository return null here? Do we want to show this metadatum if the family is inaccessible or not? IIRC, in the UIwe say the patient has a pedigree but the user cannot see it.

allasm: I think we have an assumption that if a user can at least view a patient the user should be able to view the family as well? This can be broken for technical reasons, but isnt this how things are supposes to work?

marta-: @allasm https://phenotips.atlassian.net/browse/PT-3427 describes a scenario where one can access the patient but not the family. This is how things work now. Do you think the  behavior should be changed?

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsSuggestionsResourceImpl.java
```diff
@@ -239,14 +233,6 @@ private void appentPatientXML(Patient patient, StringBuilder xmlResult, boolean
 
         String description = getDescription(patient);
 
-        // Add family info
-        Family family = this.familyRepository.getFamilyForPatient(patient);
```
veronikaslc: We do not include family in XML response?

sdumitriu: Hm, looks like I forgot to include something in the commit...

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PatientFamilyMetadataProvider.java
sdumitriu: With the current code the family ID is exposed, but no other information about the family. I think that's OK, isn't it? Without disclosing it, the following behavior will break:

The user edits another family and wants to link a patient to a node in this pedigree. When searching for that patient, the response also includes whether it is part of another family already. If it is, then it is not allowed to be added in the new family. So, if we don't disclose that it is part of a family already, we would allow moving the patient to a new family, which means removing it from the other family, which is something the current user should not be able to do.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientsSuggestionsResourceImpl.java
sdumitriu: Ah no, I remembered I did it intentionally. The suggest doesn't care about extra attributes, and having the family in the label isn't required but would make the code messier.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PatientFamilyMetadataProvider.java
marta-: Okay.

PR #2519, opened 2017-12-19
PT-3515: Add ability to show external identifiers in place of internal ones in the Patient and Family record forms

veronikaslc:


sashaandjic:
@veronikaslc what is the best way to test this PR?

veronikaslc:
@sashaandjic Go to Administration->Data formats  and check checkbox "Show external identifiers (when available) instead of internal ones in the Patient and Family records". Then create a patient and a family with external IDs and save. In the top you should see the external ID in place of internal.
image

sashaandjic:
Since external id are case sensitive, we should not do "all caps" on tabs. WDYT?

veronikaslc:
codereview comments addressed

components/base-war/src/main/webapp/templates/macros.vm
```diff
@@ -2742,3 +2742,20 @@ Recursive title display detected!##
   #set ($sentence3 = 'Click this box to get technical details.')
   #displayException("${sentence1} ${sentence2} ${sentence3}", $exception)
 #end
+
+###
+## Display external ID page title in case it is configured to be shown instead of internal ID
+###
+### @param defaultTitle The default page title to format for displaying
+#macro(pageExternalTitle $defaultTitle)
```
sdumitriu: Move this macro in `PatientSheetMacros`.

components/base-war/src/main/webapp/templates/macros.vm
```diff
@@ -2742,3 +2742,20 @@ Recursive title display detected!##
...
+#set($config = $xwiki.getDocument('data.WebHome').getObject('PhenoTips.DBConfigurationClass'))
+#set($pageExternalTitle = $defaultTitle)
+#if($config && "$!{config.getProperty('hideInternalIds').value}" == '1' && "$!{doc.getValue('external_id')}" != '')
+  #set($ext_id = "$!{doc.getValue('external_id')}")
+  #if("$!{config.getProperty('externalIdFormat').value}" == 'XXX-XXX-XXX')
+    #set($ext_id = $ext_id.replaceAll("(.{3})(?!$)", "$1-"))
```
sdumitriu: Use this instead:

    $ext_id.replaceAll("([^-])(?=([^-]{3})+(?=(-|$)))", "$1-")

Replace any non-dash character that is followed by 3 other non-dash characters then by either dash or end with that character and a dash.

components/base-war/src/main/webapp/skins/phenotips/editinline.vm
```diff
@@ -34,7 +34,7 @@ $xwiki.ssfx.use('uicomponents/widgets/fullScreen.css', true)##
 ## ----------------------------
 ## Display the title
 ## ----------------------------
-<div id="document-title"><h1>$titleToDisplay</h1></div>
+<div id="document-title"><h1>#pageExternalTitle($titleToDisplay)</h1></div>
```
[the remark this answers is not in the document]
veronikaslc: @sdumitriu I made this change in last commit but I got this
image

PR #2521, opened 2017-12-19
PT-3517: Use lighter shade of gray for placeholder text displayed in the genetic variants fields

sdumitriu:
This affects the Kiwi and Orange color themes.


sdumitriu:
@sashaandjic Those color themes are not ours, they are inherited from XWiki.

sashaandjic:
+1

PR #2523, opened 2017-12-20
PT-3536: Patient/Family search widget does not work for guest users

sdumitriu:


allasm:
Just curious how adding new files with no references to them fixed the problem? If those were missing from our XWiki installation then why did stuff work for other users?

sdumitriu:
They are present in XWiki. If you look at the commits, first I added the upstream version of the files, then I fixed the problem by adding rights letting guests access them.

PR #2527, opened 2017-12-21
PT-2391: Clicking N/A on a dropdown prenatal phenotype doesn't do the same on the duplicate.

veronikaslc:
Excluding YesNoNA pickers with values "0" and "1" from HPO terms syncing

### DIFF
--- a/components/widgets/ui/src/main/resources/PhenoTips/YesNoNAPicker.xml
+++ b/components/widgets/ui/src/main/resources/PhenoTips/YesNoNAPicker.xml
@@ -277,7 +277,7 @@ $xwiki.ssx.use('PhenoTips.YesNoNAPicker')##
     var value = element.value;
     var selected = element.checked;
     var label = element.parentNode;
-    if (value != this.yesInput.value || !selected || !label) {
+    if (value != this.yesInput.value || !selected || !label || !value.startsWith("HP:")) {
       return;
     }
     if (label.hasClassName('na')) {


sdumitriu:
Why this change?

sdumitriu:
What is the reported regression?

allasm:
In the Family History (and I think in Prenatal as well) selecting [Y]es for any phenotype marks ALL  other phenotypes get markes as Yes as well

sashaandjic:
yes, the regression was what @allasm reported. it was introduced with previous change, but now it is corrected.

components/widgets/ui/src/main/resources/PhenoTips/YesNoNAPicker.xml
```diff
@@ -277,7 +277,7 @@ $xwiki.ssx.use('PhenoTips.YesNoNAPicker')##
     var value = element.value;
     var selected = element.checked;
     var label = element.parentNode;
-    if (value != this.yesInput.value || !selected || !label) {
+    if (value != this.yesInput.value || !selected || !label || !value.startsWith("HP:")) {
```
sdumitriu: I don't like this, since it fixes PT-2391 just for the HPO. This condition seems to work, though:

     || (element.name !== '' && element.name !== this.yesInput.name)

Can you re-test it?

veronikaslc: @sdumitriu tested new fix, it fixes both original issue and regression

sashaandjic:
it was better before the last change. now if I change the selection on the duplicate, it will not be mirrored on the original. 
screen shot 2018-01-02 at 4 11 07 pm

PR #2528, opened 2017-12-22
PT-3455: The consanguinity status of a patient isn't set when identifiable from the pedigree

veronikaslc:


sashaandjic:
@veronikaslc does it go from the proband up to 3rd generation (grandparents) in the family tree?

sashaandjic:
I tried 2 scenarios and it didn't work in neither one:
1. create new patient; import the latest (consanguineous) pedigree template; save; the Consanguinity flag was not set
2. create new patient; import template with mom-dad-child; make mom and dad siblings; save; the Consanguinity flag isn't set

veronikaslc:
@sashaandjic Yes, pedigree automatically sets right consanguinity from any generation (even higher than grand-grand parents level, may be it is not right). Issue now should be corrected.

components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js
```diff
@@ -280,18 +280,35 @@ define([
...
             }
-            return PhenotipsJSON.internalToPhenotipsJSON(this.getProperties(v), this.getRawJSONProperties(v));
+
+            var consangr = "N";
+            var parentPartnershipNode = editor.getGraph().getParentRelationship(v);
+            if (parentPartnershipNode && parentPartnershipNode !== null && this._isDisplayedAsConsanguinous(parentPartnershipNode)) {
+                consangr = "Y";
+            }
+
+            return PhenotipsJSON.internalToPhenotipsJSON(this.getProperties(v), this.getRawJSONProperties(v), consangr);
```
sdumitriu: I don't like that consanguinity is passed as a parameter, it's too custom. Can it instead be computed as part of `getProperties`?

allasm: The issue here is that this is not a property of the node/person, but of the relationship and/or the pedigree layout (depending on how to look at this). So we may want to have a more generic parameter than consang., but it shouldn't be in the properties imho, maybe `relationshipProperties`?

veronikaslc: @allasm is the last commit along the lines of your proposed change?

allasm: Yes

sdumitriu: It kind of is and isn't. As Brittney said, we can't really say that a person is consanguineous, but we do have a field in the patient record for marking that the person is descending from a consanguineous relationship, so that is a property of the patient. Since the information is found in the patient record (and the patient JSON), I think it should be part of `getProperties`.

On the other hand, `relationshipProperties` would be a good place to also store the parents' ethnicities.

allasm: Pedigree and PT have different data models (where neither is "better", just one is centered around a single patient, another around the pedigree graph), in pedigree consanguinity is (in most cases) derived from pedigree structure, possibly (only) because of connections far away from either the person or the person-producing relationship. So in the context of pedigree, it does not make sense to have it in person properties - that is, until the moment we are forced to put it into PT-compatible JSON.

So having something like `relationshipProperties` may be the compromise solution in this case - in the future it can also be also useful to deal with parent ethnicities (as Sergiu suggested) which are not attached to a specific parent (as is the case for PT patients)

sashaandjic:
Problem found: once consanguinity is detected, the "Other affected relatives" is also set.
screen shot 2018-01-02 at 2 42 40 pm

I would also suggest prompting user to confirm if they want to mark family as consanguineous.

Consanguinity was only detected when proband parents are siblings. For grandparents it did not work. Also, once consanguinity is set (i.e. by mistake), removing consanguineous relationship will not reset it (this may be ok for now as it may turn out into can of worms).

veronikaslc:
@sashaandjic just tested a fresh build for grandparents, worked
image

PR #2532, opened 2018-01-05
PT-870: Add a link to browse all owned patient records on the "my patients" gadget

sdumitriu:


sdumitriu:
Can't add browse owned families button yet, depends on adding owner filter.

PR #2539, opened 2018-01-09
PT-2446: Consents administration: don't allow setting a consent as required for certain fields if it is already mandatory

veronikaslc:
Disable entering specific patient information when mandatory status of consent is set to 'yes'

### DIFF
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
@@ -707,7 +707,7 @@ table.extradata-list.withLabel td.collapsed * {
   height: 1px;
   overflow: hidden;
 }
-table input[type=text] {
+table.extradata-list.withLabel input[type=text] {
   min-width: 40em;
 }
 form .buttons {
--- a/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
+++ b/components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
@@ -511,22 +511,27 @@ $xwiki.jsx.use($displayDocName)##
     }
   };
 
-  var __consent_enableRequired = function(trigger, toggleTarget) {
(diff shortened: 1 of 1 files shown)


components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
```diff
@@ -511,7 +511,24 @@ $xwiki.jsx.use($displayDocName)##
     }
   };
 
+  var __consent_enableRequired = function(trigger, toggleTarget) {
+    if (trigger.selectedIndex &gt;= 0 &amp;&amp; trigger.options[trigger.selectedIndex].value == '1') {
```
sdumitriu: This should be reversed: if the consent is mandatory it affects all fields, so it doesn't make sense to specify which fields are required.

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
```diff
@@ -511,7 +511,24 @@ $xwiki.jsx.use($displayDocName)##
...
+      toggleTarget.disabled = false;
+    } else {
+      toggleTarget.disabled = true;
+    }
+  };
+
   var init = function() {
+    $$('select[name$="required"]').each(function(elt) {
+      var toggleTarget = elt.up('tr') &amp;&amp; elt.up('tr').down('td.affectsFields select');
+      if (!toggleTarget) return;
```
sdumitriu: Codestyle: please use `{}` even for single line blocks, just as we do in Java.

    if (!toggleTarget) {
      return;
    }

components/patient-data/ui/src/main/resources/PhenoTips/PatientConsentConfiguration.xml
```diff
@@ -685,7 +707,9 @@ table.extradata-list.withLabel td.collapsed * {
   height: 1px;
   overflow: hidden;
 }
-
+table input[type=text] {
```
sdumitriu: This is too generic, and may affect other tables. All the other rules apply to `table.extradata-list.withLabel`, so the same should be used here.

PR #2542, opened 2018-01-17
PT-3554: Create a script that would enable upload of vcf file to the given patient

veronikaslc:
Python 3.6 script to upload the VCF file as an attachment for the given patient.

### DIFF
--- a/resources/utilities/upload_vcf.py
+++ b/resources/utilities/upload_vcf.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python
+
+"""
+Usage: $0 PATIENT_ID VCF CREDENTIALS_FILE SERVER_NAME
+
+Upload the VCF file as an attachment for the given patient. For example:
+
+    $0 P0000102 path/to/file.vcf path/to/credentials.txt server_name
+
+credentials.txt should have one line with the form:
+
+    username:password
+"""
+
+from __future__ import with_statement
+
+import sys
+import os
+import logging
+import json
+
+try:
+    from httplib import HTTPConnection
+except ImportError:
+    from http.client import HTTPSConnection
+try:
+    from urllib import urlencode
+except ImportError:
+    from urllib.parse import urlencode
+from base64 import b64encode
+import xml.etree.ElementTree as ET
+
+# Set globally after parsing arguments
+CREDENTIALS = None
+SERVER = None
+
+def load_credentials(filename):
+    global CREDENTIALS
+    logging.info('Using credentials in file: {0}'.format(filename))
+    with open(filename) as ifp:
+        CREDENTIALS = ifp.read().strip()
+
+class Request(object):
+    def __init__(self, method, resource, content_type='text/plain', **kwargs):
+        auth = b64encode(CREDENTIALS.encode()).decode()
+        self._headers = {
+            'Authorization': 'Basic {0}'.format(auth),
+            'Content-Type': content_type,
+            'Accept': '*/*'
+            }
+        self._method = method
+        self._href = resource
+        self._kwargs = kwargs
+        self._conn = None
+
+    def __enter__(self):
+        self._conn = HTTPSConnection(SERVER)
+        #self._conn.set_debuglevel(1)
+        self._conn.request(self._method, self._href, headers=self._headers, **self._kwargs)
+        response = self._conn.getresponse()
+        return response
+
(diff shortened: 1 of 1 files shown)


resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env python
+
+"""
+Usage: $0 PATIENT_ID VCF CREDENTIALS_FILE SERVER_NAME
+
+Upload the VCF file as an attachment for the given patient. For example:
+
+    $0 P0000102 path/to/file.vcf path/to/credentials.txt server_name
```
sdumitriu: 1. There's no mention of the port, but this actually supports specifying a port. Use two examples:

    $0 P0000102 path/to/file.vcf path/to/credentials.txt phenomecentral.org
    $0 P0000102 path/to/file.vcf path/to/credentials.txt localhost:8080

2. This doesn't work when PhenoTips is not deployed as the root webapp. For example, in the standalone distribution edit `jetty/contexts/phenotips.xml` and change to: `<Set name="contextPath">/pt</Set>`. Either add another optional parameter for the context path, or support something like `localhost:8080/pt`.

resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
...
+            self._conn.close()
+
+def get_consent(record_id):
+    logging.info('Loading Consents for record: {0}'.format(record_id))
+    href = '/rest/patients/{0}/consents/'.format(record_id)
+    result = 'no'
+    with Request('GET', href) as response:
+
+        assert response.status == 200, \
+            'ERROR getting PhenoTips.VCF object for record: {0}'.format(response.headers)
```
sdumitriu: Wrong message?

resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
...
+        return response
+
+    def __exit__(self, *args, **kwargs):
+        if self._conn:
+            self._conn.close()
+
+def get_consent(record_id):
+    logging.info('Loading Consents for record: {0}'.format(record_id))
+    href = '/rest/patients/{0}/consents/'.format(record_id)
+    result = 'no'
```
sdumitriu: This requires that the genetic consent is enabled. Change this to "yes" so that it still works in the absence of the genetic consent.

resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
...
+            'Content-Type': content_type,
+            'Accept': '*/*'
+            }
+        self._method = method
+        self._href = resource
+        self._kwargs = kwargs
+        self._conn = None
+
+    def __enter__(self):
+        self._conn = HTTPSConnection(SERVER)
```
sdumitriu: This only works for https, not for unencrypted http. Is there a way to make it transparently work for both protocols?

resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
...
+
+from __future__ import with_statement
+
+import sys
+import os
+import logging
+import json
+
+try:
+    from httplib import HTTPConnection
```
sdumitriu: This works in python2, but not in python3. However, if the script runs in python2, the imported class is HTTPConnection, while in python2 HTTPS_S_Connection is imported. The code below only works with HTTPSConnection, so under python2 it will fail with `NameError: global name 'HTTPSConnection' is not defined`.

resources/utilities/upload_vcf.py
```diff
@@ -0,0 +1,222 @@
...
+    if consent_granted == 'no':
+        logging.info('Record {0} has not granted genetic consent, can not upload VCF'.format(record_id))
+        return
+
+    vcf_object = get_vcf_object(record_id)
+
+    vcf_basename = os.path.basename(vcf_filename)
+    if vcf_object and vcf_object.get('headline', '') == vcf_basename:
+        # Filename already appears to be properly linked to record, skip
+        logging.info('Record {0} already has linked VCF file: {1}'.format(record_id, vcf_basename))
```
sdumitriu: How to re-upload a vcf file, which may be modified? I would like a `--force-reupload` option that will still upload the file, even if it has the same name.

veronikaslc:
@sdumitriu all coderview comments addressed

PR #2546, opened 2018-01-19
[misc] Removed extra blank line on top of the Browse Families table

veronikaslc:
Removed whitespace that was generating blank paragraph


veronikaslc:
image

PR #2549, opened 2018-02-02
CORE-8: Patient data REST for labeled external IDs

teyden:
It behaves like the REST API for `external_id` with the exception of the `PUT` and `PATCH` methods. If there is no patient record that is identified by the given `label` and corresponding id `value`, a `404 Not Found` error is returned instead of creating a new patient record. This seems like better behaviour, is more defensive coding (i.e. less likely to break client applications), and allows for better integration with other systems. I had discussed this with Orion and Sebastian, and it was agreed that this behaviour would be more ideal. I would also suggest that these methods for `external_id`s also behave this way.

### DIFF
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
@@ -65,7 +65,7 @@ public interface PatientByLabeledExternalIdentifierResource
      * change is performed and an error is returned. If the indicated patient record doesn't exist, however a valid
      * JSON is provided with the query parameter "create" set to true, then a new patient record is created with the
      * provided data; otherwise no change is performed and an error is returned. If multiple records exist with the
-     * same givenidentifier for the given label, no change is performed, and a list of links to each such record is
+     * same given identifier for the given label, no change is performed, and a list of links to each such record is
      * returned. If a field is set in the patient record, but missing in the JSON, then that field is not changed.
      *
      * @param json the JSON representation of the new patient to add
@@ -80,7 +80,7 @@ public interface PatientByLabeledExternalIdentifierResource
     @RequiredAccess("edit")
     Response updatePatient(String json, @PathParam("label") String label, @PathParam("id") String id,
         @QueryParam("policy") @DefaultValue("update") String policy,
-        @QueryParam("create") @DefaultValue("false") String create);
+        @QueryParam("create") @DefaultValue("false") boolean create);
 
     /**
      * Update a patient record, identified by an arbitrary label and its corresponding identifier value, from its JSON
@@ -102,7 +102,7 @@ public interface PatientByLabeledExternalIdentifierResource
     @Consumes(MediaType.APPLICATION_JSON)
     @RequiredAccess("edit")
     Response patchPatient(String json, @PathParam("label") String label, @PathParam("id") String id,
-        @QueryParam("create") @DefaultValue("false") String create);
+        @QueryParam("create") @DefaultValue("false") boolean create);
 
     /**
      * Delete a patient record, identified by an arbitrary label and its corresponding identifier value. If the
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
@@ -62,41 +62,47 @@ public interface PatientByLabeledExternalIdentifierResource
     /**
      * Update a patient record, identified by an arbitrary label and its corresponding identifier value, from its JSON
      * representation. If the user sending the request doesn't have the right to edit the target patient record, no
-     * change is performed and an error is returned. If the indicated patient record doesn't exist, and a valid JSON is
-     * provided, no change is performed and an error is returned. If multiple records exist with the same given
-     * identifier for the given label, no change is performed, and a list of links to each such record is returned. If
-     * a field is set in the patient record, but missing in the JSON, then that field is not changed.
+     * change is performed and an error is returned. If the indicated patient record doesn't exist, however a valid
+     * JSON is provided with the query parameter "create" set to true, then a new patient record is created with the
+     * provided data; otherwise no change is performed and an error is returned. If multiple records exist with the
+     * same givenidentifier for the given label, no change is performed, and a list of links to each such record is
+     * returned. If a field is set in the patient record, but missing in the JSON, then that field is not changed.
      *
      * @param json the JSON representation of the new patient to add
      * @param label an arbitrary label, either exists in all patients, or exists in at least one patient
      * @param id the patient's given external identifier for the given label
      * @param policy the policy according to which patient data should be written
+     * @param create whether a patient record should be created given none exist for the label and id provided
      * @return a status message
      */
     @PUT
     @Consumes(MediaType.APPLICATION_JSON)
     @RequiredAccess("edit")
     Response updatePatient(String json, @PathParam("label") String label, @PathParam("id") String id,
-        @QueryParam("policy") @DefaultValue("update") String policy);
+        @QueryParam("policy") @DefaultValue("update") String policy,
+        @QueryParam("create") @DefaultValue("false") String create);
 
     /**
      * Update a patient record, identified by an arbitrary label and its corresponding identifier value, from its JSON
      * representation. Existing patient data is merged with the provided JSON representation, if possible. If the user
      * sending the request doesn't have the right to edit the target patient record, no change is performed and an
-     * error is returned. If the indicated patient record doesn't exist, no change is performed and an error is
-     * returned. If multiple records exist with the same given identifier for the given label, no change is performed,
-     * and a list of links to each such record is returned. If a field is set in the patient record, but missing in
-     * the JSON, then that field is not changed.
+     * error is returned. If the indicated patient record doesn't exist, however a valid JSON is provided with the
+     * query parameter "create" set to true, then a new patient record is created with the provided data; otherwise no
+     * change is performed and an error is returned. If multiple records exist with the same given identifier for the
+     * given label, no change is performed, and a list of links to each such record is returned. If a field is set in
+     * the patient record, but missing in the JSON, then that field is not changed.
      *
      * @param json the JSON representation of the new patient to add
      * @param label an arbitrary label, either exists in all patients, or exists in at least one patient
      * @param id the patient's given external identifier for the given label
+     * @param create whether a patient record should be created given none exist for the label and id provided
      * @return a status message
      */
     @PATCH
     @Consumes(MediaType.APPLICATION_JSON)
     @RequiredAccess("edit")
-    Response patchPatient(String json, @PathParam("label") String label, @PathParam("id") String id);
+    Response patchPatient(String json, @PathParam("label") String label, @PathParam("id") String id,
+        @QueryParam("create") @DefaultValue("false") String create);
 
     /**
      * Delete a patient record, identified by an arbitrary label and its corresponding identifier value. If the
--- a/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
+++ b/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
@@ -1,3 +1,20 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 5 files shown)


components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return a status message
+     */
+    @PATCH
+    @Consumes(MediaType.APPLICATION_JSON)
+    @RequiredAccess("edit")
+    Response patchPatient(String json, @PathParam("label") String label, @PathParam("id") String id);
+
+    /**
+     * Delete a patient record, identified by an arbitrary label and its corresponding identifier value. If the
+     * indicated patient record doesn't exist, or if the user sending the request doesn't have the right to edit the
```
sdumitriu: `have the right to delete`, not `edit`.

components/vocabularies/rest/src/main/java/org/phenotips/vocabularies/rest/VocabularyTermResource.java
```diff
@@ -45,7 +45,7 @@
      * Retrieves a JSON representation of the {@link org.phenotips.vocabulary.VocabularyTerm} by searching the specified
      * vocabulary.
      *
-     * @param vocabularyId the vocabulary identifier, which is also used as a prefix in every term identifier from that
+     * @param vocabularyId the vocabulary identifier, whic  h is also used as a prefix in every term identifier from that
```
sdumitriu: Bad change, remove from this pull request.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
```diff
@@ -0,0 +1,98 @@
...
+     * target patient record, no change is performed and an error is returned. If multiple records exist with the same
+     * given identifier for the given label, no change is performed, and a list of links to each such record is
+     * returned.
+     *
+     * @param label an arbitrary label, either exists in all patients, or exists in at least one patient
+     * @param id the patient's given external identifier for the given label
+     * @return a status message
+     */
+    @DELETE
+    @RequiredAccess("edit")
```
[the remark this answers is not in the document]
teyden: Hmm, should it be `@RequiredAccess("delete")`, then? 

The interface `PatientByExternalIdResource` has `edit` as the annotation value as well.. https://github.com/phenotips/phenotips/blob/master/components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByExternalIdResource.java#L109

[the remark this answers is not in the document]
sdumitriu: Theoretically yes, but practically the actual rights supported at the moment are just `view` and `edit`, so leave it as it is here. But do change the javadoc above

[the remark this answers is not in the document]
teyden: Alright, no problem. I've made the changes, just about to commit them.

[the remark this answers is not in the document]
teyden: Would you prefer that I squash all the commits into one, or leave it as is with the code review commits separate?

teyden:
About the behaviour of PUT, after some discussion on the Gene42 side, we think that in the scenario a patient record exists with the given label/value pair, it should return an error **by default** but could create a new patient record with a query/url parameter switched on. What are your thoughts on that @sdumitriu?

sdumitriu:
+1 for a parameter. The default for this resource should be to fail.

teyden:
Please review latest changes @sdumitriu, thanks.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
```diff
@@ -62,41 +62,47 @@
...
      * Update a patient record, identified by an arbitrary label and its corresponding identifier value, from its JSON
      * representation. If the user sending the request doesn't have the right to edit the target patient record, no
-     * change is performed and an error is returned. If the indicated patient record doesn't exist, and a valid JSON is
-     * provided, no change is performed and an error is returned. If multiple records exist with the same given
-     * identifier for the given label, no change is performed, and a list of links to each such record is returned. If
-     * a field is set in the patient record, but missing in the JSON, then that field is not changed.
+     * change is performed and an error is returned. If the indicated patient record doesn't exist, however a valid
+     * JSON is provided with the query parameter "create" set to true, then a new patient record is created with the
+     * provided data; otherwise no change is performed and an error is returned. If multiple records exist with the
+     * same givenidentifier for the given label, no change is performed, and a list of links to each such record is
```
sdumitriu: `givenidentifier` needs some space.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
```diff
@@ -62,41 +62,47 @@
...
+     * @param create whether a patient record should be created given none exist for the label and id provided
      * @return a status message
      */
     @PUT
     @Consumes(MediaType.APPLICATION_JSON)
     @RequiredAccess("edit")
     Response updatePatient(String json, @PathParam("label") String label, @PathParam("id") String id,
-        @QueryParam("policy") @DefaultValue("update") String policy);
+        @QueryParam("policy") @DefaultValue("update") String policy,
+        @QueryParam("create") @DefaultValue("false") String create);
```
sdumitriu: Would declaring it as a `boolean` work?

teyden: Potentially, looks like it could: https://stackoverflow.com/questions/35625155/what-are-the-default-values-for-queryparam-when-defaultvalue-is-not-specified

components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
```diff
@@ -102,7 +102,7 @@ Response updatePatient(String json, @PathParam("label") String label, @PathParam
     @Consumes(MediaType.APPLICATION_JSON)
     @RequiredAccess("edit")
     Response patchPatient(String json, @PathParam("label") String label, @PathParam("id") String id,
-        @QueryParam("create") @DefaultValue("false") String create);
+        @QueryParam("create") @DefaultValue("false") boolean create);
```
teyden: Technically specifying `@DefaultValue("false")` is not necessary for booleans since its value by default if uninitialized is false, but I think it's good to be explicit about the default value.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/PatientByLabeledExternalIdentifierResource.java
teyden: I changed this in the latest commit. I commented inline, explaining use of `@DefaultValue("false")` for booleans. It's technically not necessary for `boolean` since its value by default if uninitialized is false. I think it's good to be explicit about the default value.

teyden:
Writing tests for this this week.

sdumitriu:
Redone in #2578

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImplTest.java
```diff
@@ -0,0 +1,627 @@
...
+    }
+
+    @Test
+    public void checkForMultipleRecordsLogsWhenThrowsQueryException() throws QueryException
+    {
+        doThrow(QueryException.class).when(this.qm)
+            .createQuery("select doc.name from Document doc, doc.object(PhenoTips.LabeledIdentifierClass) obj "
+                         + "where obj.label = :label and obj.value = :value", Query.XWQL);
+
+        when(getPatientByLabelAndEid(LABEL, EID)).thenReturn(null);
```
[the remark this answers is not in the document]
teyden: Ah, (7f6c5d9) this commit is supposed to be rebased. That was an initial commit after copying over from the test for external ID's and quickly replacing code with the general structure of what should go there without actually running the test. I'm still working on the tests.

components/patient-data/rest/src/test/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImplTest.java
```diff
@@ -0,0 +1,627 @@
...
+     * For a given label and its corresponding id value (fields on {@code LabeledIdentifierClass} objects), query
+     * patient documents.
+     *
+     * @param label the name of the label
+     * @param id the id value for the label
+     * @return an empty list if no patients have the label identifier and its corresponding id value, else a list of
+     *         patient internal identifiers
+     * @throws Exception if there is any error during querying
+     */
+    private Patient getPatientByLabelAndEid(String label, String id)
```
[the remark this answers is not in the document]
teyden: Nope, it has already been removed.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImpl.java
```diff
@@ -0,0 +1,437 @@
...
+     * @param label the name of the label
+     * @return true if the label exists, else false
+     * @throws Exception if there is any error during querying
+     */
+    private boolean isLabelConfiguredByAdmin(String label) throws Exception
+    {
+        Query q = null;
+        try {
+            q = this.qm.createQuery(
+                "select obj.label from Document doc, doc.object(PhenoTips.LabeledIdentifierSettings) obj "
```
sdumitriu: The configuration should only be read from `XWiki.XWikiPreferences`, not any document.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImpl.java
```diff
@@ -0,0 +1,437 @@
...
+                + "where obj.label = :label", Query.XWQL);
+            q.bindValue(KEY_LABEL, label);
+            List<String> results = q.execute();
+            return !results.isEmpty();
+        } catch (QueryException ex) {
+            this.logger.warn("Failed to query LabeledIdentifierSettings object: {}", ex.getMessage(), ex);
+            throw new QueryException(ex.getMessage(), q, ex);
```
sdumitriu: Problems when checking if identifiers are allowed should not be propagated to the user; assume sane default values if the configuration is bad.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImpl.java
```diff
@@ -0,0 +1,437 @@
...
+        this.logger.debug("Creating patient record with label [{}] and corresponding external ID [{}]", label, id);
+        try {
+            final User currentUser = this.users.getCurrentUser();
+            if (!this.access.hasAccess(currentUser, Right.EDIT,
+                this.currentResolver.resolve(Patient.DEFAULT_DATA_SPACE, EntityType.SPACE))) {
+                this.logger.error("Edit access denied to user [{}].", currentUser);
+                return Response.status(Response.Status.FORBIDDEN).build();
+            }
+            final Patient patient = this.repository.create();
+            return returnUpdatePatientResponse(patient, label, id, jsonInput, true, PatientWritePolicy.UPDATE);
```
sdumitriu: If a new patient is created, a CREATED response should be returned with the canonical location for the new patient included.

components/patient-data/rest/src/main/java/org/phenotips/data/rest/internal/DefaultPatientByLabeledExternalIdentifierResourceImpl.java
```diff
@@ -0,0 +1,437 @@
...
+     * @param policy the policy according to which patient data should be written
+     * @return a {@link Response} with no content if successful, an error code otherwise
+     */
+    private Response returnUpdatePatientResponse(final Patient patient, final String label, final String id,
+        final JSONObject jsonInput, boolean wasCreated, PatientWritePolicy policy)
+    {
+        if (hasInternalIdConflict(jsonInput, patient)) {
+            return Response.status(Response.Status.CONFLICT).build();
+        }
+        if (wasCreated && !jsonInput.has(KEY_LABELED_EIDS)) {
```
sdumitriu: Specified label should always be put in the input JSON, even if other labeled identifiers are set. Remove the check for `jsonInput.has(KEY_LABELED_EIDS)`.

PR #2552, opened 2018-02-07
PT-3348: Typed gene name doesn't get replaced with the proper Ensembl ID after displaying gene information

veronikaslc:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -436,6 +436,7 @@
     // If no match found in suggest widget, perform a server-side search for the exact match.
     // This will most likely never happen.
     if (!matchSubmitted) {
+      var event = Event.fire(this.fld, "ms:suggest:not-selected");
```
sdumitriu: You don't seem to use `event`, so just remove `var event = `

PR #2557, opened 2018-02-14
PT-3580: Lookup fields in advanced filters and export/push dialog need to have a uniform look

veronikaslc:
- Added user icon in the beginning of the input field

### DIFF
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
@@ -125,3 +125,9 @@
 .filter-match-count {
   margin-left: 0.5em
 }
+input.suggestUsersAndGroups {
+  background-image: url(/resources/icons/datamodel/users.png);
+  background-position: .3em center;
+  background-repeat: no-repeat;
+  padding-left: 2em;
+}
--- a/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
+++ b/components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
@@ -125,9 +125,3 @@
 .filter-match-count {
   margin-left: 0.5em
 }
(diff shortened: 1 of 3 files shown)


components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
```diff
@@ -125,3 +125,9 @@
 .filter-match-count {
   margin-left: 0.5em
 }
+input.suggestUsersAndGroups {
+  background-image: url(/resources/icons/datamodel/users.png);
```
sdumitriu: This doesn't work when PhenoTips is not deployed as the root webapp. Use `$xwiki.getSkinFile('icons/datamodel/users.png')`

components/widgets/ui/src/main/resources/XWiki/DateTimePicker.xml
```diff
@@ -2890,7 +2890,7 @@ var SimpleDateFormat;
       <code>#template('colorThemeInit.vm')
 input.xwiki-date {
   background-image: url("$xwiki.getSkinFile('icons/silk/date.png')");
-  background-position: right center;
+  background-position: .3em center;
```
sdumitriu: Once you enter a date, the date is overlapping the icon and thus is unreadable. Also add a `2em` left padding.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.css
```diff
@@ -125,3 +125,9 @@
 .filter-match-count {
   margin-left: 0.5em
 }
+input.suggestUsersAndGroups {
```
sdumitriu: This doesn't really belong here, move it to `PhenoTips.Widgets`, in the SSX object named `Suggest workgroups`, which should be renamed to `Suggest users, groups and workgroups`.

veronikaslc:

PR #2559, opened 2018-02-15
PT-3552: XSS vulnerability in patient form data

sdumitriu:
Fixed


allasm:
Just curious: it is easy to confirm that the fixes do what they are expected to do, but hard to check that all possible places that need to be fixed (e.g. by adding `$escapetool.xml`) have been fixed - is it possible - in the long term - to develop a more generic approach which will clean the data much earlier so that downstream code does not have to deal with it individually? E.g. a JS snipped included on every page that strips away html/xwiki syntax for every text input? (and I guess a similar piece of code for REST/controllers)?

sdumitriu:
@sashaandjic That problem should now also be fixed

PR #2560, opened 2018-02-15
PT-2238: Exception while saving a patient

sdumitriu:
Temporary fix. Tests will be added soon.

### DIFF
--- a/components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
+++ b/components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
@@ -55,7 +55,7 @@ import com.xpn.xwiki.doc.XWikiDocument;
  * @since 1.4
  */
 @Component
-@Named("document-locking-listener")
+@Named("concurrency-locking")
 @Singleton
 public class LockingListener extends AbstractEventListener
 {
--- a/components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
+++ b/components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
@@ -1,0 +1,75 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+package org.phenotips.locks.internal;
+
+import org.phenotips.locks.DocumentLockManager;
+
+import org.xwiki.bridge.event.ActionExecutedEvent;
+import org.xwiki.bridge.event.ActionExecutingEvent;
+import org.xwiki.bridge.event.ActionExecutionEvent;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.observation.AbstractEventListener;
+import org.xwiki.observation.event.Event;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/TimeoutDocumentLockManager.java
```diff
@@ -0,0 +1,77 @@
...
+                lock = new StampedLock().asWriteLock();
+                this.locks.put(document, lock);
+            } else {
+                lock = this.locks.get(document);
+            }
+        }
+        try {
+            lock.tryLock(10, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            // We don't expect any interruptions
```
allasm: should we add a `logger.error()` here, so that we know that something we did not expect happened?

components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
```diff
@@ -0,0 +1,75 @@
...
+    public LockingListener()
+    {
+        super("concurrency-locking", new ActionExecutingEvent(), new ActionExecutedEvent());
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        String name = ((ActionExecutionEvent) event).getActionName();
+        if (!SUPPORTED_EVENTS.contains(name)) {
```
allasm: Do we now lock on any `get` of any document? While this may look hack-ish, but as discussed in the concurrency channel may be it is better to only lock on document types which are modifiable? (it seems that can be generalized to be any kind of entity and only that?)

danielpgross:
One thing to consider with this approach is how other event listeners will behave. Is the execution order of event handlers deterministic? If yes, what determines the order?
A thread could become blocked for a few seconds or more while waiting for to obtain the lock on a given document. Other event handlers executing before or after that wait could behave in unexpected ways.

components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /** Basic constructor. */
+    public LockingListener()
+    {
+        super("concurrency-locking", new ActionExecutingEvent(), new ActionExecutedEvent());
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
```
danielpgross: Any uncaught exception thrown within this method will be caught [here](https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-7.1.4/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java#L367) by XW core and logged as an error, but will not interrupt the handling of the action. This is important because it means that e.g. a save action will proceed anyway, which could lead to modification of data in unexpected ways.

danielpgross:
Just to be clear to all reviewers (I know @sdumitriu is aware), this approach does **not** solve concurrency problems for REST requests or modifications made via the Java API, either in Velocity code or Java code. That's because those operations don't trigger an `ActionExecutingEvent`.

components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/LockingListener.java
sdumitriu: This is a best-effort improvement, not a perfect guarantee. Parallel requests for updating data may always come, and even though we can technically make sure the requests are not overlapping in the critical section, the data itself may be bad, so actually solving data concurrency is a very hard thing to do (operational transformations are the best solution, and that is a hard and restrictive thing).

And there are no exceptions expected here in the locking listener, the code isn't that complex and the only expected exceptions are already handled.

components/utilities/concurrency-locking/src/main/java/org/phenotips/locks/internal/TimeoutDocumentLockManager.java
```diff
@@ -0,0 +1,82 @@
...
+    {
+        try {
+            final Lock lock = this.locks.computeIfAbsent(document, k -> new StampedLock().asWriteLock());
+            final boolean cleanLock = lock.tryLock(10, TimeUnit.SECONDS);
+            if (!cleanLock) {
+                this.logger.debug("Timed out while waiting for lock on [{}], proceeding anyway", document);
+            }
+        } catch (InterruptedException ex) {
+            // We don't expect any interruptions
+            this.logger.error("Unexpected interruption why waiting for lock: {}", ex.getMessage(), ex);
```
danielpgross: Typo: `why waiting for lock`

PR #2563, opened 2018-02-21
PT-3484: Gene lookup should provide free text option at the bottom as  the other lookups do

veronikaslc:


sashaandjic:
tested in master

PR #2566, opened 2018-02-21
PT-3594: Don't append a free text option if input has an exact match in the suggestion list

veronikaslc:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -1194,8 +1194,15 @@
     // loop throught arr of suggestions
     // creating an XlistItem for each suggestion
     //
+    // at the same time check if any of the suggestions
+    // is the exact case-sentitive match of the input
```
sdumitriu: case-insensitive

veronikaslc: right, that's what I actually meant :)

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -1194,8 +1194,15 @@
...
+    //
+    var exactMatch = false;
     for (var i = 0,len = arr.length; i &lt; len; i++)
     {
+       if (arr[i].value.toLowerCase() == this.fld.value.strip().toLowerCase()) {
```
sdumitriu: This should also look in `arr[i].id`

sashaandjic:
tested in master

PR #2567, opened 2018-02-21
PT-3591: Labeled external identifier fields should be shown in Advanced filter panel only if they are allowed and defined

a-gvo:


sashaandjic:
tested in master

PR #2570, opened 2018-02-23
PT-3598: Remove the space between the menu and the permissions block

a-gvo:


veronikaslc:
Omg, something funny happened. I made exactly same fix in another PR today https://github.com/phenotips/phenotips/pull/2571/files. Then I got email with comment and thought its about my PR, and forced-pushed here. Sorry :)

PR #2576, opened 2018-02-27
PT-3464: Logging out while in edit mode can lock the object for 30 minutes

veronikaslc:


components/base-war/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -1706,18 +1709,19 @@ public void onEvent(Event event, Object source, Object data)
      * Release all of the locks held by the currently logged in user.
      *
      * @param ctx the XWikiContext, used to start the connection and get the user name.
+     * @param userName the optional user name explicitly passed.
      */
-    private void releaseAllLocksForCurrentUser(final XWikiContext ctx)
+    private void releaseAllLocksForCurrentUser(final XWikiContext ctx, String userName)
```
allasm: I would rename it, since it no longer works for "current user"

components/base-war/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -1706,18 +1709,19 @@ public void onEvent(Event event, Object source, Object data)
...
             this.beginTransaction(ctx);
             Session session = this.getSession(ctx);
             final Query query =
                 session.createQuery("delete from XWikiLock as lock where lock.userName=:userName");
             // Using deprecated getUser() because this is how locks are created.
-            // It would be a maintainibility disaster to use different code paths
+            // It would be a maintainability disaster to use different code paths
             // for calculating names when creating and removing.
-            query.setString("userName", ctx.getUser());
+            query.setString("userName", userName);
```
allasm: @sdumitriu is this right to call this when `userName` is `null`? We know this is the way it used to work so probably nothing "bad" happens, but we also know that this code did not work right. So when user is null, maybe no work should be done and we can save an unnecessary `query.execute`?

sdumitriu: It is unneeded, so yes, a null check should make the function return early.

components/base-war/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -0,0 +1,2841 @@
...
+            if (comment != null && comment.length() > 1023) {
+                attachment.setComment(StringUtils.abbreviate(comment, 1023));
+            }
+
+            // The version number must be increased and the date must be set before the attachment meta data is saved.
+            // Changing the version and date after calling session.save()/session.update() "worked" (the altered version
+            // was what Hibernate saved) but only if everything is done in the same transaction and as far as I know it
+            // depended on undefined behavior.
+            // Note that the second condition is required because there are cases when we want the attachment content to
+
```
sdumitriu: Compilation failure, this doesn't exist anymore.

PR #2577, opened 2018-03-01
PT-1645: New patient link in browse all patients page should have cursor pointer; PT-3574: Clearing Referrer and Last Author fields in familie's advanced filters does not automatically refresh the results; PT-3476: Add owner, collaborators and visibility to family directory filters 

veronikaslc:


veronikaslc:
image

veronikaslc:
Pushing to the same branch for convenience of testing small UI changes in the same one build.

components/patient-access-rules/migrations/src/test/java/org/phenotips/data/permissions/internal/R71509PhenoTips3476DataMigrationTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Always executes. When the initial database version is greater than this migrator's, the manager decides not to
+     * run.
+     */
+    @Test
+    public void shouldExecute() throws Exception
+    {
+        Assert.assertTrue(this.mocker.getComponentUnderTest().shouldExecute(new XWikiDBVersion(54689)));
+        Assert.assertTrue(this.mocker.getComponentUnderTest().shouldExecute(new XWikiDBVersion(54693)));
```
sdumitriu: These numbers are copy/pasted, change them to `71500` and `71510`.

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
```diff
@@ -386,7 +386,11 @@ document.observe('xwiki:dom:loading', function() {
   }
 
   document.observe('xwiki:multisuggestpicker:selectionchanged', function() {
-    document.fire('xwiki:livetable:patients:filtersChanged');
+    if (XWiki.currentDocument.space == 'data') {
+      document.fire('xwiki:livetable:patients:filtersChanged');
+    } else {
+      document.fire('xwiki:livetable:families:filtersChanged');
```
sdumitriu: This isn't very extensible... Is there a way to get the table name from somewhere? Is this generic enough:

```
if ($$(".entity-directory")[0]) {
  document.fire('xwiki:livetable:' + $$(".entity-directory")[0].id.replace(/-.*/, '') + ':filtersChanged');
}
```

components/patient-access-rules/migrations/src/test/java/org/phenotips/data/permissions/internal/R71509PhenoTips3476DataMigrationTest.java
allasm: if this component is version 71509 why should it execute for DB version 71510? Should it be `assertFalse` then?

veronikaslc:

components/widgets/resources/src/main/resources/resources/uicomponents/widgets/exportData.js
```diff
@@ -386,7 +386,10 @@ document.observe('xwiki:dom:loading', function() {
   }
 
   document.observe('xwiki:multisuggestpicker:selectionchanged', function() {
-    document.fire('xwiki:livetable:patients:filtersChanged');
+    var tableName = $$('.xwiki-livetable')[0] ? $$('.xwiki-livetable')[0].id.replace(/-.*/, '') : '';
```
sdumitriu: If you get the ID directly from the table, then there's no need for the `replace`.

components/patient-access-rules/migrations/src/test/java/org/phenotips/data/permissions/internal/R71509PhenoTips3476DataMigrationTest.java
sdumitriu: The pruning based on the current version is done someplace else, this method is for a different purpose. It helps with scenarios like:

- this migrator does the same as an older one, so don't run if the other one didn't run yet and will also run in the current batch
- this migrator fixes something broken by another migrator, so only run it the other one didn't run yet

allasm: so in this particular case why should it run when DB version is 71510?

sdumitriu: Because it has no explicit conflict with that version, and the migration manager is the one that already decides that 71509 shouldn't run when the db is already at 71510.

components/patient-access-rules/migrations/src/main/java/org/phenotips/data/permissions/internal/R71509PhenoTips3476DataMigration.java
```diff
@@ -0,0 +1,147 @@
...
+import com.xpn.xwiki.store.migration.hibernate.HibernateDataMigration;
+
+/**
+ * Migration for PhenoTips issue #PT-3476: Add owner, collaborators and visibility to family directory filters. This is
+ * better implemented if all families have a default visibility object private which until now they do not.
+ *
+ * @version $Id$
+ * @since 1.4RM4
+ */
+@Component(roles = { HibernateDataMigration.class })
```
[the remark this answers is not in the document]
sdumitriu: Ah, this is needed when there are tests for the class... That's why the build is failing now.

PR #2588, opened 2018-03-15
PT-3472: Add cancers to the Patient record form

a-gvo:


sashaandjic:
3 issues found:
- No cancers are shown under the PREDEFINED CANCERS
- PREDEFINED CANCERS should be changed to BROWSE CANCERS to match format used in Clinical symptoms and physical finding section
- multiple occurrences of the same cancer should be sorted in ASC order for better presentation

Question: Should all cancers be sorted by age?
cancers section

PR #2590, opened 2018-03-15
PT-3630: PhenoTips doesn't run on Java 9

sdumitriu:
Fixed.

Building still doesn't work because of https://github.com/highsource/maven-jaxb2-plugin/issues/120


buske:
I think `start.bat` needs to be updated with the command line parameter fixes in this PR and #3880.

PR #2591, opened 2018-03-20
Improve performance of "clear details" for phenotype metadata

danielpgross:
Adding `ajax=1` to the request causes XWiki to return an HTTP 204
instead of redirecting to the full patient form


danielpgross:
One problematic behaviour that I've observed so far while testing this is that in some cases when clicking "Clear details" for multiple phenotypes in a row, the request fails with "Failed: Conflict" displayed in the UI. Trying to click "Clear details" for that same phenotype will then result in "Failed: Conflict" again, even if attempted multiple times. If a quick save is performed, "clear details" for that phenotype will work again.
p0001100 phenotype database 2018-03-20 16-42-30

danielpgross:
@sdumitriu, can we cherry-pick this into `stable-1.3.x` so that it can be part of the 1.3.7 release?

sdumitriu:
Done.

PR #2596, opened 2018-03-22
PT-3566: Disable SAVE button and auto-save while saving record is in the process

a-gvo:


danielpgross:
Does this prevent an autosave from happening while a quick save is in progress? Or a quick save from being pressed when an autosave is in progress?

a-gvo:
@danielpgross Autosave should trigger the "xwiki:actions:save" event which, in turn, should disable the save buttons. Autosave may still happen while quick save is in progress though. I might as well change the issue title and add it to this PR?

danielpgross:
Yeah, it would be a major improvement to also prevent autosave from triggering if a quick save is in progress. We just dealt with that exact issue in a custom component for one of our clients.

a-gvo:
There's a bug. Will fix later tonight.

components/skin/ui/src/main/resources/PhenoTips/Skin.xml
```diff
@@ -560,16 +560,30 @@ document.observe('xwiki:suggest:containerCreated', function(event) {
...
         // form elements have been modified already.
         setTimeout(recordFormData, 0);
+        document.observe('xwiki:actions:save', updateSaveState.bind(this, true, false));
         // Record the state of the form again once the form is saved.
-        document.observe('xwiki:document:saved', recordFormData);
-        $$("input[name^='action_']").invoke('observe', 'click', recordFormData);
+        document.observe('xwiki:document:saved', updateSaveState.bind(this, false, true));
+        // Make sure to unlock form saving if save failed.
+        document.observe("xwiki:document:saveFailed", updateSaveState.bind(this, false, false));
+        document.observe("phenotips:scrollValidation:invalid", updateSaveState.bind(this, false, false));
```
sdumitriu: I don't like this dependency. Ideally, the `xwiki:actions:save` event shouldn't be sent at all, but the order of the events is reversed: first the event is sent, then the validation script stops the form submission. So, it seems that there is no other way at the moment to detect that the validation failed and the form submission was stopped.

The whole validation script should be refactored, starting with the upstream `LiveValidation`, but at the moment this is good enough.

PR #2599, opened 2018-03-26
PT-3602: Add rejected gene status

veronikaslc:


components/pedigree/resources/src/main/resources/pedigree/view/rejectedGeneLegend.js
```diff
@@ -1,15 +1,15 @@
 /**
- * Specific version of GeneLegend for rejected genes. It is nbot displayed but used
+ * Specific version of GeneLegend for rejected genes. It is not displayed but used
  * because of the way gene update machinery works (to update gene IDs for gene known only by their symbols)
  *
- * @class RejecetdGeneLegend
+ * @class RejecetedGeneLegend
```
sdumitriu: Still wrong: `Rejeceted` -> `Rejected`

veronikaslc: right, will amend commit

veronikaslc:
@sashaandjic Here are 2 issues fixed in one PR, just for testing convenience. PT-3602 (a big one, affecting a lot) and PT-3639 (a very small one, just rewording). 
PT-3602 is about adding one more `rejected` gene status. To be tested: patient record as well as export, import, push, pedigree (rejected genes are just displayed in pedigree legend) and gene panes.

components/patient-data/ui/src/main/resources/PhenoTips/GeneClass.xml
```diff
@@ -117,7 +117,7 @@
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
-      <values>candidate=Candidate|rejected=Negative|solved=Confirmed causal|carrier=Carrier</values>
+      <values>candidate=Candidate|rejected=Rejected candidate|solved=Confirmed causal|carrier=Carrier|negative=Negative</values>
       <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
```
sdumitriu: I wonder if this is the correct order in which the options are listed.

sashaandjic: We should ask Brittney...

sashaandjic:
PT-3643 - tested
PT-3639 - removed "Global" from both Age of onset and Mode of inheritance, we need "i" button for the "Mode of inheritance" (minor; my bad in the subtask description)

sashaandjic:
PT-3602: push isn't working properly when pushing instance has new version, and the receiving instance is still not supporting "Negative candidates"?
screen shot 2018-04-02 at 1 57 29 pm

allasm:
We need to fix push. Push knows it pushes to an old version, it should do some data conversions before pushing

sashaandjic:
Tomorrow at dev meeting we should confirm the order. I was expecting to see the new status being added as a last one.

veronikaslc:
@sashaandjic Just tested push and it worked as expected. 
Can you provide some details of the push testing? which version to which version did you push? Also where do you get those tables?

sashaandjic:
Yes, I was pushing to the current version of phenomecentral that does not support newly added gene status. I am also not sure what to do with Rejected candidate when pushing to an older version. Should we push them as negative? Let see with Brittney tomorrow.

veronikaslc:
What is expected behaviour?

sashaandjic:
@sdumitriu

sashaandjic:
Why is Negative converting to Candidates? Is it because we changed the order?

sashaandjic:
@veronikaslc I don't think it is good idea to reuse "rejected" for the new status of "Rejected Candidate", as "rejected" for already used for "Negative" genes. This may cause a lot of problems down the road.

veronikaslc:
That is the question: was the original negative genes "tested negative" or they were more of a "rejected candidate"?

sashaandjic:
The original "Negative" genes had value "rejected" in JSON.

veronikaslc:
I mean not by value but by sence

sashaandjic:
What we called Negative before should be called Tested Negative in 1.4

sashaandjic:
Rejected Candidates we never had

veronikaslc:
Brittney confirms?

allasm:
We should NOT be changing JSON representation of existing statuses, as it breaks compatibility (even if the string is now a bit confusing). So no renaming of "rejected" (which is now "negative") to anything else, at least for now. @marta- , do you confirm?

allasm:
After discussing this with Brittney I think we decided the best way to push new status is to mark it as "negative" (the way old instances support it) but add a text comment "rejected candidate". This will ensure this gene will not be used for matching (most important, since this is what PT is about), but the user will know it is a "rejected candidate" due to text label

veronikaslc:
I think here is a bit of misunderstanding. I am wirried, what if genes marked previously as negative were not tested and users were just rejecting them as candidates that way.

components/patient-data/ui/src/main/resources/PhenoTips/GeneClass.xml
veronikaslc: changed the order up to suggested by Brittney

veronikaslc:
codeview comments addressed,
commits are rebased and some are squashed for the codereview convenience

sashaandjic:
Problems with push identified:
- pushing on self no longer works
- pushing to ver 1.4M3: genes/variants are not pushed
- pushing to PREPROD (1.3.6) omits pushing genes/variants and Maternal/Paternal age at

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
```diff
@@ -86,7 +86,7 @@
 
     private static final String CONTROLLER_NAME = VARIANTS_STRING;
 
-    private static final String VARIANTS_ENABLING_FIELD_NAME = "genes";
+    private static final String VARIANTS_ENABLING_FIELD_NAME = "variants";
```
allasm: Just want to make sure we do not merge this until we have the new UIX field? I think as discussed right now having enabling field to be `variants` does not work in some cases (e.g. consents don't play well with this and we may end up exporting variants when there is no consent for that?)

sashaandjic:
pt-3602

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/VariantListController.java
veronikaslc: This PR doesn't not affect any consent configurations or work and can be merged separately.

sashaandjic:
Maybe add an "*" before?
_"* rejected candidate"_

veronikaslc:
@sashaandjic comments fixed

veronikaslc:
oops, one extra commit was pushed from another branch, now removed

veronikaslc:
@sashaandjic updated last commit wrt new comments

veronikaslc:
@sashaandjic Maternal/Paternal age push is already fixed in PT 1.4

a-gvo:
Tested. Looks good.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/oldversions/GeneListControllerV1.java
```diff
@@ -0,0 +1,151 @@
...
+
+import javax.annotation.Nonnull;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+/**
+ * Handles the patients genes in pre-1.4M1.
```
sdumitriu: Wrong version, just say `pre-1.4`.

components/patient-data/impl/src/main/java/org/phenotips/data/internal/controller/oldversions/GeneListControllerV1.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @version $Id$
+ * @since 1.4M1
```
sdumitriu: `@since 1.4`

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
```diff
@@ -2413,6 +2413,7 @@ document.observe('xwiki:dom:loaded', function()
         this._builder.rejectedGeneIDsInput.value = '';
         for (var gene in this._builder.genes) {
           (this._builder.genes[gene].status == 'rejected') &amp;&amp; (_this._builder.rejectedGeneIDsInput.value += gene + ' ');
+          (this._builder.genes[gene].status == 'rejected_candidate') &amp;&amp; (_this._builder.rejectedGeneIDsInput.value += gene + ' ');
```
sdumitriu: Shouldn't both values be considered here?

components/patient-data/impl/pom.xml
```diff
@@ -30,7 +30,7 @@
 
   <properties>
     <checkstyle.suppressions.location>${basedir}/src/checkstyle/checkstyle-suppressions.xml</checkstyle.suppressions.location>
-    <coverage.instructionRatio>0.70</coverage.instructionRatio>
+    <coverage.instructionRatio>0.7</coverage.instructionRatio>
```
sdumitriu: The rule is to always use two digits of precision.

components/patient-data/ui/src/main/resources/PhenoTips/PatientSheetCode.xml
veronikaslc: It seems that line 2415 is checking 'rejected' status, or I still did not get the question

sdumitriu: You are right, I thought line 2415 was removed and line 2416 added in its place.

PR #2601, opened 2018-04-02
PT-3485:  Permissions dialog: clicking on the new owner field should automatically select Transfer ownership to another user or group

veronikaslc:


veronikaslc:
That would be strange, because we decided to disable input field until user selects corresponding radio button.

sdumitriu:
Right... It doesn't make sense to have the input disabled. Make the input always enabled, and when it gets clicked, also select the associated radio button.

veronikaslc:
We decided to leave PR like that. The JIRA issue name changed accordingly implementation.

veronikaslc:
@sashaandjic this one was decided to be implemented as originally planned, ready to test

veronikaslc:
@sdumitriu fixed

PR #2606, opened 2018-04-10
PT-3430: Remove or replace all occurrences of the word "node" from the pedigree UI

marta-:


sashaandjic:
screen shot 2018-04-10 at 5 08 24 pm

marta-:
Thank you @sashaandjic . Please re-test.

components/pedigree/resources/src/main/resources/pedigree/view/legend.js
```diff
@@ -713,11 +713,11 @@ define([
          */
         _onDropObject: function(node, objectID) {
             if (!this._dropOnGroupNodes && node.isPersonGroup()) {
-                this._onFailedDrag(node, "Can't drop onto this node: group nodes do not support " + this._droppableName, "Can't drag to this node");
+                this._onFailedDrag(node, this._droppableName + " cannot be assigned to groups of individuals.", "Unsupported operation");
                 return false;
             }
             if (!editor.getPatientAccessPermissions(node.getPhenotipsPatientId()).hasEdit) {
-                this._onFailedDrag(node, "You do not have edit right for this patient", "Can't drag to this patient");
+                this._onFailedDrag(node, "You do not have the permission to modify this patient.", "Unauthorized operation");
```
sdumitriu: ...this patient *record*?

components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
```diff
@@ -311,7 +311,7 @@ define([
             },
             {
                 'name' : 'setproband',
-                'label' : 'Set this person to be the proband node',
+                'label' : 'This person is the proband',
```
allasm: When there is a button with this text I think it can be confused with a statement (e.g. it may look like an _affirmation_ "yes, this patient is the proband" as opposed to _action_ "make this patient the proband")

components/pedigree/resources/src/main/resources/pedigree/view/personHoverbox.js
```diff
@@ -106,7 +106,7 @@ define([
                     }
                     // parent handle
                     this.generateHandle('parent', x, splitLocationY, x, y - PedigreeEditorParameters.attributes.personHandleLength,
-                                        "Click to create new nodes for the parents or drag to an existing person or partnership (valid choices will be highlighted in green). Dragging to a person will create a new relationship.", "F", topHandleHint);
+                                        "Click to draw parents or drag to an existing person or partnership (valid choices will be highlighted in green). Dragging to a person will create a new relationship.", "F", topHandleHint);
```
allasm: just curious, in the context of creating pedigree nodes, we want to use the word "draw" wherever we used to say "create"?

marta-: "Create" is a word that we, devs, tend to use because we know we're instantiating an object. Our target users are accustomed to "drawing" pedigrees.

allasm: That was my question - so I see that for them new stuff is "drawn"

components/pedigree/resources/src/main/resources/pedigree/nodeMenuFields.js
marta-: Right. Which reminds me I wanted to change this button to a checkbox. For now I'll change the label ti "make this...". Thanks.

components/pedigree/resources/src/main/resources/pedigree/view/legend.js
marta-: Right, thanks.

marta-:
> Both individual and person are used in multiple places, should we standardize on one of them, or both are valid in their respective places?

@sdumitriu I think both work fine where they are. Since it's not a technical term, we don't need to worry that much about consistency. Let me know if you disagree.

marta-:
@allasm @sdumitriu Your comments have been addressed.

PR #2608, opened 2018-04-12
PT-3651: Add ability to consent/push/export genes and variants separately

veronikaslc:
Added new UIX genes extension that is generating UI only if gene-variants UIX is disabled

### DIFF
--- a/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
+++ b/components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
@@ -296,6 +296,7 @@ public class DataToCellConverter
     {
         String sectionName = "variants";
         Set<String> present = this.enabledHeaderIdsBySection.get("genes");
+        // if genes are not included or variants are not enabled
         if (present == null || present.isEmpty() || !enabledFields.remove(sectionName)) {
             return;
         }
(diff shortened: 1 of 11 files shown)


veronikaslc:
@sashaandjic this PR is based on PT-3602 branch

veronikaslc:
@sashaandjic ready to be tested

veronikaslc:
should be neither, committed a fix for auto checking variants consent if genes consent is checked.

a-gvo:
After both genes and variants are selected, i can still un-check just variants.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -294,9 +294,9 @@ public DataSection genesBody(Patient patient) throws Exception
 
     public void variantsSetup(Set<String> enabledFields) throws Exception
     {
-        String sectionName = "genes";
-        Set<String> present = this.enabledHeaderIdsBySection.get(sectionName);
-        if (present == null || present.isEmpty()) {
+        String sectionName = "variants";
```
sdumitriu: Excel export never includes variants with this change. `sectionName` must also be changed in the next two methods, `variantsHeader` and `variantsBody`,

PR #2611, opened 2018-04-18
PT-3635: Cases with confirmed causal genes should be solved

veronikaslc:


sashaandjic:
Should we prompt the user on save to avoid unnecessary multiple prompts while entering data?

sashaandjic:
It looks much better. I noticed that in the warning message we only mention Final diagnosis to be populated, however if Final is blank and Clinical diagnosis is populated this message does no appear. 
We should either remove "final" from "Please also enter the final diagnosis in the “Diagnosis” section." or skip checking Clinical diagnosis all together. (e.g. create new patient, enter clinical diagnosis, the msg pops up even though final diagnosis is blank)

veronikaslc:
But the code currently does not check clinical diagnosis at all, just final diagnosis. The message intentionally appears only when either gene us entered, or final diagnosis or case checked as solved.

sashaandjic:
@veronika create new patient and enter clinical diagnosis, you will get this
screen shot 2018-04-19 at 8 03 38 pm

sashaandjic:
I believe it doesn't. I'll ask Brittney to confirm.

veronikaslc:
Brittney confirmed that only final diagnoses should be watched.

veronikaslc:
Paused for now

PR #2615, opened 2018-04-20
PT-3660: If the DetailsDialog contains no data, "no additional information" should be displayed on loss of focus

a-gvo:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -7166,11 +7177,31 @@ document.observe("xwiki:dom:updated", init);
           var _this = this;
           this._addDetailsButton.observe('click', function (event) {
               event.stop();
-              _this.addDialog(false, _this._globalQualifierCount &amp;&amp; _this._globalQualifierCount())
+              _this._shouldAddDialog.bind(_this)() &amp;&amp; _this.addDialog(false, _this._globalQualifierCount &amp;&amp; _this._globalQualifierCount())
```
sdumitriu: Why `bind` it?

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -6066,6 +6066,11 @@ document.observe("xwiki:dom:updated", init);
               this._termDetails.removeClassName('focused');
               this._termDetails.addClassName('blurred');
               this._toggleSummarize();
+              if(!this._termDetails.down("dl &gt; *:not([style='display: none;']):not(.hidden)")) {
```
sdumitriu: I'm not a big fan of the `style=` check. Are you sure it works in all browsers, including IE?

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
a-gvo: Oops. Removed.

components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
a-gvo: Not sure, actually. Replaced it with a different check.

PR #2620, opened 2018-04-25
Stable 1.3.7 fixes

allasm:


allasm:
@sashaandjic: For testing:

- PT-3588: just do what issue description says and make sure the behaviour is correct

- PT-3586: check what the issue description says, but also check patient REST import, in particular make sure REST works correctly with features and non-standard features (adds all theat are included, removed those that are not included)

sashaandjic:
tested. works as expected.

PR #2631, opened 2018-05-03
PT-3656: Pushing patient record and assigning it to one of user group does not work

veronikaslc:
Fixed by re-fetching patient document after the owner was updated


components/patient-data-sharing/receiver-api/src/main/java/org/phenotips/data/receive/internal/DefaultReceivePatientData.java
```diff
@@ -545,6 +545,8 @@ public JSONObject receivePatient()
                 this.logger.warn("Created new patient successfully");
             }
 
+            // get updated saved instance of patient document
+            affectedPatient = this.patientRepository.get(affectedPatient.getDocument());
```
allasm: I think we only need to do this if we changed permissions in all other cases it would be the same object

PR #2632, opened 2018-05-04
PT-3669: Rename Gene Panel as Suggested Genes

veronikaslc:


components/widgets/ui/src/main/resources/PhenoTips/Widgets.xml
```diff
@@ -4456,7 +4456,7 @@ document.observe("xwiki:dom:updated", init);
                         { name : 'Entrez',       url : 'http://www.ncbi.nlm.nih.gov/gene/?term=',                     field : 'entrez_id'},
                         { name : 'RefSeq',       url : 'http://www.ncbi.nlm.nih.gov/nuccore/',                        field : 'refseq_accession'},
                         { name : 'Ensembl',      url : 'http://useast.ensembl.org/Homo_sapiens/Gene/Compara_Tree?g=', field : 'ensembl_gene_id'},
-                        { name : 'Gene panels',  url : 'https://www.ncbi.nlm.nih.gov/gtr/all/tests/?term=',           field : 'genecards_id'}
```
sdumitriu: Changing this is not right, but after looking at the code this property doesn't exist at all, so this is effectively a dead entry, it could be removed completely.

PR #2636, opened 2018-05-14
 PT-3663: Modify Case Resolution and Diagnosis sections

veronikaslc:


veronikaslc:
For now PR is for initial codereview/functional testing while SolvedControllerTest is in progress.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1363,23 +1364,32 @@ public DataSection isSolvedBody(Patient patient)
...
         DataSection bodySection = new DataSection();
-        PatientData<String> patientData = patient.getData("solved");
+        PatientData<SolvedData> patientData = patient.getData("solved");
+        SolvedData data = patientData != null ? patientData.getValue() : null;
 
         int x = 0;
-        if (present.contains("solved")) {
-            String solved = patientData != null ? patientData.get("solved") : null;
+        if (present.contains(SolvedData.STATUS_PROPERTY_NAME)) {
+            String solved = data != null ? data.getStatus() : null;
```
marta-: We're on Java 8, so now you can do `String solved = data?.getStatus();`

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
```diff
@@ -1363,23 +1364,32 @@ public DataSection isSolvedBody(Patient patient)
...
         }
-        if (present.contains("solved__pubmed_id")) {
-            String pubmedId = patientData != null ? patientData.get("solved__pubmed_id") : null;
-            DataCell cell = new DataCell(pubmedId != null ? pubmedId : "", x, 0);
-            bodySection.addCell(cell);
-            x++;
+
+        if (present.contains(SolvedData.PUBMED_ID_PROPERTY_NAME)) {
+            List<String> pubmedIds = data != null ? data.getPubmedIds() : null;
+            if (!pubmedIds.isEmpty()) {
```
marta-: Won't this give you a NPE if `data` is `null`?

Again, you could use the short syntax on the line before:
`List<String> pubmedIds =  data?.getPubmedIds() ?: Collections.emptyList();`

components/patient-data/impl/pom.xml
```diff
@@ -30,7 +30,7 @@
 
   <properties>
     <checkstyle.suppressions.location>${basedir}/src/checkstyle/checkstyle-suppressions.xml</checkstyle.suppressions.location>
-    <coverage.instructionRatio>0.70</coverage.instructionRatio>
+    <coverage.instructionRatio>0.69</coverage.instructionRatio>
```
marta-: Noooooooo! :(

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
veronikaslc: This won't give NPE, both SolvedData class sets pubmedIds to empty list for null inputs.

components/patient-data/impl/pom.xml
veronikaslc: Please see the comment to the PR and to the commit. This PR is for codereview/functional testing only, tests are in progress.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
veronikaslc: It will give NPE, will fix.

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
veronikaslc: It doesn't compile..

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
veronikaslc: for short syntax: same - doesn't compile

components/export/api/src/main/java/org/phenotips/export/internal/DataToCellConverter.java
marta-: Sorry, I meant to also give the link to the `Optional` documentation but forgot. Here's an article: http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html

veronikaslc: Still don't see how to use `data?.getStatus()` syntax with Optional or anything that would be as easy as `?.`, it seems to be from another language. In general I personally find explicit null checks more readable.

allasm: `?.` syntax would be ideal in this case, but it seems to be a Groovy (?) and not Java construct? At the same time IMHO in simple cases like this `Optional` does not add much clarity or safety and is actually longer/harder to comprehend. IMHO it makes more sense for deep embedded values like in the example from the link above and/or when the optional value in question is used much further or much more in the code (so all those other places can benefit from the optional nature of the value or are forced to treat it as optional, even if by that line of code it is no longer obvious)

sashaandjic:
Clinical diagnosis label and field are misaligned.
Other UI changes are logged in subtask PT-3678.

veronikaslc:
@sashaandjic made a separate PR for PT-3678

sashaandjic:
Should this PR be retested?

sashaandjic:
alignments are off. this will be logged in a separate task.
diagnosis section alignments

sashaandjic:
tested

PR #2640, opened 2018-05-23
PT-3678: Improve the look of Diagnosis section after merge with Solved section

veronikaslc:


sashaandjic:
Birttney and I think it looks very good. One minor change: publication fields need to be aligned with diagnoses.
align1

@marta- please check when you have a minute

veronikaslc:
@sashaandjic codereview comments addressed

sashaandjic:
Tested. Looking much better. It bothers me that we use two different presentation for diagnoses and for publications, however that can be addressed separatelly.
screen shot 2018-05-26 at 2 08 43 pm

PR #2647, opened 2018-06-01
PT-3632: Upon successful reindexing corresponding vocabulary fields turn into edit mode

veronikaslc:
Enabling just reindex button instead of enabling the whole form


veronikaslc:
Don't know how to reproduce this " Hello world!" issue, but edit mode is fixed.

sdumitriu:
The hello world bug has been fixed already.

PR #2649, opened 2018-06-03
PT-3694: Add support for auditing who viewed and edited records

sdumitriu:
Done


marta-:
Code looks good. In the next iteration I would add an admin section to display the audit logs (even in this rudimentary format) and the ability to turn auditing on/off from the administration.

danielpgross:
@sdumitriu, this looks to be a very useful feature, one that users have been wanting for a long time. Thank you.

Does the current implementation handle the case where a record is opened in edit mode, (using the `edit` action), instead of view mode (using the `get` action)? It is important that an audit entry be logged in this case because users often open records in edit mode just to view the contents. From what I can see, this case is not currently being handled.

sdumitriu:
Yes it is, all (struts) actions are logged.

PR #2657, opened 2018-06-04
PT-3706: Measurements don't work for non-admin users

sdumitriu:
Fixed by backporting http://jira.xwiki.org/browse/XWIKI-13057


marta-:
Would have been easier to follow in 2 commits, one with the version we are currently using of the two files and one with the modified version.

PR #2658, opened 2018-06-04
PT-3708: Automatic layout may leave pedigree in an un-saveable un-exportable state

allasm:
@sashaandjic talk with me about testing


sdumitriu:
@sashaandjic Is that problem introduced by this pull request? If not, since "automatic layout" is not used in your procedure, it is completely unrelated and should be filed as a separate bug.

PR #2659, opened 2018-06-04
PT-3581: Prevent opening same patient/family in multiple tabs from pedigree editor and patient form

veronikaslc:
Fixed.

### DIFF
--- a/components/family-studies/ui/src/main/resources/PhenoTips/FamilyReferenceClass.xml
+++ b/components/family-studies/ui/src/main/resources/PhenoTips/FamilyReferenceClass.xml
@@ -53,7 +53,7 @@
       <customDisplay>{{velocity}}
 #set ($family = $services.families.getFamilyById($value))
 #if ("$!family.getId()" != '')
-[[$family.getId()#if ("$family.getExternalId()" != '') ($family.getExternalId())#end&gt;&gt;path:$xwiki.getURL($family.getDocumentReference(), 'view')||id="familyId"]]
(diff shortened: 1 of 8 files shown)


components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
```diff
@@ -1285,6 +1285,7 @@ define([
                 } else {
                     if (label) {label.update("Linked to patient record");}
                     suggestContainer.hide();
+                    link.target = _this.targetNode.getPhenotipsPatientId();
```
allasm: i think it should be `value` not `_this.targetNode.getPhenotipsPatientId()`. After all this is what is used to generate the url on the very next line

components/family-studies/ui/src/main/resources/PhenoTips/FamilyReferenceClass.xml
```diff
@@ -53,7 +53,7 @@
       <customDisplay>{{velocity}}
 #set ($family = $services.families.getFamilyById($value))
 #if ("$!family.getId()" != '')
-[[$family.getId()#if ("$family.getExternalId()" != '') ($family.getExternalId())#end&gt;&gt;path:$xwiki.getURL($family.getDocumentReference(), 'view')||id="familyId"]]
+[[$family.getId()#if ("$family.getExternalId()" != '') ($family.getExternalId())#end&gt;&gt;path:$xwiki.getURL($family.getDocumentReference(), 'view')||id="familyId" target="$family.getExternalId()"]]
```
sdumitriu: For families without an external ID, the family will open in the same tab.

PR #2660, opened 2018-06-05
PT-3717: Show internal/external family ID under Family history according to configuration for IDs from Administration: Data formats

veronikaslc:


veronikaslc:
@marta- Now when we display just family external ID under Family history (if configured such), should we prep-end a label `Family:`, so it will look like:
Family: 123123123

sashaandjic:
retested. works as expected.

veronikaslc:
I can not, that macro gets ID of a currently opened document, I need a family of a current patient.

PR #2666, opened 2018-06-07
PT-2091: User profile's setting label and icon are overlapped

veronikaslc:
Fixed


veronikaslc:
@sdumitriu found alternative way

sdumitriu:
The actual problem was a misplaced `#if` in `PhenoTipsGroupSheet`.

PR #2667, opened 2018-06-07
Pedigree bugfixes for 1.4-rc1

allasm:


allasm:
Note: a few more fixes will be added to this branch. Some of them depend in the fixes already here, thus all are grouped into the same pull request

allasm:
@sashaandjic : need to pay extra attention to cancers: set different cancers with details, same caner but with different details for different patients, make sure node menu shows them correctly when switching nodes, make sure save/load is correct

sashaandjic:
Ok. Will do.

allasm:
@sashaandjic also need to test how more un-linked patients behave, and how load and save works: create pedigree with unlinked, try to close, make sure it warns about unsaved changes, save, change, undo, close -> now should not warn. Re-load, check that all un-linked are there, add more, remove more, undo/redo/.save/etc..

This relates to PT-3727, PT-3718, PT-2850, PT-3680, all of which should  be fixed now

PR #2668, opened 2018-06-08
More tests and better codestyle

sdumitriu:


components/patient-data/impl/src/main/java/org/phenotips/data/internal/EmptyGenotypeObjectsRemover.java
```diff
@@ -74,54 +74,29 @@ public void onEvent(Event event, Object source, Object data)
...
-            genes.add(field.getValue());
-        }
+        // Remove gene object from the document if key "gene" field is empty
+        geneXWikiObjects.stream().filter(Objects::nonNull)
+            .filter(gene -> StringUtils.isBlank(gene.getStringValue(GENE_KEY)))
+            .forEach(gene -> doc.removeXObject(gene));
 
         if (variantXWikiObjects == null || variantXWikiObjects.isEmpty()) {
-            // nothing to remove
+            // Nothing else to remove
```
veronikaslc: we can move line 75 to line 87 since we don't need variants before that point but can exit function.

PR #2670, opened 2018-06-14
PT-3674: When removing genes from a person in pedigree editor, all associated gene variants need to be removed too

veronikaslc:
Fixed NPE if there are no genes and variants


veronikaslc:
before (no genes && no variants) situation was not caught for immediate return

allasm:
After a quick glance I also thought there is no difference, but the difference is that now `return` is called when there are no genes (while `doc.removeXObjects` is called in exactly the same cases as before)

sdumitriu:
Indeed, didn't notice the return.

PR #2672, opened 2018-06-16
PT-3732: Family external_id cannot be saved when a family has at least one member

marta-:
This fixes a variable leak. However the root problem may be somewhere else.


marta-:
Closing in favor of a general fix. @sdumitriu discovered the issue is caused by https://phenotips.atlassian.net/browse/PT-3706 which was fixed by a backport that carried the bug https://jira.xwiki.org/browse/XWIKI-13558.

sashaandjic:
tested, but had to build it with -Pquick option

PR #2676, opened 2018-06-20
PT-3711: Administrators can browse and filter audited events

veronikaslc:
Tests are coming.


components/security/audit/backend/src/main/java/org/phenotips/security/audit/internal/HibernateAuditStore.java
```diff
@@ -104,13 +118,151 @@ public void store(AuditEvent event)
...
+            }
+
+            return 0;
+        } catch (HibernateException ex) {
+            this.logger.error("Failed to count audit event documents: {}", ex.getMessage(), ex);
+            return -1;
+        }
+    }
+
+    private void setTimeInterval(Criteria c, Calendar fromTime, Calendar toTime)
```
sdumitriu: It would be better to not enforce time between [Epoch .. NOW], and allow `from` and `to` to be null, in which case instead of adding a `between` filter, we'd add, as needed, one of `between`, `ge`, `le`, or not restriction at all.

components/security/audit/frontend/src/main/resources/PhenoTips/AuditResults.xml
```diff
@@ -0,0 +1,82 @@
...
+#if (!$limit)
+  #set ($limit = 25)
+#end
+#set($user = $request.get('user'))
+#set($ip = $request.get('ip'))
+#set($action = $request.get('action'))
+#set($entity = $request.get('record'))
+#set($fromTime = $request.get('fromTime'))
+#set($toTime = $request.get('toTime'))
+#set($data = $services.audit.getEvents($offset, $limit, $action, $user, $ip, $entity, $fromTime, $toTime))
```
sdumitriu: Bug: the java code is 0-indexed, but the livetable is 1-indexed. The first event is always inaccessible.

components/security/audit/frontend/src/main/resources/PhenoTips/AuditResults.xml
```diff
@@ -0,0 +1,82 @@
...
+  <date>1401822196000</date>
+  <contentUpdateDate>1401822196000</contentUpdateDate>
+  <version>1.1</version>
+  <title/>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>{{velocity wiki="false"}}
+$response.setContentType('application/json')
```
sdumitriu: This makes the page unusable in view mode, for example to try and export it. Usually we place a check like:

    #if ($xcontext.action != 'get')
      #break
    #end

PR #2678, opened 2018-06-25
PT-3720: Move SHOW EXTERNAL IDENTIFIERS... to the Admin's External ID configuration page

veronikaslc:
Done.


components/application-configuration/ui/src/main/resources/PhenoTips/LabeledIdentifierGlobalSettings.xml
```diff
@@ -48,7 +48,7 @@
     <nameField/>
     <validationScript/>
     <allowOtherEids>
-      <defaultValue>1</defaultValue>
```
sdumitriu: Why this change?

PR #2681, opened 2018-06-26
PT-3703: The "required" flag is ignored when a patient record is populated from the pedigree

allasm:


sashaandjic:
Pedigree editor is enforcing required fields, however Patient record form isn't.
For DOB Patient record form only enforces the year of birth.
For DOD only year of birth is enforced, however if year of death isn't entered there is no error msg. The only noticeable thing is that all 3 buttons at the bottom become "pale" blue.
screen shot 2018-06-27 at 12 37 15 pm
screen shot 2018-06-27 at 12 37 44 pm
screen shot 2018-06-27 at 12 42 06 pm

marta-:
@sashaandjic please report the faulty required date behavior as a separate issue.

components/pedigree/resources/src/main/resources/pedigree/externalEndpoints.js
```diff
@@ -105,6 +105,18 @@ define([
             return XWiki.contextPath + "/rest/patients/" + patientId + "?method=DELETE";
         },
 
+        getPatientUpdateURL: function(patientId) {
+            // note: have to use "?method=PUT" because of pecularities of how prototype Ajax.Request works
+            //       see: http://api.prototypejs.org/ajax/
+            //       see: https://stackoverflow.com/questions/4476810/is-it-really-possible-to-call-http-put-using-prototype
+            return XWiki.contextPath + "/rest/patients/" + patientId + "?method=PUT";
```
veronikaslc: we can add `this.patientService = XWiki.contextPath + "/rest/patients";` to have one point where we store REST URL and use it here like `return this.patientService + patientId + "?method=PUT"`

PR #2684, opened 2018-06-26
PT-3742: Add uniqueness ID check configuration to Administration

veronikaslc:
Done.

### DIFF
--- a/components/application-configuration/api/src/main/resources/ApplicationResources.properties
+++ b/components/application-configuration/api/src/main/resources/ApplicationResources.properties
@@ -20,4 +20,4 @@ PhenoTips.LabeledIdentifierSettings_allowDuplicateEids=Allow duplicate external 
 PhenoTips.LabeledIdentifierGlobalSettings_allowOtherEids=Allow other external identifiers
 PhenoTips.LabeledIdentifierGlobalSettings_hideInternalIds=Show external identifiers (when available) instead of internal ones in the Patient and Family records
 PhenoTips.LabeledIdentifierGlobalSettings_externalIdFormat=External id display format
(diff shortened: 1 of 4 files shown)


components/application-configuration/api/src/main/resources/ApplicationResources.properties
```diff
@@ -20,3 +20,4 @@ PhenoTips.LabeledIdentifierSettings_allowDuplicateEids=Allow duplicate external
 PhenoTips.LabeledIdentifierGlobalSettings_allowOtherEids=Allow other external identifiers
 PhenoTips.LabeledIdentifierGlobalSettings_hideInternalIds=Show external identifiers (when available) instead of internal ones in the Patient and Family records
 PhenoTips.LabeledIdentifierGlobalSettings_externalIdFormat=External id display format
+PhenoTips.LabeledIdentifierGlobalSettings_validateExternalId=Check uniqueness of external ID
```
marta-: Maybe "Require unique external identifiers" is a bit clearer?

veronikaslc:
Yes. This PR is a part of the preparation to be used in PT-3704 : "The id uniqueness is not checked when a patient record is populated from the pedigree editor".

components/application-configuration/api/src/main/resources/ApplicationResources.properties
veronikaslc: comment addressed

allasm:
As discussed with @marta- yesterday this change is needed for PT-3704 which will use it, and we will create a separate issue to use this setting on the patient form (the urgency comes from the fact that it is easy/easier to adjust required check behaviour on the patient form, e.g. on PhenomeCentral which should NOT check ID uniqueness, but it is harder to adjust pedigree behaviour)

PR #2691, opened 2018-06-29
PT-3749: Saving and loading pedigree may fail if current user has no view rights for some of the patients in the family

allasm:


allasm:
Note: github does not handle indentation changes for large blocks, commit 8f40920 mostly just added try-catch blocks around large chunks of code but when checking the diff on github it seems like there are lots of chnages

marta-:
Add `?w=1` to the url to ignore whitespace modifications when vieweing changes. E.g. https://github.com/phenotips/phenotips/commit/8f409206d1ed228e958e0ecd355f8aa777dbb753?w=1

PR #2694, opened 2018-07-03
PT-2134: When creating a new patient within pedigree editor, the system does not ask if the patient is a part of a study

veronikaslc:
Fixed.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/controller.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/controller.js
@@ -261,6 +261,35 @@ define([
                 }
             }
 
+            var callPatientCreate = function(studyName) {
+                var createPatientURL = editor.getExternalEndpoint().getFamilyNewPatientURL();
+                if (studyName && studyName.length > 0) {
+                    createPatientURL += "&studyName=" + studyName;
+                }
+                document.fire("pedigree:blockinteraction:start", {"message": "Waiting for the patient record to be created..."});
+                new Ajax.Request(createPatientURL, {
+                    method: "GET",
+                    onSuccess: function(response) {
+                        if (response.responseJSON && response.responseJSON.hasOwnProperty("newID")) {
+                            console.log("Created new patient: " + Helpers.stringifyObject(response.responseJSON));
+                            onCreated(response.responseJSON.newID)
+                        } else {
+                            onFailureHandler();
+                        }
+                    },
+                    onFailure: onFailureHandler,
+                    onComplete: function() { document.fire("pedigree:blockinteraction:finish"); }
+                });
+            };
+
+            var studies = editor.getPreferencesManager().getConfigurationOption("studies");
+            if (studies.length > 0) {
+                var studySelectionDialog = editor.getStudySelectionDialog();
+                studySelectionDialog.show(callPatientCreate);
+            } else {
+                callPatientCreate();
+            }
+
             var onCreatedHandler = event.memo.onCreatedHandler;
             var onFailureHandler = event.memo.onFailureHandler
                                    ? event.memo.onFailureHandler
@@ -308,22 +337,6 @@ define([
                     onCreatedAndUpdated(newID);
                 }
             }
-
-            var createPatientURL = editor.getExternalEndpoint().getFamilyNewPatientURL();
-            document.fire("pedigree:blockinteraction:start", {"message": "Waiting for the patient record to be created..."});
-            new Ajax.Request(createPatientURL, {
-                method: "GET",
-                onSuccess: function(response) {
-                    if (response.responseJSON && response.responseJSON.hasOwnProperty("newID")) {
-                        console.log("Created new patient: " + Helpers.stringifyObject(response.responseJSON));
-                        onCreated(response.responseJSON.newID)
-                    } else {
-                        onFailureHandler();
-                    }
-                },
-                onFailure: onFailureHandler,
-                onComplete: function() { document.fire("pedigree:blockinteraction:finish"); }
-            });
         },
 
         handleRemove: function(event)
--- a/components/pedigree/resources/src/main/resources/pedigree/controller.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/controller.js
@@ -263,7 +263,7 @@ define([
 
             var callPatientCreate = function(studyName) {
                 var createPatientURL = editor.getExternalEndpoint().getFamilyNewPatientURL();
-                if (studyName && studyName.length > 0) {
+                if (studyName) {
                     createPatientURL += "&studyName=" + studyName;
                 }
                 document.fire("pedigree:blockinteraction:start", {"message": "Waiting for the patient record to be created..."});
--- a/components/pedigree/resources/src/main/resources/pedigree/controller.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/controller.js
@@ -261,35 +261,6 @@ define([
                 }
             }
 
-            var callPatientCreate = function(studyName) {
-                var createPatientURL = editor.getExternalEndpoint().getFamilyNewPatientURL();
-                if (studyName) {
-                    createPatientURL += "&studyName=" + studyName;
-                }
-                document.fire("pedigree:blockinteraction:start", {"message": "Waiting for the patient record to be created..."});
-                new Ajax.Request(createPatientURL, {
-                    method: "GET",
-                    onSuccess: function(response) {
-                        if (response.responseJSON && response.responseJSON.hasOwnProperty("newID")) {
-                            console.log("Created new patient: " + Helpers.stringifyObject(response.responseJSON));
-                            onCreated(response.responseJSON.newID)
-                        } else {
-                            onFailureHandler();
-                        }
-                    },
-                    onFailure: onFailureHandler,
-                    onComplete: function() { document.fire("pedigree:blockinteraction:finish"); }
-                });
-            };
-
-            var studies = editor.getPreferencesManager().getConfigurationOption("studies");
-            if (studies.length > 0) {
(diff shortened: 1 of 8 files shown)


components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -261,6 +261,55 @@ define([
                 }
             }
 
+            // Ask if the patient should be a part of a study
```
allasm: This block should be moved into its own dialog, similar to `newPersonMenu`, mixing UI code with patient creation logic is not a very good idea

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -261,6 +261,55 @@ define([
...
+            });
+
+            var getStudyDialogURL = editor.getExternalEndpoint().getStudyBindingService();
+            new Ajax.Request(getStudyDialogURL, {
+                onCreate : function() {
+                  //templateSelectionContainer.update('<img src="$xwiki.getSkinFile('icons/xwiki/ajax-loader-large.gif')"/>');
+                  modalBox.showDialog();
+                },
+                onSuccess : function(response) {
+                   templateSelectionContainer.update(response.responseText);
```
allasm: this is "OK" for now, but using HTML from PhenoTips is a not a very good idea, since
 a) we have to reply on a certain HTML structure to parse the list of available studies
 b) (less important) this adds another dependence on specifics of XWiki code to pedigree

I think both problems can be avoided when unrelated necessary fix descrobed below is implemented

allasm: unrelated problem is that this dialog pops up even if there are no studies configured in the PT instance, which is the case for most standalone out-of-the box instances. So regardless of everything else a check if there are studies should be done in `https://github.com/phenotips/phenotips/blob/master/components/pedigree/ui/src/main/resources/PhenoTips/PedigreeInterface.xml` and passed to the pedigree editor, and study selection dialog should only be shown if ther eis at least one study. But if that has to be done, might as well pass the list of all study names o the pedigree, in which case we do not need a `StudyBindingService` any more, and a simple dialogue with a simple list of radio buttons or a simple select can be used instead

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -261,6 +261,35 @@ define([
                 }
             }
 
+            var callPatientCreate = function(studyName) {
+                var createPatientURL = editor.getExternalEndpoint().getFamilyNewPatientURL();
+                if (studyName && studyName.length > 0) {
```
allasm: no need for `studyName.length > 0`, in javascript `if ("")` fails

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
```diff
@@ -179,6 +183,8 @@ define([
                     this._partnershipMenu = this.generatePartnershipMenu();
                     this._exportSelector = new ExportSelector();
                     this._printDialog = new PrintDialog();
+                    var studies = editor.getPreferencesManager().getConfigurationOption("studies");
```
allasm: the list of studies should be obtained inside `StudySelectionDialog` constructor, no need to clutter the main file (since both places have access to `editor.getPreferencesManager().getConfigurationOption("studies")`

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
```diff
@@ -100,8 +102,9 @@ define([
             //  drawNodeShadows:              {true|false}   - display small shadow under node graphic; default: "true"
             //  disabledFields:               [array]        - list of node-menu fields disabled for this installation
             //  replaceIdWithExternalID:      {true|false}   - when true, patient links display external ID as the link text (instead of PT ids)
-            //  displayCancerLabels:          {true|false}   - display labels for each afecting cancer; default: "true"
+            //  displayCancerLabels:          {true|false}   - display labels for each affecting cancer; default: "true"
             //  lineStyle:                    {"thin"|"regular"|"bold"} - controls the thickness of all lines in pedigree
+            //  studies:                      [array]        - array of all available studies
```
allasm: need to expand comment to say something like `array of available study objects in the following format: {"id": ..., "name": ..., "description": ...}`  to make it clear what kind of data is expected there

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * The UI Element for displaying prompts to the user in a movable/semi-transparent dialogue for study selection
+ *
+ * @class OkCancelDialogue
```
allasm: wrong class name

components/pedigree/resources/src/main/resources/pedigree/controller.js
```diff
@@ -261,6 +261,35 @@ define([
...
+                            onFailureHandler();
+                        }
+                    },
+                    onFailure: onFailureHandler,
+                    onComplete: function() { document.fire("pedigree:blockinteraction:finish"); }
+                });
+            };
+
+            var studies = editor.getPreferencesManager().getConfigurationOption("studies");
+            if (studies.length > 0) {
```
allasm: I'm not sure, but in case there are required fields to me it seems more logical to first ask for a study, and then ask for required fields (which matches the workflow in PT itself, when a new patient is created first a study selection dialog pops up, then the patient form appears (which would pop up the dialog to enter required fields in the CLIn instance).

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
```diff
@@ -0,0 +1,97 @@
...
+          var _this = this;
+
+          this.studies = studies;
+
+          var mainDiv = new Element('div', {'class' : "study-selection-dialog"});
+          var container = new Element('div', {'class' : 'xform'});
+
+          mainDiv.insert(container);
+
+          var submitButton = new Element('input', {type : 'button', value : "Submit", 'class' : 'button'});
```
allasm: I would say the button should be named `Select` and not `Submit`.

Less important, but maybe `Cancel` can also be renamed to something more explicit like "cancel patient creation", to make it clear that cancel not just dismissed the studies dialog, but the entire patient creation procedure?

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
```diff
@@ -0,0 +1,97 @@
...
+              selectedStudies.each( function(item) {
+                  if (item.checked) {
+                      studyName = item.value;
+                  }
+              });
+              _this.submit(studyName);
+              _this.hide();
+          });
+
+          var cancelButton = new Element('input', {type : 'button', value : "Cancel", 'class' : 'button secondary'});
```
allasm: a minor point, but study selection dialogue pops up either from the node menu, or from patient legend, both of which use different styles for the buttons, I think we should use uniform button styles within the same workflow

marta:
I noticed people on this repo keep @-ting me instead of @marta-. It would be great if that stopped, thanks.

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
```diff
@@ -0,0 +1,97 @@
...
+      hide: function() {
+          this.dialog.closeDialog();
+      },
+      
+      /**
+       * Default 
+       *
+       * @method hide
+       */
+      submit: function() {
```
allasm: I don't think we need a default `submit()` method at all, we will _always_ call `user provided `onOKFunction()`, and whenever the button is pressed the `hide()` is called anyhow

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
```diff
@@ -0,0 +1,97 @@
...
+          var header = new Element('dt').update('Please select one of the available studies');
+          studiesContainer.insert(header);
+
+          var studiesList = new Element('dd');
+          var noneStudy = new Element('div', {'class' : 'form-template-option'});
+          var input = new Element('input', {'type' : 'radio', 'name' : 'study-selection'});
+          input.checked = true;
+          var label = new Element('label').insert(input).insert(' None');
+          noneStudy.insert(label);
+          noneStudy.insert(new Element('p', {'class' : 'xHint'}).insert('Use the system\'s default form configuration: a generic form covering basic patient information and covering phenotypes related to major anatomic systems.'));
```
allasm: I think it would be better to make it uniform and add the default "no study" as one of the options passed ot the pedigree via the `PedigreeInterface` script (only if there is at least one other study), so that at least the text above is not hardcoded in pedigree Ui code? I'm sure it is a translation, which can be accessed form velocity, which is an added reason to use velocity to generate this

components/pedigree/resources/src/main/resources/pedigree/controller.js
veronikaslc: In the case of pedigree there are input fields presented and here it will be more logical to ask to fill them first, then ask for study. When a new patient is created there is no patient form opened yet. I think current workflow is reasonable.

components/pedigree/resources/src/main/resources/pedigree/view/studySelectionDialog.js
veronikaslc: "No study" is not one of the studies, the dialog is a right place to handle the accommodation for the this case to present in UI.

allasm: ...but hard-coding the description that is supposed to come from PhenoTips is not

PR #2696, opened 2018-07-04
PT-3700: Broken homepage displayed after upgrading from older versions

sdumitriu:


sdumitriu:
Problem already fixed in c051b4dc0

PR #2699, opened 2018-07-04
PT-3752: Pedigree editor does not prompt for save if patient is added to an existing family and left unliked

allasm:


components/pedigree/resources/src/main/resources/pedigree/pedigree.js
```diff
@@ -186,10 +186,23 @@ define([
...
+                        // to a node in the template - for that to happen it should be added to the patient legend
+                        // before the template is loaded
                         this.getPatientLegend().addCase(newPatientId);
+
                         this._saveLoadEngine.load(XWiki.currentDocument.page);
+
+                        // normally once a pedigree is loaded from disk it is assumed to have no changes,
+                        // but when a new patient is added to an existing family there is an unsaved change
+                        // (and it is easier to mark it as such than to complicate the logic in the load code)
+                        var markAsNotSaved = function() {
```
marta-: Why is one function named `markAsNotSaved` and one `markAsUnsaved`? Looks messy.

allasm:
@sashaandjic the problem you described is PT-3710

components/pedigree/resources/src/main/resources/pedigree/pedigree.js
allasm: Will rename, just because one is a one-time-use local variable I did not give its name much thought

sashaandjic:
@allasm the problem was not present in master built on July 3rd and 4th. Please double check.

allasm:
@sashaandjic yes, you are right, because the bug you found as introduced in PT-3749, which got merged after July 4th (I'm surprised we did not find it at that time - I think we need to think how to add automated pedigree tests). But conceptually it is related to PT-3710 and is very similar in effect to other things PT-3710 describes, so it mde sense to just do it as part of 3710

PR #2700, opened 2018-07-06
PT-3710 Pedigree import should not import patient records which are not already in the family, but keep those who are

allasm:


components/pedigree/resources/src/main/resources/pedigree/pedigree.js
```diff
@@ -731,20 +731,21 @@ define([
...
 
         /**
-         * Returns the list of {id: "...", name: "...", identifier: "..."} of all the patients which
-         * were part of this patient's family at the time pedigree was last (re-)loaded
-         * @method getFamilyMembersBeforeChanges
+         * Returns the set of PhenoTips patient IDs as {id1: true, id2: true} object
+         * @method getAllLinkedPatients
          * @return {Object}
          */
-        getFamilyMembersBeforeChanges: function() {
```
allasm: note: this method was no longer used

components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js
```diff
@@ -1853,13 +1854,27 @@ define([
                 importData.unlinkedMembers = {};
             }
 
-            // FIXME: for now, we do not allow linking patients via import dialogue, so remove all links to all patients
+            // TODO: for now, we do not allow linking patients via import dialogue, so remove
+            //       all links to all patients records that are not laready in the family
```
sdumitriu: Typo in `lalready`

components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js
```diff
@@ -1853,13 +1854,27 @@ define([
...
             if (importOptions && importOptions.doNotLinkPatients) {
-                unlinkedMembers = [];
+                // remove all unlinked patients loaded form import data that are not already in family:
+                for (var patientRecordId in importData.unlinkedMembers) {
+                    if (importData.unlinkedMembers.hasOwnProperty(patientRecordId)) {
+                        if (!editor.isFamilyMember(patientRecordId)) {
```
veronikaslc: why not use && to combine two `if`s?

components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js
```diff
@@ -1853,13 +1854,27 @@ define([
...
+                        }
+                    }
+                }
+                // remove all linked patients loaded form import data that are not already in family:
                 var personIDs = importData.baseGraph.getAllPersons(true);
                 for (var p = 0; p < personIDs.length; p++) {
-                    delete importData.baseGraph.rawJSONProperties[personIDs[p]].id;
-                    delete importData.baseGraph.properties[personIDs[p]].phenotipsId;
+                    var patientRecordId = importData.baseGraph.properties[personIDs[p]].phenotipsId;
+                    if (patientRecordId !== undefined && !editor.isFamilyMember(patientRecordId)) {
```
veronikaslc: why not use `!patientRecordId` instead of `patientRecordId !== undefined`?

allasm: I guess I wrote that because if `variable` == 0 ` `if (variable)` fails. But patient IDs are assumed to be strings, and if `variable == "0"` `if (variable)` works, so we can use just `if (patientRecordId && ...` here

...The funny thing is, despite the difference in how `if` handles `"0"` and `0` above, `"0" == 0`

components/pedigree/resources/src/main/resources/pedigree/model/dynamicGraph.js
allasm: to distinguish between technical `.hasOwnProperty()` check for object properties that has to be done in javascript, and actual application logic. If `hasOwnProperty` were applied because of application logic I would combine the two

sashaandjic:
tested

PR #2701, opened 2018-07-06
PT-3709: Update add-to-family API to also modify pedigree JSON

allasm:
note: PT-3719 "Write a migrator which updates pedigree JSONs to include all family members" may be added to this PR

### DIFF
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
@@ -48,6 +48,8 @@ public class DefaultPedigree extends AbstractBasePedigree implements Pedigree
 
     private static final String PATIENT_DATA_NAMES_LASTNAME_KEY = "last_name";
 
+    private static final String PATIENT_NOTINPEDIGREE_KEY = "notInPedigree";
+
     /**
      * Create a new default pedigree from data (in "old internal" format) and image (a text representing SVG).
      *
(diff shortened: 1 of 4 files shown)


components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
```diff
@@ -108,6 +110,47 @@ public static boolean isSupportedPedigreeFormat(JSONObject data)
...
+    {
+        int maxExistingPedigreeID = 0;
+
+        // go through all members to find first unused ID, and to check if the patient
+        // is already in the pedigree
+        JSONArray members = (JSONArray) this.data.opt(PEDIGREE_JSON_MEMBERS_KEY);
+        for (Object nodeObj : members) {
+            JSONObject node = (JSONObject) nodeObj;
+
+            maxExistingPedigreeID = Math.max(maxExistingPedigreeID, node.optInt(PATIENT_NODEID_JSON_KEY, 0) + 1);
```
sdumitriu: Why the `+1`? If it's supposed to be the new id for the next patient, then the name is wrong, and if it's supposed to be the existing id, then incrementing here is wrong.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
```diff
@@ -108,6 +110,47 @@ public static boolean isSupportedPedigreeFormat(JSONObject data)
...
+
+            JSONObject properties = (JSONObject) node.opt(PATIENT_DATA_JSON_KEY);
```
sdumitriu: There's `optJSONObject` which doesn't require a cast.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
```diff
@@ -108,6 +110,47 @@ public static boolean isSupportedPedigreeFormat(JSONObject data)
...
     }
 
+    @Override
+    public void addLink(String patientId)
+    {
+        int maxExistingPedigreeID = 0;
+
+        // go through all members to find first unused ID, and to check if the patient
+        // is already in the pedigree
+        JSONArray members = (JSONArray) this.data.opt(PEDIGREE_JSON_MEMBERS_KEY);
```
sdumitriu: `opt` is for when the value may be null. If this can indeed be null (unlikely, a pedigree should have at least one node), then check for null. If not, then it's simpler to use `getJSONArray` instead of `opt` + casting.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
```diff
@@ -108,6 +110,47 @@ public static boolean isSupportedPedigreeFormat(JSONObject data)
...
+        for (Object nodeObj : members) {
+            JSONObject node = (JSONObject) nodeObj;
+
+            maxExistingPedigreeID = Math.max(maxExistingPedigreeID, node.optInt(PATIENT_NODEID_JSON_KEY, 0) + 1);
+
+            JSONObject properties = (JSONObject) node.opt(PATIENT_DATA_JSON_KEY);
+            if (properties == null || properties.length() == 0) {
+                continue;
+            }
+            Object id = properties.opt(DefaultPedigree.PATIENT_DATA_PHENOTIPSID_KEY);
```
sdumitriu: Maybe `optString` would be better?

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/DefaultPedigree.java
allasm: agree with all JSON manipulation comments, those were copied form some existing code, I will update both places

PR #2702, opened 2018-07-06
PT-3757: Add ability to specify selected controllers in patient GET

danielpgross:


danielpgross:
Ah, I was not aware of the inconsistent controller behaviour when I implemented this. After thinking through this a bit more, I thought of an alternative way to approach the implementation. It would be much simpler and would totally remove the need for far-reaching code refactoring in the future.

The approach would be to invoke patient.toJSON() with no argument, exactly as is done currently, without the changes in this PR. Then, after the complete JSON has been generated (with all controllers and fields), separately go through and remove the key value pairs from the JSON according to the includeField arguments or excludeField arguments. So the JSON would be post-processed entirely separately from any controller logic.

Advantages of this approach:
Refactoring controllers is not necessary
Every field (i. e. key) in returned JSON can be controlled via includeKeys and excludeKeys (not just top-level controller keys)

What does everyone think?

allasm:
Hmm, IMHO this works well to produce the JSON with only the desired fields

...but all the controllers would have to do their job anyway, so the question is was the slowness that initiated this issue in the communication (sending large JSON over TCP) or the controller(s) needing to read/process some files?

allasm:
However note that controller names do not always match JSON key names, ao this would add yet another field naming scheme in public APIs. E.g. push u ctionality which (currently) uses controller names and its own endpoint wont be easily portable to use this API (though it would take some work in either case)

On the plus side, users can get a JSON once, see the fids they need (or dont need) and request those, not having to figure out what those are called internally, so actually this is a good idea after all IMHO

danielpgross:
>...but all the controllers would have to do their job anyway, so the question is if the slowness that initiated this issue in the first place is because of the communication delays (sending large JSON over TCP) or the controller(s) needing to read/process some files?

This is a very good point. I just tested and found that for an example patient whose record is about 130MB of JSON (including base64-encoded attachments), the time-to-first-byte (TTFB) is over 6 seconds. Based on this, I assume that the JSON-filtering approach would take approx. the same amount of time to generate the response for that patient, even with the attachment keys filtered out of the response. >6s is still too long for a patient GET request.

I discussed with @buske and we decided that it's best for now to stick with the original approach, with @sdumitriu's suggested changes. 

>On the plus side, users can get a JSON once, see the fields they need (or don't need) and request those, not having to figure out what those are called internally, so actually this is a good idea after all IMHO (which will kepp us free to change xontroller names as long as JSON stays the same, which is also a plus)

I think we should keep this in mind for the future. When (if) the refactor happens, it would probably make sense to make it required that the controller name (returned by `getName()`) match the root-level JSON key. That would ensure that controllers can be switched on/off in request arguments based on the set of keys in the returned JSON; you wouldn't need to look up controller names in order to utilize this functionality.

allasm:
yet another possible (temporarily) approach is to add optional  'excludedFields` parameter to the `toJSON()` call, so that `toJSON` will be able to duplicate what we ned from `REST` directly

danielpgross:
@a-gvo completed the implementation in a way that addresses @sdumitriu concerns. The implementation in this PR should be replaced with that implementation. See here: https://github.com/Gene42/phenotips/pull/2

PR #2703, opened 2018-07-06
PT-3704: The id uniqueness is not checked when a patient record is populated from the pedigree editor

veronikaslc:
Done.


sashaandjic:
tested in 1.4-rc-1. works as expected.

PR #2713, opened 2018-07-17
PT-3773: When the pedigree thumbnail size is medium or large, family members list overlaps it and pedigree actions are displayed at the bottom on narrow screens

marta-:


sashaandjic:
one small suggestion: move family members little bit down

family members list

proposed look:
screen shot 2018-07-17 at 2 30 43 pm

marta-:
Comments addressed in the latest commit. Removing the margin completely makes things look bad when the family name extension is disabled. I just set the same margin for both pedigree wrappr and gamily members.

marta-:
@sdumitriu: according to @sashaandjic the second commit also fixes PT-3776. Please make a reference when you merge.

PR #2714, opened 2018-07-18
PT-3756: Always show the close button in the pedigree editor Template/Import dialog

sdumitriu:


components/pedigree/resources/src/main/resources/pedigree/view/tabbedSelector.js
```diff
@@ -98,18 +98,25 @@ define([
...
-            } else {
-                // show close button and enable keyboard close shortcut
-                $(this.dialog.getBoxId()).down('.msdialog-close').show();
-                this.dialog.registerShortcuts("close");
+            var closeButton = this.dialog.dialog.down('.msdialog-close');
+            var closeFunction = this.hide.bindAsEventListener(this);
+            if (!stayInEditorOnClose) {
+                closeFunction = function(event) {
+                    event.stop();
+                    new Ajax.Request(XWiki.currentDocument.getURL('delete', 'confirm=1&ajax=1&form_token=' + $$('meta[name="form_token"]')[0].content), {
```
allasm: Hmm, but what if there is a family with no pedigree. A user opens pedigree editor and clicks close, the expected action is not to delete the family but to just return to the family page

marta-: Right, good point, thanks @allasm . Can you confirm that we'll know we're coming from a patient record page if there's `new_patient_id=P#######` in the url?

marta-: Also, @allasm , would it be better to use [`getFamilyRemoveURL`](https://github.com/phenotips/phenotips/blob/master/components/pedigree/resources/src/main/resources/pedigree/externalEndpoints.js#L74) from external endpoints instead of this hand-crafted deletion URL?

allasm: 1) `new_patient_id` will be there if a patient is added to either  anew family or existing, but if the patient is added to an existing family the template/import dialogue will not be presented

2) yes, I think it is better to use `getFamilyRemoveURL`

marta-: Actually, the dialog will appear for an existing family without pedigree. This apparently small task is starting to involve potentially dangerous changes, and it's too late in 1.4 for them. I'm postponing it until the next version.

allasm: "the dialog will appear for an existing family without pedigree" - yes, but in that case there will be no `new_patient_id` so nothing should be removed. The family should be removed if both dialog pops up and `new_patient_id` exists. But editor should close in any case. +1 for postponing

components/pedigree/resources/src/main/resources/pedigree/view/tabbedSelector.js
```diff
@@ -98,18 +98,25 @@ define([
...
+                        'method': 'GET',
+                        'onSuccess': function() { history.go(-1); }
```
allasm: ...arguably `history.go(-1)` can be done even if family removal fails? Possibly after a dialog warning about that?

PR #2719, opened 2018-07-22
PT-3759: The pedigree editor still allows duplicate external IDs

allasm:
@sashaandjic please check error messages, they should (probably) be reworded

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/externalIdManager.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/externalIdManager.js
@@ -1,0 +1,150 @@
+/**
+ * A class designed to keep track of all external IDs assigned to all patient records in the pedigree (but not pedigree nodes).
+ *
+ * The class is also used as a cache of known external IDs, to avoid constant calls to the back-end if e.g. all IDs start with the
+ * same prefix - the manager will know after the first query that the prefix (as wel las all partially completed versions of it)
+ * is valid or invalid and will be able to report that instantly.
+ *
+ * @class ExternalIdManager
+ * @constructor
+ */
+
+define([
+        "pedigree/model/helpers"
+    ], function(
+        Helpers
+    ){
+    var ExternalIdManager = Class.create( {
+        initialize: function() {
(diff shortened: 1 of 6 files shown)


marta-:
@allasm Unless I am misunderstanding your concern, PT-3742 should be sufficient.

sashaandjic:
Another problem: 
- create a new patient
- open pedigree editor; pick template with many nodes
- type existing ExternalID on more than one node that has no patient assigned 
- for each node click on "Create new"; the new patient record will be created and an error "This identifier already exists" will be presented
- move on to another node with the same ID and repeat the process
- this way you can create many patient records with identical external ID; pedigree editor will save with no error
screen shot 2018-07-23 at 1 52 51 pm

screen shot 2018-07-23 at 1 53 55 pm

screen shot 2018-07-23 at 1 54 40 pm

allasm:
@sashaandjic all comments/bugs have been addressed

sashaandjic:
i'm still able to create multiple records with same identifier within Pedigree editor.
- create patient
- click on pedigree thumbnail and select child/mother/father template
- click on any parent node; type "abc" into identifier field and click on "Create new" bttn
- click on other parent node ; type "abc" into identifier field and click on "Create new" button
- click on Save and exit the pedigree
- there are two patient records created with the same ID

components/pedigree/resources/src/main/resources/pedigree/externalIdManager.js
```diff
@@ -0,0 +1,154 @@
...
+        },
+
+        duplicateIDPresent: function()
+        {
+            for(var id in this.externalIDs) {
+                if (this.externalIDs.hasOwnProperty(id)) {
+                    if (this.externalIDs[id].length > 1) {
+                        var involvesExternalPatient = false;
+                        if (Helpers.arrayContains(this.externalIDs[id], this._EXTERNAL_PATIENT)) {
+                            involvesExternalPatient = true;
```
veronikaslc: `var involvesExternalPatient = Helpers.arrayContains(this.externalIDs[id], this._EXTERNAL_PATIENT);`

PR #2720, opened 2018-07-22
PT-3750: Misleading error message when assigning record with view only rights to a pedigree

allasm:


allasm:
"Insufficient permissions to ADD patient record" may be more correct, but I think it is more clear how to resolve "Insufficient permissions to EDIT", since there are no "add" permissions anywhere in the UI, while there are "edit" permissions (so e.g. a user can ask someone else to give them edit rights). I think we can get back to wording later when we rework patient rights in pedigree in PT-3769

marta-:
Actually "modify" would be better than "edit".

PR #2722, opened 2018-07-25
PT-3807: The "Group Shared Cases" table doesn't include open cases for non-members

sdumitriu:


components/users/ui/src/main/resources/PhenoTips/GroupSharedCases.xml
```diff
@@ -92,7 +92,14 @@ $xwiki.ssx.use('XWiki.DateTimePicker')##
 #set ($filterWhere = "and collaboratorObj.name = doc.fullName and collaboratorObj.className = 'PhenoTips.CollaboratorClass' and collaborator.id.id = collaboratorObj.id and collaborator.value = '${escapetool.sql($doc.documentReference)}'")
 #if (!$isMyGroup)
   #set ($filterFrom = "${filterFrom}, BaseObject accessObj, StringProperty access")
-  #set ($filterWhere = "${filterWhere} and accessObj.name = doc.fullName and accessObj.className = 'PhenoTips.VisibilityClass' and access.id.id = accessObj.id and access.value = 'public'")
+  #set ($allowedVisibilities = [])
+  #foreach ($visibility in $services.permissions.listVisibilityOptions())
+    #if ($visibility.getPermissiveness() &gt;= 50)
```
marta-: Please add an explanation for the magic number.

sdumitriu: Updated the code to be more clear.

PR #2725, opened 2018-07-27
PT-3810: Remove ctakes from the PhenoTips deployment

a-gvo:


sashaandjic:
tested
http://localhost:8080/ctakes/ option is removed from Administration: Text mining

PR #2727, opened 2018-07-29
PT-3809: Export to PDF fails for non-admin users

sdumitriu:


sashaandjic:
I had to use -Pquick option.

PR #2729, opened 2018-08-10
PT-3812: Saving pedigrees fails when family members have medical reports attached

allasm:
Fixed by defaulting to including only a link to an attachment in Patient.toJSON() (for medical reports, images and "additional documents")


allasm:
@sashaandjic Note that push also needs to be tested

allasm:
The limit on the filesize when pushinv is just another manifestation of the same root issue (just in this case it is not a crash but a failure to push). We will probably have to rework how pushing files work - I will create an issue for that

PR #2732, opened 2018-08-23
PT-3834: Bug and style fixes for pedigree editor

buske:
Most of these were identified via eslint. Some are quite serious, e.g., undefined variables.


components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
```diff
@@ -248,7 +250,7 @@ define([
               values.push({'value': v, 'text' : ("0" + v).slice(-2)});
           }
       } else {
-          for (var v = end; v <= start; --v) {
+          for (var v = start; v >= end; --v) {
```
buske: If the loop should really start at `end`, then it needs to be: `for (var v = end; v <= start; ++v) {`

components/pedigree/resources/src/main/resources/pedigree/model/import.js
```diff
@@ -1561,7 +1560,7 @@ define([
                                    properties["comments"] = "(untested)\n" + properties["comments"];
                                }
                                break;
-                           case "O":
+                           case "C":
```
buske: "O" is already a case in this switch statement, and "C" seems like the best fit based on the documentation.

components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
```diff
@@ -13,6 +14,7 @@ define([
         "pedigree/view/ageCalc"
     ], function(
         PedigreeDate,
+        DynamicPositionedGraph,
```
allasm: `DynamicPositionedGraph` is only used for (old) testing/.debug code, the correct change is to move the code that uses it

components/pedigree/resources/src/main/resources/pedigree/model/positionedGraph.js
```diff
@@ -3156,38 +3156,17 @@ define([
 
         //=[position]========================================================================
 
-        displayGraph: function(xcoord, message)
```
allasm: this is used when debugging layout algorithms (I have som etest code in another epository). I would keep it

components/pedigree/resources/src/main/resources/pedigree/view/datepicker.js
allasm: change is correct

components/pedigree/resources/src/main/resources/pedigree/model/import.js
allasm: looks correct

buske:
@allasm You suggested moving your test code somewhere else, but I'm not sure where you'd like it. I went ahead and removed it, so the dependency it had could also be removed. You're welcome to add back the test code wherever you want it.

allasm:
Sorry, I was actually going to do all of that myself, did not expect you to go back to this PR. I will do it after Christmas break

PR #2736, opened 2018-09-09
PT-2205: Ability to manually rearrange nodes within the same generation in the pedigree

allasm:


allasm:
@veronikaslc ready for codereview

sashaandjic:
tested. it is easy to move around siblings and spousal nodes. the only complain that i have is losing focus from the node upon move.

sashaandjic:
Brittney had a quick look at the new feature. She would prefer to have the ability to drag the nodes. If this is impossible, than the current implantation with fixed focus problem after the move would be OK.

buske:
I built this branch from scratch but didn't see any new functionality in the pedigree editor. Do I need to do something special to enable this functionality?

buske:
- If it's not too hard, having CSS that animates the node movement would make it much less disorienting
- This behaves a bit strangely with twins. I would not have expected twins to separate (or other nodes to move between twins), since this makes it look like triplets.
- This behaves very strangely with descendants deeper in the tree, where space separates different branches:

nov-17-2018 16-52-56

sashaandjic:
I am assuming that rearranging children from more than one marriage is a feature? 
step1
step2
step3

allasm:
Yes. This PR is about rearranging any kind of nodes, it was never restricted to rearranging children (which btw in some rare cases may have been spaced far apart by the auto-layout algorithm)

sashaandjic:
1. Automatic layout does not handle nodes with life statuses other that alive and deceased - this could be due to https://phenotips.atlassian.net/browse/PT-4018.
Pregnancies and Automatic layout
 
2. twin/triples/etc nodes should always kept together so the situation below cannot happen
keep twins together
keep twins together 2

PR #2737, opened 2018-09-12
PT-3780: Event Listeners that set permissions or puts records in studies based on preferences; PT-3830: Correct study selection dialog pop up for New Patient button wrt default study preference; PT-3831: Correct study selection dialog pop up the Pedigree Editor wrt default study preference

veronikaslc:


veronikaslc:
Closed in favour of aggregated [PR 2772](https://github.com/phenotips/phenotips/pull/2772)

PR #2740, opened 2018-09-17
PT-3840: Server logs are flooded with "deprecated parseContent" warnings

sdumitriu:


veronikaslc:
Noticed, files are backported from xwiki.

PR #2744, opened 2018-09-27
PT-3842: Disable livetable column header filters

marta-:


danielpgross:
Looks good to me. (I have not tested locally)

As @buske raised earlier, there are column filters disabled by this change that are not present in the filter panel, so those columns will no longer be filterable at all. For "Browse patients", those columns are:
- First name
- Last name
- Family ID
- Internal ID

Should those be added to the filter panel as part of this PR?

marta-:
Yes, they should. That's why the issue status is Paused rather than Code review. I should not have opened the PR before making that change too, sorry for the confusion.

marta-:
Cleaning up PRs: I'm closing this since the solution is incomplete.

PR #2752, opened 2018-10-02
PT-3845: Changing a vocabulary source doesn't work

sdumitriu:


sdumitriu:
One thing I still don't like is that the page reloads after changing a vocabulary location.

PR #2753, opened 2018-10-03
[release] Updated vocabularies

sdumitriu:


sashaandjic:
Path for ONCOTREE should be http://oncotree.mskcc.org/api/tumorTypes?version=oncotree_2018_08_01

sdumitriu:
Because we're not actively using oncotree, it is there in case we ever want
to use it for cancers instead of hpo. And it fails because they often
change it, either the location, the format, or the layout of the ontology.

>

PR #2754, opened 2018-10-06
[misc] Dependency updates

sdumitriu:


sashaandjic:
@sdumitriu Would you please give me instructions as how to test this PR?

sdumitriu:
Build, start, do some simple activities (create and browse patients, json import export, pedigree export) and make sure nothing is broken.

components/consents/api/pom.xml
```diff
@@ -34,6 +34,10 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
```
allasm: according to https://github.com/findbugsproject/findbugs `FindBugs is now SpotBugs` - since we are updating dependencies should we try to move to `SpotBugs` instead?

sdumitriu: There is no newer version of this jar under the spotbugs name, so there is nothing to switch to.

PR #2758, opened 2018-10-16
PT-3858: Administrators cannot filter push events in audit logs

marta-:


marta-:
Agreed, and doing it a bit better right now,

PR #2759, opened 2018-10-16
PT-3859: Refactor audit logs UI to make it reusable for when a specific filter is set

marta-:
This isn't exactly done-done, but it's good enough for using like this:

```
http://localhost:12345/P0000001?sheet=PhenoTips.AuditLogs&action=push&hideActionFilter=1
```
which is what we needed it for in the first place.

### DIFF
--- a/components/security/audit/frontend/src/main/resources/PhenoTips/AuditFilters.xml
+++ b/components/security/audit/frontend/src/main/resources/PhenoTips/AuditFilters.xml
@@ -40,6 +40,56 @@
   <content>{{velocity output=false}}
 #set ($dateFormat = "MM/dd/yyyy")
 ##
+## PROCESS THE REQUEST FILTERS
+##
+## TIMESTAMP
+##
+#set ($fromTimeFilter = $request.fromTime)
+#set ($toTimeFilter = $request.fromTime)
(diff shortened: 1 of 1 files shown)


components/security/audit/frontend/src/main/resources/PhenoTips/AuditFilters.xml
```diff
@@ -56,24 +106,43 @@
...
+                                                                           #end
                                                                           &lt;/select&gt;{{/html}}
+#end
+##
+## RECORD FILTER
+##
+#if ($hideRecordFilter)
+* {{html clean=false}}&lt;input type="hidden" name="record" value="$!{recordFilter}"/&gt;{{/html}}
+#else
 * $services.localization.render('phenotips.auditFilters.entity') {{html clean=false}}&lt;input type="text" class="suggestEntity" value=""/&gt;{{/html}}
```
sdumitriu: Weird that this input doesn't have a `name`...

marta-:
@sasha IIRC when @sdumitriu implemented auditing push events he found no elegant way to obtain information on whether the push was successful. I do agree with your observation, so please create a task.

PR #2766, opened 2018-11-06
PT-3874: Allow disabling filters on the All Families table

sdumitriu:
Done


marta-:
I didn't like the `if`s in the patient directory filters code and I don't like them here either. I hope there can be a nicer way to do this.

sdumitriu:
There will be once each filter is defined in an extension, but right now this is the best option without a major rewrite.

PR #2772, opened 2018-11-07
PT-3779: Ability to associate specific permissions or studies to records automatically based on user/group preferences

veronikaslc:
UI and backend to allow users to specify default settings for owner, collaborators, visibility, and studies for their records.


marta-:
The layout is broken when setting up default permissions for a study.

image

(Environment: Chrome 70/Linux)

components/entity-access-rules/api/src/main/java/org/phenotips/data/permissions/internal/DefaultEntityPermissionsPreferencesManager.java
```diff
@@ -0,0 +1,349 @@
...
+    {
+        DocumentReference currentUserRef = entity == null ? this.helper.getCurrentUser() : entity;
+        // Guest user or invalid entity
+        if (currentUserRef == null) {
+            return null;
+        }
+
+        String studyID = getDefaultPreferenceStringValue(currentUserRef, DEFAULT_STUDY_PROPERTY_NAME);
+        if (studyID != null) {
+            return this.userOrGroupResolver.resolve(studyID);
```
marta-: Same comment about using something called `userOrGroupResolver` to resolve an entity that is not supposed to be a user or a group.

marta-:
Because the study update event is not fired at the right time, the following scenario doesn't work:

* Mary Green sets Study C as her default study.
* Study C has George Red as default collaborator.
* Expectation: when Mary creates a new patient record, it is automatically placed in Study C and George is added as collaborator
* Reality: when Mary creates a new patient record, it is automatically placed in Study C BUT George is NOT added as collaborator

This scenario is actually the main reason why we implemented the whole thing, so that we can do the proper study/collaborators assingnment for C4R cases in PC.

PR #2775, opened 2018-11-07
PT-3879: Mark PhenoTips v1.5 Patient JSONs as supported in pedigree editor

allasm:


sdumitriu:
Are there changes needed for the older versions to support 1.5 jsons?

allasm:
No, pedigree only sees PhenoTips JSONs it gets from the host PT instance. Even if a JSON from some other version is imported (e.g. via REST) it gets converted to internal PT representation, and when pedigree asks for it it will always be the "host" version

it may get a bit more complicated with family REST (since patient JSON is part of family JSON), but that will only happen in 1.5, so nothing to update in older versions as well

PR #2779, opened 2018-11-09
PT-3882: ContactInfo does not parse a string with multiple emails

veronikaslc:
Fixed.


components/patient-contacts/src/test/java/org/phenotips/data/ContactInfoTest.java
```diff
@@ -36,6 +37,8 @@
 
     private static final String USER_EMAIL = "padams@hospital.org";
 
+    private static final String USER_EMAILS = "padams@hospital.org,rodens@hospital.org";
```
allasm: I would prefer `rodents@hospital.org` ;)

PR #2781, opened 2018-11-14
PT-3076: Upgrade to XWiki 10

sdumitriu:
Also includes:
PT-3861: Upgrade org.restlet


marta-:
This PR is a beautiful example of monster changes that are easy to follow thanks to well organized commits @veronikaslc @allasm . We should show it as an example to new hires.

marta-:
In the "Jump to page" modal the suggestions are not fully visible because the box now has `overflow: auto`. I suggest adding this somewhere:
```
.xdialog-box.jump-dialog {
    overflow: visible;
}
```

marta-:
When starting PhenoTips, the message displayed is "XWiki is initializing.." instead of "PhenoTips is initializing...". That is because they appear to have changed the translation keys from `platform.web.init.message.*` to `platform.web.init.message.wiki.*`.

sashaandjic:
once new user is successfully added the ui looks strange 
screen shot 2018-11-16 at 3 07 20 pm

sashaandjic:
reference genome is blank (should be defaulted to Grch37 (hg19) 
screen shot 2018-11-19 at 2 28 58 pm

sashaandjic:
Owner and Collaborator lookups do not work in Advanced filters
screen shot 2018-11-19 at 2 44 23 pm

marta-:
> Owner and Collaborator lookups do not work in Advanced filters

This is the same issue as above: an unescaped `&` from the result wit the id `Mary&Peter` breaks the suggestions.

marta-:
> reference genome is blank (should be defaulted to Grch37 (hg19)

@sdumitriu I think #2783 should fix this.

marta-:
The "`&` in the group name" bug affects the master branch as well, so it's unrelated to the upgrade. I've reported it: https://phenotips.atlassian.net/browse/PT-3891 . @sashaandjic I suggest conducting further testing on data that doesn't expose this bug, i.e. removing the "Mary&Peter" group from the test data and using a more conventional name without special characters for the time being.

components/widgets/ui/src/main/resources/Main/MessageSenderMacro.xml
```diff
@@ -25,12 +25,9 @@
   <defaultLanguage/>
   <translation>0</translation>
   <creator>xwiki:XWiki.Admin</creator>
-  <creationDate>1424473682000</creationDate>
```
veronikaslc: Why not use XWiki's `MessageSenderMacro.xml` entirely? Seems the fix that was the reason of backporting is already included in the release.

sdumitriu:
@sashaandjic @veronikaslc @marta- All comments so far have been addressed, please retest,

marta-:
Now that [PT-3914](https://phenotips.atlassian.net/browse/PT-3914) is merged into master, [`GroupCases.xml`](https://github.com/phenotips/phenotips/blob/5dcfb7f639d3ab90ef6d016da89edb5edffd0274/components/users/ui/src/main/resources/PhenoTips/GroupCases.xml) needs to be updated in this PR to use the new API for group membership:

```
#set ($groupsAndSelf = [$xcontext.userReference.toString()])
#foreach ($i in $services.user.group.getGroupsFromAllWikis($xcontext.userReference))
  #set ($group = $xwiki.getDocument($i))
  #if ($group.getObject('PhenoTips.PhenoTipsGroupClass'))
    #set ($discard = $groupsAndSelf.add($escapetool.sql($group.documentReference.toString())))
  #end
#end
```

marta-:
@sdumitriu Let's merge this please before id diverges too much from master and needs to be redone.

PR #2789, opened 2018-12-04
PT-3901: Study descriptions may break the layout of the study page if the text does not end in a newline

marta-:


marta-:
@sdumitriu Sounds good.

PR #2791, opened 2018-12-10
PT-3903: When a new variant is added in the Genes/Variants table, the variant table title does not display the gene name

veronikaslc:
Fixed


veronikaslc:
In the variant header template [here](https://github.com/phenotips/phenotips/blob/master/components/widgets/resources/src/main/resources/templates/geneVariantMacros.vm#L32) the placeholder for gene name is `GENE_NAME_PLACEHOLDER`, so javascript had incorrect one.

sdumitriu:
Caused by d1df93943

PR #2795, opened 2018-12-12
PT-3907: Default entity name to the id for permissions dialog

veronikaslc:
Also fixed backend to consider that the group may be XWikiGroup and not only PhenoTipGroupClass

### DIFF
--- a/components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
+++ b/components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
@@ -642,9 +642,9 @@
            // to identify whether the collaborator has been already added we compare resolved names
            // to be independent on the entity name format passed by suggest widget or REST service
            // (could be "xwiki:XWiki:Admin", or "XWiki:Admin", or "Admin")
-           var resolvedId = XWiki.Model.resolve(id, XWiki.EntityType.DOCUMENT).name;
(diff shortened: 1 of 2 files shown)


components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
```diff
@@ -1075,7 +1076,7 @@
       var row = new Element('tr', {'class' : (highlight === true ? 'new' : '')});
       row.insert(new Element('td').insert(new Element('span', {'class' : 'fa fa-' + c.type}).update(' ')));
       row.insert(new Element('td')
-              .insert(c.name)
+              .insert(c.name ? c.name : XWiki.Model.resolve(c.id, XWiki.EntityType.DOCUMENT).name)
```
allasm: do we need the same code (as on line 1056 above) here as well, wouldn't that fix fix this too?

veronikaslc: Line 1056 is for permissions detains on top of patioent form, line 1079 is for collaborators suggestions selected in the permissions dialog.
The resolving logic is adopted from studying XWiki code, userpicker.js in particular. I assumed, it's a right approach if its in the platform codebase.

allasm: I guess it is then. I was jusf surprised that is done on the front-end and not om the back-end before even sending data to the front-end

PR #2797, opened 2018-12-14
PT-3914: Inconsistency between the list of cases shown on a group page and list displayed when filtering by that group as owner

marta-:


components/users/ui/src/main/resources/PhenoTips/GroupCases.xml
```diff
@@ -90,19 +90,6 @@ $xwiki.ssx.use('XWiki.DateTimePicker')##
 ##
 #set ($filterFrom = ", BaseObject ownerObj, StringProperty owner")
 #set ($filterWhere = "and ownerObj.name = doc.fullName and ownerObj.className = 'PhenoTips.OwnerClass' and owner.id.id = ownerObj.id and owner.value = '${escapetool.sql($doc.documentReference)}'")
-#if (!$isMyGroup)
```
sdumitriu: Without any permissions check, all inaccessible patients will show up in the table as inaccessible documents (no data except the ID with an asterisk).

marta-: You are right. The issue is addressed in the latest commit.

PR #2800, opened 2018-12-18
PT-3853: Obfuscate pedigree tumbnail if "Warn about sensitive information" is checked

allasm:
Need to be cherry-picked into stable when it is approved

### DIFF
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
@@ -212,8 +212,8 @@ public class PhenotipsFamily extends AbstractPrimaryEntity implements Family
     {
         BaseObject familyObject = getXDocument().getXObject(Family.CLASS_REFERENCE);
         return familyObject.getIntValue(FIELD_HAS_SENSITIVE_DATA) == FIELD_VALUE_CHECKED
-            ? StringUtils.EMPTY
-            : familyObject.getStringValue("warning_message");
+            ? familyObject.getStringValue("warning_message")
+            : StringUtils.EMPTY;
     }
 
     @Override
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
@@ -193,6 +193,13 @@ public class PhenotipsFamily extends AbstractPrimaryEntity implements Family
         return getXDocument().getURL(actions, getXContext());
     }
 
+    @Override
+    public boolean containsSensitiveData()
+    {
+        BaseObject familyObject = getXDocument().getXObject(Family.CLASS_REFERENCE);
+        return (familyObject.getIntValue(WARNING) != 0);
+    }
+
     /*
      * Some pedigrees may contain sensitive information, which should be displayed on every edit of the pedigree. The
      * function returns a warning to display, or empty string
--- a/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
+++ b/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
@@ -60,7 +60,10 @@ public class PhenotipsFamily extends AbstractPrimaryEntity implements Family
     /** Field name in the family document which holds the list of member patients. */
     public static final String FAMILY_MEMBERS_FIELD = "members";
 
-    private static final String WARNING = "warning";
+    private static final String FIELD_HAS_SENSITIVE_DATA = "warning";
+
+    // the value used by XWiki in XObjects to indicate a boolean is set to true/a checkbox is checked
+    private static final int FIELD_VALUE_CHECKED = 1;
 
     private static PatientRepository patientRepository;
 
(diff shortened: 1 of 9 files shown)


components/pedigree/resources/src/main/resources/pedigree/familyData.js
```diff
@@ -33,8 +33,12 @@ define([
                 this.familyMembersIndex[this.familyMembers[i].id] = i;
             }
 
-            this.warningMessage = familyJSON.hasOwnProperty("warning") ?
-                                  familyJSON.warning : "";
+            this.warningMessage = null;
+            if (familyJSON.hasOwnProperty("contains_sensitive_data") && familyJSON["contains_sensitive_data"]) {
+                this.warningMessage = (familyJSON.hasOwnProperty("sensitive_data_message") && familyJSON["sensitive_data_message"] != "")
+                                      ? familyJSON["sensitive_data_message"]
+                                      : "This pedigree was marked as containing sensitive information. No further details were provided.";
```
marta-: Can you please make this translatable?

allasm: there are no (zero) other translations in the pedigree right now. It is a separate issue to introduce translations in pedigree

marta-: Nevermind, i just noticed it's part of `pedigree` where nothing is translatable.

marta-: Sounds good, thanks

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
```diff
@@ -193,6 +193,13 @@ public String getURL(String actions)
         return getXDocument().getURL(actions, getXContext());
     }
 
+    @Override
+    public boolean containsSensitiveData()
+    {
+        BaseObject familyObject = getXDocument().getXObject(Family.CLASS_REFERENCE);
+        return (familyObject.getIntValue(WARNING) != 0);
```
marta-: Just wondering why not `== 1`.

components/pedigree/resources/src/main/resources/pedigree/familyData.js
```diff
@@ -33,8 +33,12 @@ define([
                 this.familyMembersIndex[this.familyMembers[i].id] = i;
             }
 
-            this.warningMessage = familyJSON.hasOwnProperty("warning") ?
-                                  familyJSON.warning : "";
+            this.warningMessage = null;
+            if (familyJSON.hasOwnProperty("contains_sensitive_data") && familyJSON["contains_sensitive_data"]) {
```
marta-: Personally I find the `hasOwnProperty` checks overzealous. I would have written lines 36-41 as:

```
this.warningMessage = familyJSON["contains_sensitive_data"] ? (familyJSON["sensitive_data_message"] || this.DEFAULT_WARNING_MESSAGE) : null;
```

components/pedigree/resources/src/main/resources/pedigree/familyData.js
marta-: Defining texts as properties of the class rather than in code will make the pedigree application easier to translate in the future. I suggest to add
```
DEFAULT_WARNING_MESSAGE : "This pedigree was marked as containing sensitive information. No further details were provided.",
```
just before like 14.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
allasm: Just used the same check already in the code. I'm not 100% sure off the top of my head what xwiki uses for booleans, apparently this check used to work well so reused it. If you know it is 1 i agree `== 1` looks better

components/pedigree/resources/src/main/resources/pedigree/familyData.js
allasm: I think some hasOwnPropert can be removed, but I don't like to rely on pecularities of js such as `""` evaluating to false, it may not be obvious to every developer who looks at this. I may be wrong

marta-: IMO, a dev who looks at this and doesn't know about truthy and falsy values in js should use the opportunity to learn something new.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
sdumitriu: Booleans can have 3 values: 0 means false, 1 means true, -1 means not specified. The warning should only be displayed if it is set to 1, since "unspecified" shouldn't imply there's a warning.

So, change the check to `== 1`.

marta-: @allasm Added this change to #2801 .

sashaandjic:
Sensitive info can be seen on the Family page. http://172.20.4.117:8080/Families/FAM0000001

image

sashaandjic:
I like that we now obfuscate pedigree image even if there is not sensitive info description provided.

components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java
```diff
@@ -201,7 +211,7 @@ public String getURL(String actions)
     public String getWarningMessage()
     {
         BaseObject familyObject = getXDocument().getXObject(Family.CLASS_REFERENCE);
-        return familyObject.getIntValue(WARNING) == 0
+        return familyObject.getIntValue(FIELD_HAS_SENSITIVE_DATA) == FIELD_VALUE_CHECKED
```
sashaandjic: The logic changed here. You used to check if the "has sensitive data" box is unchecked, now you're checking if it is checked. For this reason, the message sent to the frontend is always an empty string and the generic message is displayed to the user.

```suggestion
        return familyObject.getIntValue(FIELD_HAS_SENSITIVE_DATA) != FIELD_VALUE_CHECKED
```

PR #2807, opened 2019-01-16
PT-3930: Patient records entered in PhenoTips <1.4rc2 that has been upgraded to a version >= 1.4rc2 may have an empty PubMed ID

veronikaslc:
Fixed

### DIFF
--- a/components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R74692PhenoTips3930DataMigration.java
+++ b/components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R74692PhenoTips3930DataMigration.java
@@ -76,7 +76,7 @@ public class R74692PhenoTips3930DataMigration extends AbstractHibernateDataMigra
     @Override
     public String getDescription()
     {
-        return "Make Pubmed ID to be a multiple value field.";
+        return "Clean up empty pubmed id entries.";
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R74692PhenoTips3930DataMigration.java
```diff
@@ -0,0 +1,125 @@
...
+     * Resolves unprefixed document names to the current wiki.
+     */
+    @Inject
+    @Named("current")
+    private DocumentReferenceResolver<String> resolver;
+
+    @Override
+    public String getDescription()
+    {
+        return "Make Pubmed ID to be a multiple value field.";
```
marta-: The description is incorrect. Should be something along the lines of "remove empty pubmed id entries"

allasm:
When testing this PR, I used two testcases: one I created via "import JSON" form a JSON with a blank pubmedID. Another patient was exported from `preprod` (which is based on PT-1.4.2 and has the same patient structure) as XAR and then imported into the instance. Then a migrator JAR was added, and the instance restarted. The first case was correctly migrated, the second one was not

veronikaslc:
I have tested both cases in phenotips-standalone-1.4.3 zip from phenotips.org and migrator works as expected on both of them. May it be os thing?

PR #2808, opened 2019-01-17
PT-3927: Push code should handle usernames with extra leading and/or trailing spaces correctly

allasm:


allasm:
> store that username with the token instead of the one from the request.

That is already done

> validateLogin should return a more comprehensive "login outcome" in the JSON, not just the login error, meaning it should include the username that was actually authenticated (read from the Principal object returned by XWiki's authenticate)

I think that is outside of the scope of the bugfix, it also potentially requires UI changes to properly display the error (otherwise there is no practical reason to provide more details)

components/patient-data-sharing/receiver-api/src/main/java/org/phenotips/data/receive/internal/DefaultReceivePatientData.java
```diff
@@ -401,7 +401,7 @@ protected JSONObject validateLogin(XWikiRequest request, XWikiContext context)
                 return generateIncompatibleVersionResponse();
             }
 
-            String userName = request.getParameter(ShareProtocol.CLIENT_POST_KEY_NAME_USERNAME);
+            String userName = request.getParameter(ShareProtocol.CLIENT_POST_KEY_NAME_USERNAME).trim();
```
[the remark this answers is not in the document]
allasm: 1) +1 about NPE, did not think about that

2) I got a bit too lazy to check that. Would never think all spaces would be removed... Will fix

marta-:
>>store that username with the token instead of the one from the request.

>That is already done

My mistake, I should have said "instead of the one *post-processed* from the request. What I mean is: don't duplicate the username processing, just use the one that was authenticated, since it is known.

>> validateLogin should return a more comprehensive "login outcome" in the JSON, not just the login error, meaning it should include the username that was actually authenticated (read from the Principal object returned by XWiki's authenticate)

> I think that is outside of the scope of the bugfix, it also potentially requires UI changes to properly display the error (otherwise there is no practical reason to provide more details)

I disagree, I don't think it's out of scope. The refactoring would allow for a more robust fix and would be compatible with any future changes of the "username cleanup" happening in XWiki. I haven't looked at the code in detail, but I'm surprised you're foreseeing UI changes. Can't you just add a username field in the JSON object when the status is success? how would that affect the UI?

marta-:
@allasm please see here the kind of change I'm suggesting:

https://gist.github.com/marta-/c44081d67143499e9cc06576812261d4/revisions

allasm:
I agree using post-processed name is better, didn't know it is available. Will fix today

allasm:
The latest implementation follows what has been discussed at the dev. meeting. @marta- please review

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatientService.xml
```diff
@@ -109,6 +109,9 @@
     #set ($data.unauthorizedserver = 'true')
   #elseif ($response.isLoginFailed())
     #set ($data.loginfailed = 'true')    ## one form or another of incorrect username, password or stored token
+    #if ($response.isLoginFailed_UsernameNotCanonical())
+      #set ($data.usernameNotCanonical = 'true')
```
marta-: Where is this used?

components/patient-data-sharing/share-protocol/src/main/java/org/phenotips/data/shareprotocol/ShareProtocol.java
```diff
@@ -165,6 +165,10 @@
 
     public static final String SERVER_JSON_KEY_NAME_ERROR_ACTIONFAILED = "action_failed";
 
+    public static final String SERVER_JSON_KEY_NAME_USERNAME_NOT_IN_CANONICAL_FORM = "username_not_canonical";
+
+    public static final String SERVER_JSON_KEY_NAME_USERNAME_SUGGESTED_CANONICAL_FORM = "username_suggested_canonical";
```
marta-: I'm not a fan of these field names. In 1 year we'll be scratching our heads wondering what this is all about. I understand that you're incorporating the terminology used by XWiki, but TBH that's not the best either.

I would name these fields in plain English `provided_username` (or `entered_username`) and `expected_username` (or `deduced_username`), similar to your variable names in `DefaultReceivePatientData.java` lines 230-231.

components/patient-data-sharing/push-ui/src/main/resources/PhenoTips/PushPatientService.xml
allasm: https://github.com/phenotips/phenotips/pull/2808/files#diff-6c0b884a881ac3f49d2dcfe3e4d6f737R619

marta-: Right, thank you.

components/patient-data-sharing/share-protocol/src/main/java/org/phenotips/data/shareprotocol/ShareProtocol.java
allasm: The first key is used to give more details about a login problem, so what about naming it `username_not_accepted`? Possibly with a `true\false` instead of the entered username? (but can be either). And naming the other one `expected_username` like you suggested

sdumitriu:
Closing in favor of #2818

PR #2815, opened 2019-02-05
PT-3939: Include whitespace separator in parsing multiple emails in ContactInfo

veronikaslc:
Added a blank space to possible separators

### DIFF
--- a/components/patient-contacts/src/main/java/org/phenotips/data/ContactInfo.java
+++ b/components/patient-contacts/src/main/java/org/phenotips/data/ContactInfo.java
@@ -225,7 +225,7 @@ public interface ContactInfo
             {
                 List<String> collectedEmails = new ArrayList<String>();
                 for (String email : emails) {
-                    for (String parsedEmail : StringUtils.split(email, ",|;")) {
+                    for (String parsedEmail : StringUtils.split(email, ",; ")) {
                         collectedEmails.add(parsedEmail.trim());
                     }
                 }


components/patient-contacts/src/main/java/org/phenotips/data/ContactInfo.java
```diff
@@ -225,7 +225,7 @@ private void setEmails(List<String> emails)
             {
                 List<String> collectedEmails = new ArrayList<String>();
                 for (String email : emails) {
-                    for (String parsedEmail : StringUtils.split(email, ",|;")) {
+                    for (String parsedEmail : StringUtils.split(email, ",|;| ")) {
```
marta-: Afaik the second parameter of `split` which specifies the separators is treated as a list of characters, not a regular expression. You should just add the space, there's no need for another `|`.

allasm:
PT-3882 is closed and included in some already released versions. I think this should be under  anew PT issue, not PT-3882

components/patient-contacts/src/main/java/org/phenotips/data/ContactInfo.java
allasm: I think Marta is right (https://github.com/phenotips/phenotips/pull/2815#issuecomment-460821455), and I would remove `|` altogether as a highly unlikely separator

veronikaslc: `|` serves as an `OR` in this case

allasm: But it looks like it does not, from the specification of this function

veronikaslc: will remove it then

veronikaslc:
Closed in favour of https://github.com/phenotips/phenotips/pull/2816

PR #2819, opened 2019-02-07
PT-3927-stable-1.3.x: Push code should handle usernames with extra leading and/or trailing spaces correctly

veronikaslc:


components/patient-data-sharing/receiver-api/src/main/java/org/phenotips/data/receive/internal/DefaultReceivePatientData.java
```diff
@@ -151,6 +152,9 @@
     @Inject
     private ConsentAuthorizer consentAuthorizer;
 
+    @Named("user/current")
```
marta-: Why is this not `@Inject`ed? @veronikaslc

PR #2823, opened 2019-02-26
PT-3945-master: The variant's cDNA field may contain extra spaces and wrong prefixes

veronikaslc:
Done

### DIFF
--- a/components/patient-data/impl/src/main/java/org/phenotips/data/internal/VariantCdnaCleanupEventListener.java
+++ b/components/patient-data/impl/src/main/java/org/phenotips/data/internal/VariantCdnaCleanupEventListener.java
@@ -1,0 +1,81 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


components/patient-data/impl/src/main/java/org/phenotips/data/internal/VariantCdnaCleanupEventListener.java
```diff
@@ -0,0 +1,81 @@
...
+        List<BaseObject> variantXWikiObjects = doc.getXObjects(VARIANT_CLASS_REFERENCE);
+
+        if (variantXWikiObjects == null || variantXWikiObjects.isEmpty()) {
+            return;
+        }
+
+        variantXWikiObjects.stream().filter(Objects::nonNull)
+            .filter(variant -> !StringUtils.isBlank(variant.getStringValue(VARIANT_KEY)))
+            .forEach(variant -> {
+                String cdna = variant.getStringValue(VARIANT_KEY).trim().replaceFirst("^C\\.", "c\\.");
```
sdumitriu: There is no need for the `\\` in the replacement string, since that is not a regular expression.

sdumitriu: Also change on the stable branch.

veronikaslc: done

marta-:
I suggest that the cDNA input be corrected realtime on blur (i.e. remove spaces and make "c." lowercase) to "educate" the user regarding how they're supposed to enter this data.

PR #2828, opened 2019-03-11
PT-3959: "Tested Negative/Rejected Candidate" Gene breaks Pedigree Editor

BrianLiSK:
Bug fix.

`editor.getGeneLegend(legendType)` will return `null` if there is no associated legend with the gene. Examples would be statuses of "Tested Negative" and "Rejected Candidate". We need to check for a null value before calling `.getGeneColor()`

I tested this small fix on my browser directly using dev tools. Please test on VM (once deployment completes): http://172.20.4.111:8080/


marta-:
VM for testing: http://172.20.4.111:8080

PR #2835, opened 2019-03-21
PT-3942: Profile urls should be generated on the backend

veronikaslc:
Added url to the UserSummary object factory


components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
```diff
@@ -979,7 +979,7 @@
       new Ajax.Request(this._serviceURL + "&amp;xaction=user-access&amp;id=" + this.recordId, {
         onSuccess : function(response) {
            var data = response.responseJSON;
-           level = data.level.name;
+           level = data.level;
```
allasm: confirmed with @veronikaslc that this is intended

PR #2849, opened 2019-04-15
PT-3974: Incorrect access rights shown in UI

veronikaslc:
Fixed


BrianLiSK:
Maybe we should have a unit test with Mockito for this situation (as Marta mentioned). But, since you are about to do a major refactor, it might be okay for now.

PR #2850, opened 2019-04-16
PT-3968: Modifying a group name is not reflected on patient/match table

veronikaslc:
Fetch correct group name in OwnerRepresentation

### DIFF
--- a/components/entity-access-rules/rest/src/main/java/org/phenotips/data/permissions/rest/internal/NameAndEmailExtractor.java
+++ b/components/entity-access-rules/rest/src/main/java/org/phenotips/data/permissions/rest/internal/NameAndEmailExtractor.java
@@ -109,7 +109,7 @@ public class NameAndEmailExtractor
         if (groupObject != null) {
             email = groupObject.getStringValue("contact");
         }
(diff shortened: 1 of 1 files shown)


components/entity-access-rules/rest/src/main/java/org/phenotips/data/permissions/rest/internal/NameAndEmailExtractor.java
```diff
@@ -109,7 +109,7 @@
         if (groupObject != null) {
             email = groupObject.getStringValue("contact");
         }
-        String name = document.getDocumentReference().getName();
+        String name = document.getTitle();
```
marta-: What if the `title` field is missing? `getTitle()` does not default to the document name, while `getDisplayTitle()` does.

sdumitriu: `getDisplayTitle` (and `getPlainTitle`) only exist in the scriptable `Document` API, not in `XWikiDocument`. But yes, the problem is real.

PR #2851, opened 2019-04-17
PT-3968: Modifying a group name is not reflected on patient/match table

sdumitriu:
Bulletproofing fix


components/entity-access-rules/rest/src/main/java/org/phenotips/data/permissions/rest/internal/NameAndEmailExtractor.java
```diff
@@ -109,7 +115,7 @@
         if (groupObject != null) {
             email = groupObject.getStringValue("contact");
         }
-        String name = document.getTitle();
+        String name = document.getRenderedTitle(Syntax.PLAIN_1_0, this.xcontextProvider.get());
```
marta-: TBH I would have done something like this, to avoid introducing a dependency on XWikiContext:
```suggestion
        String name = StringUtils.isBlank(document.getTtile()) ? : document.getName() : document.getTitle()
```
(not tested)

PR #2857, opened 2019-04-24
PT-3979 JS injection via group names

BrianLiSK:
- JS did not escape HTML for names. Use PrototypeJS's `.escapeHTML()`
- Use `getDisplayTitle()` which is already escaped, instead of `plainTitle` for group names

These are front end changes and need to be tested more thoroughly.

### DIFF
--- a/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
+++ b/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
@@ -128,7 +128,7 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
       avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
     }
     container.insert(avatarWrapper);
-    var userName = source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info;
+    var userName = (source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info).escapeHTML();
     container.insert(new Element('div', {'class': 'user-name'}).update(userName));
     var referenceWrapper = new Element('div');
     var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id;
--- a/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
+++ b/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
@@ -131,7 +131,7 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
     var userName = source.highlight ? this.emphasizeMatches(this.sInput.escapeHTML(), data.info.escapeHTML()) : data.info.escapeHTML();
     container.insert(new Element('div', {'class': 'user-name'}).update(userName));
     var referenceWrapper = new Element('div');
-    var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id;
+    var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id.escapeHTML();
     referenceWrapper.insert(new Element('span', {'class': 'user-alias'}).update(userAlias));
     var userReference = XWiki.Model.resolve(data.value, XWiki.EntityType.DOCUMENT);
     var wiki = userReference.extractReferenceValue(XWiki.EntityType.WIKI) || XWiki.currentWiki;
--- a/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
+++ b/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
@@ -120,6 +120,8 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
 
   // @Override
   createItemDisplay: function(data, source) {
+    data.info = data.info.escapeHTML();
+
     var container = new Element('div', {'class': 'user'});
     var avatarWrapper = new Element('div', {'class': 'user-avatar-wrapper'});
     if (data.icon.indexOf('/') >= 0) {
@@ -128,7 +130,7 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
       avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
     }
     container.insert(avatarWrapper);
-    var userName = (source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info).escapeHTML();
+    var userName = source.highlight ? this.emphasizeMatches(this.sInput.escapeHTML(), data.info) : data.info;
     container.insert(new Element('div', {'class': 'user-name'}).update(userName));
     var referenceWrapper = new Element('div');
     var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id;
--- a/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
+++ b/components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
@@ -120,17 +120,15 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
 
   // @Override
   createItemDisplay: function(data, source) {
-    data.info = data.info.escapeHTML();
-
     var container = new Element('div', {'class': 'user'});
     var avatarWrapper = new Element('div', {'class': 'user-avatar-wrapper'});
     if (data.icon.indexOf('/') >= 0) {
-      avatarWrapper.insert(new Element('img', {src: data.icon, alt: data.info, 'class': 'icon'}));
+      avatarWrapper.insert(new Element('img', {src: data.icon, alt: data.info.escapeHTML(), 'class': 'icon'}));
     } else {
-      avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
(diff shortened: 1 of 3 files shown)


components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
```diff
@@ -128,7 +128,7 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
       avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
     }
     container.insert(avatarWrapper);
-    var userName = source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info;
+    var userName = (source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info).escapeHTML();
```
marta-: `emphasizeMatches` probably adds some HTML tags to format the matching substring differently. Escaping should ve done before this formatting is applied.

BrianLiSK: Yes, you're right. 
I also missed committing the spot after "This record is owned by"
Fixing...

components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
```diff
@@ -120,6 +120,8 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
 
   // @Override
   createItemDisplay: function(data, source) {
+    data.info = data.info.escapeHTML();
```
BrianLiSK: I'm not a fan of this kind of mutation, but its about as close as I can think of escaping before element insertion (without affecting other logic). We have to escape `sInput` right before the `emphasizeMatches()` so that the html tags can be added in.

Still cannot merge this PR as I'm still trying to fix the injection via the initial creation of Work Group. I think the document name is not being escaped.

sdumitriu: There are two preferred locations for escaping:

1. ASAP, when the data is first obtained from the database; in this case it would be in the code that serves the suggestions, serverside
2. ALAP, when the data is placed in the output

The first one has the advantage of being somewhat safer, since it's only one place, so there's no risk of someone later forgetting to escape. But it has the big disadvantage that only one kind of escape is allowed, and in our platform data is usually used in multiple different ways: SQL, JSON, HTML, JavaScript, so it doesn't really work for us.

The second one is more semantically correct, since you escape right where the variable is inserted in content, and the escape method is obvious. The problem is that it is very easy to forget to apply the right escape, which may lead to different kinds of injections.

An escape in between these points makes it difficult to comprehend the code, since one must read all the code to figure out if some variable has already been escaped or not. Oother kinds

components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
```diff
@@ -128,7 +130,7 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
       avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
     }
     container.insert(avatarWrapper);
-    var userName = source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info;
+    var userName = source.highlight ? this.emphasizeMatches(this.sInput.escapeHTML(), data.info) : data.info;
```
sdumitriu: Are you sure the input needs to be escaped, and not the data?

components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
```diff
@@ -1479,7 +1479,7 @@ $xwiki.ssx.use($serviceDocName)
     #if ($firstCollaborator.isUser())
       #set ($collaboratorName = $xwiki.getUserName("$firstCollaboratorReference", false))
     #else
-      #set ($collaboratorName = $xwiki.getDocument($firstCollaboratorReference).plainTitle)
+      #set ($collaboratorName = $xwiki.getDocument($firstCollaboratorReference).getDisplayTitle())
```
sdumitriu: Why this change?

components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
sdumitriu: Ah, the data is already escaped, right at the top where the big comment is. I guess this is the ALAP escaping point.

BrianLiSK: The data gets escaped in the change above.
Without escaping `sInput`, inputting something like `<img ` does not highlight properly (if I remember correctly).

components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
BrianLiSK: `plainTitle` is un-escaped for group name, unlike `getDisplayTitle()`. I had it originally as `$escapetool.html($xwiki.getDocument($firstCollaboratorReference).plainTitle)` right before insertion which also works.

With you're previous comment, following ALAP, perhaps the better place to escape would be right before HTML insertion (Line 1484) and keep `plainTitle`. I ended up putting it here since `$collaboratorName` is only used for insertion and it felt like I was hiding an escape at insertion.

BrianLiSK: However, it might be wise to start using `getDisplayTitle()` which returns a string already escaped. Maybe @marta-  can give a suggestion?

BrianLiSK: After some discussion, I'll change it back to `plainTitle` and follow ALAP convention. Thanks for the help.

sashaandjic:
When entering genes: 
image

components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml
```diff
@@ -1055,14 +1055,14 @@
           collaboratorText = "$escapetool.javascript($services.localization.render('phenotips.accessRightsManagement.collaboratorsSummary.noCollaborators', ['__beginHtml__', '__endHtml__']))".replace("__beginHtml__", beginHtml).replace("__endHtml__", endHtml);
         } else {
           var collaborator = data.collaborators.collaborators[0];
-          collaborator.name = collaborator.name || XWiki.Model.resolve(collaborator.id, XWiki.EntityType.DOCUMENT).name;
+          var safeCollaboratorName = (collaborator.name || XWiki.Model.resolve(collaborator.id, XWiki.EntityType.DOCUMENT).name).escapeHTML();
```
BrianLiSK: Since the variable `collaborator` was only used for injection, it might be permissible to leave `collaborator.name` as possibly null and use `safeCollaboratorName` for injection.

components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
```diff
@@ -123,12 +123,12 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
     var container = new Element('div', {'class': 'user'});
     var avatarWrapper = new Element('div', {'class': 'user-avatar-wrapper'});
     if (data.icon.indexOf('/') >= 0) {
-      avatarWrapper.insert(new Element('img', {src: data.icon, alt: data.info, 'class': 'icon'}));
+      avatarWrapper.insert(new Element('img', {src: data.icon, alt: data.info.escapeHTML(), 'class': 'icon'}));
```
BrianLiSK: Unfortunately, there is a regression somewhere. Do not merge.
`(xwiki)` appears on the userpicker. Fixing...

components/base-war/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.js
```diff
@@ -123,15 +123,15 @@ widgets.UserPicker = Class.create(widgets.Suggest, {
...
-      avatarWrapper.insert(new Element('span', {alt: data.info, 'class': data.icon}));
+      avatarWrapper.insert(new Element('span', {alt: data.info.escapeHTML(), 'class': data.icon}));
     }
     container.insert(avatarWrapper);
-    var userName = source.highlight ? this.emphasizeMatches(this.sInput, data.info) : data.info;
+    var userName = source.highlight ? this.emphasizeMatches(this.sInput.escapeHTML(), data.info.escapeHTML()) : data.info.escapeHTML();
     container.insert(new Element('div', {'class': 'user-name'}).update(userName));
     var referenceWrapper = new Element('div');
-    var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id;
+    var userAlias = source.highlight ? this.emphasizeMatches(this.sInput, data.id.replace(/.*\//g, '')) : data.id.escapeHTML();
```
sdumitriu: Missing escape on the true part of the ternary operator.

sdumitriu:
> Also broken: adding a collaborator with these chars will fail with `Failed to update. Unknown collaborator`, plus the alert is shown when this message is displayed.

I think this happens when adding the collaborator, then deleting that workgroup, then trying to update the permissions.

BrianLiSK:
> > Also broken: adding a collaborator with these chars will fail with `Failed to update. Unknown collaborator`, plus the alert is shown when this message is displayed.
> 
> I think this happens when adding the collaborator, then deleting that workgroup, then trying to update the permissions.

Was the collaborator a group? I encountered this at one point and thought I fixed it. Guess not: https://github.com/phenotips/phenotips/pull/2857/files#diff-441aa11a7addf2a1128d806f8ecee202R1058

PR #2859, opened 2019-04-24
PT-3980: Study descriptions are not rendered in the study selection dialog

marta-:


marta-:
When testing, check that this change does not introduce regressions such as https://phenotips.atlassian.net/browse/PT-3092 .

components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
```diff
@@ -58,7 +58,7 @@
 {{html clean="false" wiki="false"}}
 &lt;div class="form-template-option"&gt;
 &lt;label&gt;&lt;input type="radio" name="${prefix}${name}" value=""#if ("$!{value}" == '') checked="checked"#end/&gt; $services.localization.render('phenotips.studyBindingClass.noStudyOption')&lt;/label&gt;
-&lt;p class="xHint"&gt;$services.localization.render('phenotips.studyBindingClass.noStudyOption.hint')&lt;/p&gt;
+&lt;div class="xHint"&gt;$services.localization.render('phenotips.studyBindingClass.noStudyOption.hint')&lt;/div&gt;
```
BrianLiSK: I think it looks a bit better with a `<p>` before the hint. Other than that, I couldn't find any regressions caused by this change, just somethings that were always there. Based on the comments on the ticket, I would otherwise approve.
Minor Styleling

marta-: Good point, the spacing is a bit off now. I've improved it in the latest commit with some styling. I'd like to keep the `div` instead of `p` because:
* the study description container can't be `p`, as `p` elements can't contain other block elements that might be generated when rendering the description
* the container of the hint for "None" should be the same type of element as for study descriptions, for consistency.

components/studies/ui/src/main/resources/PhenoTips/StudyBindingClass.xml
```diff
@@ -72,7 +72,7 @@
     #end
     &lt;div class="form-template-option"&gt;
     &lt;label&gt;&lt;input type="radio" name="${prefix}${name}" value="${escapetool.xml($services.model.serialize($s.documentReference, 'default'))}" #if ("$!{value}" == "${s.documentReference}") checked="checked" #end/&gt; $escapetool.xml($configTitle)&lt;/label&gt;
-    &lt;p class="xHint"&gt;$!{escapetool.xml($s.content)}&lt;/p&gt;
+    &lt;div class="xHint"&gt;$!s.getRenderedContent()&lt;/div&gt;
```
sdumitriu: I wonder if a "plainitization" of the rendered content would be the best approach. I agree that the raw content isn't right, but I also don't think that we need the full HTML in here. Here's the difference between the two:

image

```suggestion
    &lt;div class="xHint"&gt;$!{escapetool.xml($s.getRenderedContent($services.rendering.resolveSyntax('plain/1.0'))).replaceAll('\n', '&lt;br /&gt;')}&lt;/div&gt;
```

marta-: Sounds good, that would prevent weird font sizes and layouts caused by elements such as headings/img .

sashaandjic:
It'll be nice to have study names taken care of too.
image

PR #2860, opened 2019-04-25
PT-3981: After visiting a "secondary action" link, the text becomes white and therefore unreadable in the default color theme

marta-:
Fixed the CSS selector that was too broad and affected more elements than originally intended.

### DIFF
--- a/components/gene-panels/ui/src/main/resources/PhenoTips/PanelsDisplay.xml
+++ b/components/gene-panels/ui/src/main/resources/PhenoTips/PanelsDisplay.xml
@@ -549,8 +549,10 @@
 }
 
 #panels-livetable input.button,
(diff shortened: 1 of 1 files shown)


marta-:
> Gene suggestions seem to not be working
This PR couldn't possibly have caused such a regression, it simply changes a css selector for links. The problem must be somewhere In master with one of the recent commits. @sdumitriu

veronikaslc:
`GenePanelsLiveTableResource` REST request returns with no content cause [this line](https://github.com/phenotips/phenotips/blob/master/components/gene-panels/rest/src/main/java/org/phenotips/panels/rest/internal/DefaultGenePanelLoader.java#L138) throws `ExecutionException`

PR #2866, opened 2019-04-30
PT-3936: Ability to support gene statuses that are organized hierarchically

veronikaslc:


components/patient-data/indexing/src/test/java/org/phenotips/data/indexing/internal/SolrPatientIndexerTest.java
```diff
@@ -91,8 +91,7 @@
 {
     private static final String STATUS_KEY = "status";
 
-    private static final List<String> STATUS_VALUES = Arrays.asList("candidate", "rejected", "rejected_candidate",
-        "solved", "carrier");
+    private static final List<String> STATUS_VALUES = Arrays.asList("status1", "status2", "status3");
```
marta-: This testing is incomplete. Since you're introducing hierarchical statuses, you should also test statuses that contain one or more the descendant delimiters `>` to ensure this character does not pose any problems.

veronikaslc: addressed in last commit

PR #2867, opened 2019-05-03
Fix BOADICEA export format to work with BOADICEA web application

danielpgross:
The webapp UI requires the testing method column to be set when the BRCA
mutation column has a value other than 0.

(committing on behalf of @buske)

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/model/export.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/model/export.js
@@ -415,6 +415,7 @@ define([
      var alertUnknownGenderFound = false; // BOADICEA does not support unknown genders
      var warnAboutMissingDOB     = false; // BOADICEA seem to require all individuals with cancer to have some age specified
      var warnMissingDOBUnaff     = false; // BOADICEA recommends age information for unaffected individuals
+     var warnBrcaTestMethod      = false; // BOADICEA needs a gene testing method to estimate FN rate for negative results, but this isn't currently able to be specified in the pedigree
 
      for (var i = 0; i <= pedigree.GG.getMaxRealVertexId(); i++) {
          if (!pedigree.GG.isPerson(i)) continue;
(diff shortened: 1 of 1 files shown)


danielpgross:
@sdumitriu is there anything holding this back from being merged? We'd like to get it merged ASAP.

buske:
@allasm Nope, no column was added. The column was already being exported and was always being set to `0` (untested), which caused BOADICEA to crash if there was a reported BRCA mutation.

components/pedigree/resources/src/main/resources/pedigree/model/export.js
```diff
@@ -656,6 +660,11 @@ define([
                            " year of death, if applicable, specified or estimated." +
                            " Not doing so may lead to an overestimation of risk.");
          }
+         if (warnBrcaTestMethod) {
+             warnings.push("BOADICEA requires the testing method (search vs. direct) to be specified to" +
+                           " estimate false negative rate. The testing method has been set to 'search' to" +
+                           " avoid overestimating risk, but you should review and correct this information in BOIDICEA.");
```
sdumitriu: Typo in `BOIDICEA`

components/pedigree/resources/src/main/resources/pedigree/model/export.js
```diff
@@ -580,9 +580,8 @@ define([
             warnMissingDOBUnaff = warnMissingDOBUnaff || !is_affected;
          }
 
-         output += "0\t"; // TODO: Genetic test status
```
sdumitriu: So the old order was an always `0` column, followed by a `status` column, and the new order is a status column, followed by a testing method column. Was it wrongly implemented in the first place, or is this something that changed between BOADICEA versions? I see that the documentation in https://pluto.srl.cam.ac.uk/bd3/v3/docs/BWA_v3_user_guide.pdf lists them in the original order.

danielpgross: Line 625 in the new version:
```
         output += testingMethod + "\t";
         output += mutationStatus + "\t";
```
The order hasn't changed, it's still testing method followed by mutation status.

components/pedigree/resources/src/main/resources/pedigree/model/export.js
danielpgross: Thanks for catching this, I will make another commit to fix.

PR #2869, opened 2019-05-06
PT-4004: Patient Solr index should include ENS IDs and not gene names in the index

allasm:


components/patient-data/migrations/pom.xml
```diff
@@ -63,6 +63,11 @@
       <artifactId>patient-data-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
sdumitriu: This migrator shouldn't be here, but in the indexing module.

PR #2871, opened 2019-05-07
PT-4003 (stable): Empty PubMedID are found after PhenoTips 1.3.10 upgrade to 1.4.6

veronikaslc:


components/patient-data/migrations/src/main/java/org/phenotips/data/internal/R54693PhenoTips1500DataMigration.java
```diff
@@ -90,9 +91,12 @@ public void hibernateMigrate() throws DataMigrationException, XWikiException
         /** The names of the properties to fix. */
         private final String propertyName = "global_mode_of_inheritance";
 
+        private XWikiContext context;
+
         @Override
         public Object doInHibernate(Session session) throws HibernateException, XWikiException
         {
+            this.context = getXWikiContext();
```
allasm: any reason not to get context where we need it (to make it more clear why we need it), i.e. on line 126 for this migrator? Can also simplify the change to just one line `getXWikiContext().getWiki().flushCache(this.context);` to avoid creating an unneeded private variable. Same in all other migrators.

PR #2872, opened 2019-05-08
PT-4002 (stable): Reindexing HPO fails when an HPO term has more than one single value type field

veronikaslc:
Fixed.

### DIFF
--- a/components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractOBOSolrVocabulary.java
+++ b/components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractOBOSolrVocabulary.java
@@ -115,7 +115,7 @@ public abstract class AbstractOBOSolrVocabulary extends AbstractSolrVocabulary
             return 2;
         }
         try {
-            Set<String> singleValuedFileds = getSingleValuedFields();
+            Set<String> singleValuedFields = getSingleValuedFields();
 
             Collection<SolrInputDocument> termBatch = new HashSet<>();
             Iterator<Map.Entry<String, TermData>> dataIterator = data.entrySet().iterator();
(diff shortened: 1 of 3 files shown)


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractOBOSolrVocabulary.java
```diff
@@ -118,12 +128,23 @@ protected int index(String sourceUrl)
                     termBatch = new HashSet<>();
                     batchCounter = 0;
                 }
+                singleValuedFileds.forEach(fieldName -> singleValueFieldsTracker.put(fieldName, false));
```
allasm: you only need `singleValueFieldsTracker` here, so the right way is to define it here, and initialize it here.

You also don't need a `Map` in that case, it can be done much simpler, as a `Set`, where a presense of  afield indicates it has been encountered already, no need to add a `true` to that

E.g. on this line initialize `Set<String> alreadyPresentFields = new HashSet<>();`

then add all fields encountered on line #146 (can even remove an `if` to reduce cyclomatic complexity).

this way also never have to clear the Set

components/vocabularies/chebi/api/src/test/java/org/phenotips/vocabulary/internal/solr/ChEBIOntologyTest.java
```diff
@@ -88,7 +88,6 @@ public void testChEBIOntologyReindex()
         Mockito.verify(this.server, Mockito.atLeast(1)).commit();
         Mockito.verify(this.server, Mockito.atLeast(1)).add(Matchers.anyCollectionOf(SolrInputDocument.class));
         Mockito.verify(this.cache, Mockito.atLeast(1)).removeAll();
-        Mockito.verifyNoMoreInteractions(this.cache, this.server);
```
allasm: why this change btw?

veronikaslc: because of the interaction with SolrClient to read single-valued schema fields.

allasm: I think there is a way to say "exactly one interaction"

PR #2873, opened 2019-05-08
PT-4002: Reindexing HPO fails when an HPO term has more than one single value type field

veronikaslc:
Fixed


components/vocabularies/api/src/main/java/org/phenotips/vocabulary/internal/solr/AbstractOBOSolrVocabulary.java
```diff
@@ -184,4 +201,28 @@ public String getVersion()
         }
         return null;
     }
+
+    /**
+     * Returns list of single-valued schema filed names .
+     *
+     * @return list of names
+     */
+    private List<String> getSingleValuedFileds()
```
buske: `Fileds` -> `Fields`

sashaandjic:
image

PR #2879, opened 2019-05-15
PT-4004: Patient Solr index should include ENS IDs and not gene names in the index

sdumitriu:
Fixed build issue


allasm:
I don'g like the idea of spreading migrators all over the place. Even though this migrator is related to indexing, it relies on DB version to either start or not, it is much easier to have them all in one place to know which DB version corresponds to what state of data

sdumitriu:
But we're planning on moving the whole indexing module to patient-network, so the migrator should stay close for the move.

PR #2882, opened 2019-05-31
PT-4018: Pedigree node life status not loaded from saved JSON

buske:
The `life_status` field takes values other than "alive" and "deceased" in regular, unlinked family member node json representations. Perhaps it was meant to be a special pedigreeProperty, but as is this drops all other life status values, causing it to return to the default of "alive".


buske:
Thanks @allasm @sashaandjic

PR #2884, opened 2019-06-11
PT-3779 (stable 1.4.x): Ability to associate specific permissions or studies to records automatically based on user/group preferences

veronikaslc:


sashaandjic:
- Defaults are missing for owner and global visibility in the Data Entry form

sashaandjic:
Peter had his own defaults set, but then pick his group defaults. UI is showing both and it is not clear what values will be used as defaults:
image
Once group default are chosem, the user default should be hidden. @Marta suggested adding a radio buttons for:
- Use my group's preferences
- Use my preferences

Once Peter was removed from the group:
- his profile still shows that he prefers group default setting
- his personal setting should are not applied to new records

PR #2886, opened 2019-06-12
PT-4020 In the pedigree editor, cancer labels are sometimes vertically offset from the relevant individual

danielpgross:
@allasm helpfully suggested a quick fix for this: making the user-defined `comments` label be the last one displayed (i.e. changing the order of labels displayed). He suggested adding better error checking for the values returned by `getBBox()` as a long-term solution.

After taking a closer look, I decided to fix the `comments` label rendering behaviour. This seems to me like a more targeted fix, since it does not change UX behaviour (i.e. order of the labels) and fixes the bug that's really relevant here (i.e. that the user-defined `comments` label is rendered even when it's effectively empty). Better error-checking for the values returned by `getBBox()` could still be added as an improvement, but shouldn't really be necessary.

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/view/personVisuals.js
@@ -822,9 +822,10 @@ define([
          */
         updateCommentsLabel: function() {
             this.getCommentsLabel() && this.getCommentsLabel().remove();
+            var text = typeof this.getNode().getComments() === 'string' ? this.getNode().getComments() : "";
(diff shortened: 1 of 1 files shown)


allasm:
Code looks OK, did not test

danielpgross:
Thanks @allasm, is there any normal case where `this.getNode().getComments()` could be `null`?

allasm:
I don't think so. Comments are initially set to an empty string, and then to a value from the comment box which will be empty and not null if not set I think. At the same time if time is spent examining that piece of code we can as well add a null check if we worry about future code changes.

PR #2887, opened 2019-06-12
PT-4019 After any edits, cursor is placed to the very end of the comment box in IE11

danielpgross:
This broke because `document.selection` stopped working in IE11 (it's `undefined`).

My solution is to use a different way to check whether the browser is IE, and then to restore the cursor position if it had been saved (rather than checking the browser type again in the restore).

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/view/nodeMenu.js
@@ -612,8 +612,9 @@ define([
         _saveCursorPositionIfNecessary: function(field) {
             this.__lastSelectedField  = field.name;
             this.__lastNodeID         = this.targetNode;
-            // for text fields in all browsers, and textarea only in IE9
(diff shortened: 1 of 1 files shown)


danielpgross:
I have tested this in IE9, IE10, IE11 on Windows and Chrome 74 on macOS.

danielpgross:
@buske and I looked at this more together today. Strangely, `_saveCursorPositionIfNecessary()` is sometimes called _after_ `_restoreCursorPositionIfNecessary()` when editing a given field, if the keyboard keys are pressed very quickly in succession. We're not sure exactly why this is, the answer lies deeper in the internals of the pedigree editor logic.
Screen Shot 2019-06-13 at 1 34 19 PM

We were concerned that this might cause a bug (since a restore should always happen after the corresponding save), but it seems to be a nonissue, because a new save will happen before the next restore. We also tested more extensively in IE11, focusing on edge cases, and did not find any issues. Therefore, we agreed that this change is ready to be merged. At some point in the future, the reason for the out-of-order save and restore behaviour can be examined.

PR #2889, opened 2019-06-25
PT-3917: sensitive pedigree details are shown when the pedigree is opened

buske:
In the current version, if you open the pedigree, it immediately alerts you to exactly that sensitive information that you might not want to see with the patient present:
image

In the PR version, it first displays a generic break-glass message, with options to continue, go to the record, or show the details:
image

If you click "Show details", it displays them below:
image

If you didn't provide any details, it doesn't show the button at all.
image

### DIFF
--- a/components/pedigree/resources/src/main/resources/pedigree/familyData.js
+++ b/components/pedigree/resources/src/main/resources/pedigree/familyData.js
@@ -11,12 +11,11 @@ define([
         Helpers
     ){
     var FamilyData = Class.create( {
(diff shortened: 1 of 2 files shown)


buske:
After feedback from users, I've moved the "Show/Hide Details" button to the center with the other two buttons.

buske:
> Tested. Works as described in the requirement.
> Perhaps the sensitive info should also be hidden in the Family page.
> image

Thanks for testing and the feedback, @sashaandjic. I agree with it not being shown on the Family page, but it isn't an issue for our client yet so we didn't need to tackle it. Instead, the sensitive info should probably just be moved into the pedigree entirely.

PR #2896, opened 2019-08-21
PT-4031: Gestational age in pedigree isn't saved

buske:


buske:
Decided to not bother merging this into master

